PR #10, opened 2023-05-10
feat: node.js export

Yash-Singh1:
Currently the `server` export is missing the `node` export for Node.js
I needed to patch this to add support in my websocket server which was running on Node.js

### DIFF
--- a/packages/uploadthing/package.json
+++ b/packages/uploadthing/package.json
@@ -11,7 +11,7 @@
     "./server": {
       "import": "./dist/server.mjs",
       "types": "./dist/server.d.ts",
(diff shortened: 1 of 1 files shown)


packages/uploadthing/package.json
```diff
@@ -10,7 +10,8 @@
     },
     "./server": {
       "import": "./dist/server.mjs",
-      "types": "./dist/server.d.ts"
+      "types": "./dist/server.d.ts",
+      "node": "./dist/server.mjs"
```
juliusmarminge, line 14: I think its better if we use the more generic `default`

```suggestion
      "default": "./dist/server.mjs"
```

PR #17, opened 2023-05-10
don't ignore cache output

juliusmarminge:
any reason you were ignoring this?


t3dotgg (approved):
I copied this from somewhere, no idea why lol

PR #19, opened 2023-05-10
:bug: (refactor): adding an exponential backoff

guillermoriv:
I was testing out, and I thought to add an exponential backoff for the "dev" env, as I suppose you cannot open a S3 in local


markflorkowski:
Closing in favor of #21

PR #21, opened 2023-05-10
Add exponential backoff for conditionalDevServer

GithubAccountOfDanielDavis:


markflorkowski:
There should probably be a `MAX_RETRIES` to prevent infinite calls

GithubAccountOfDanielDavis:
> There should probably be a `MAX_RETRIES` to prevent infinite calls

I added this in and also put the return statement and final error handling back in. That got lost in the shuffle before while I was breaking out the retry logic into its own function.

markflorkowski (approved):

markflorkowski:
OH, good catch on the missing await cc @t3dotgg I think we need that eslint rule for hanging promises

PR #22, opened 2023-05-10
fixes issue where pnpm dev fails on first run

### LINKED ISSUE #4: `pnpm dev` fails the first time you run it
ONLY WHEN WORKING ON THIS REPO TO BE CLEAR

This bug is because `@uploadthing/react` depends on `uploadthing`, but you can't block in dev or it will block forever. I tried hacking this with a `prebuild` step before but it was buggy as hell. Surprised there's no recommendations here

iway1:
closes #4

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -5,11 +5,11 @@
   "license": "MIT",
   "scripts": {
     "dev": "next dev --port 3030",
+    "clean": "git clean -xdf dist node_modules",
     "build": "next build",
     "start": "next start",
     "debug": "NODE_OPTIONS='--inspect' next dev",
-    "vercel-build": "yarn workspace nextra run build && yarn workspace nextra-theme-docs run build && yarn build",
-    "clean": "rimraf .next .turbo"
+    "vercel-build": "yarn workspace nextra run build && yarn workspace nextra-theme-docs run build && yarn build"
   },
   "dependencies": {
     "next": "13.4.1",
--- a/examples/appdir/package.json
+++ b/examples/appdir/package.json
@@ -4,6 +4,7 @@
   "private": true,
   "scripts": {
     "dev": "next dev -p 3000",
+    "clean": "git clean -xdf .next node_modules",
     "build": "next build",
     "start": "next start",
     "lint": "next lint",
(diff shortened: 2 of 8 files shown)


t3dotgg:
Concern w/ this solution is it breaks the "contract" - the types in dev will not necessarily be the same as in prod. My gut feel is that using the types in dist is the better call, iirc @juliusmarminge agrees

iway1:
I don't think this changes the types that end up in the build at all it still imports the types from `uploadthing/server`

For example if running `tsup` in the `react` package the output `index.d.ts` will still import the type:
 
Screenshot 2023-05-10 at 8 00 02 PM

doesn't seem like this changes the output of build

iway1:
Changed it to use `wait-on` package as per https://github.com/vercel/turbo/discussions/1299. Doesn't seem like there's a way to do it with normal turbo config 🤷

iway1:
Now uses `wait-on` package to wait for core to build before running dev in react package as per https://github.com/vercel/turbo/discussions/1299, doesn't look like there's a way to change it with normal turbo config 🤷

iway1:
Now uses `wait-on` package to wait for core to build before running tsup in react package as per https://github.com/vercel/turbo/discussions/1299. 

Doesn't seem like there's a way to do it with turbo config 🤷

iway1:
BTW pulling a fresh repo and doing `pnpm install && pnpm dev` creates the react `dist` correctly. It does throw an error in the terminal somewhere:

```sh
@uploadthing/react:dev: src/useUploadThing.ts(3,40): error TS2307: Cannot find module 'uploadthing/client' or its corresponding type declarations.
```

But it generates `dist` correctly and the `tsup --watch`  is still running. 

I changed my PR to use the `wait-on` package to wait for core to build as per https://github.com/vercel/turbo/discussions/1299 if you want to get rid of the error though

juliusmarminge (approved):

juliusmarminge (approved):
Works like a charm. Nice and clean.

I also added a clean script since i noticed we didn't have one when testing. It's good to have when things aren't working unexpectedly.

PR #30, opened 2023-05-11
added dark mode in docs

EzraChai:


t3dotgg:
Why did you commit a package-lock?

t3dotgg:
You also failed to update the logo so it will be invisible for all dark mode users. Closing this

markflorkowski:
Additionally, adding dark mode is not just flipping a switch, as the logo will also need to be modified to support this change.

PR #34, opened 2023-05-11
(deps) Consider adding Renovate

LZL0:
I believe Renovate would significantly improve the dependency management process.
Renovate is a free app, a much better version of dependabot, that opens pull requests and takes care of most dependency updates. There are several settings that you can adjust, you can even ask it to automatically merge as seen below (which might not be a good idea for this project):
```
"packageRules": [
    {
      "matchUpdateTypes": ["minor", "patch", "lockFileMaintenance"],
      "automerge": true
    }
  ]
```

You can install it per repository or for the entire organization [here](https://github.com/marketplace/renovate). It uses GitHub actions to process your dependencies. Yes, the constant dependency notifications can be annoying. 🤣

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,4 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": ["config:base"]
+}


t3dotgg:
Mark and I are neurotic enough to keep things up to date ourselves, ty tho

PR #35, opened 2023-05-11
Adds i18n support and German translations to docs

Filsommer:
Following Nextra's [i18n guide](https://nextra.site/docs/guide/i18n), I've added support for internationalisation and translated the docs into German 🇩🇪

Notes: I don't really like the way the guide mentioned above splits the locales (index.en.md, index.de.md, etc) and we should consider refactoring this once more languages are added, to avoid a messy /pages directory.


t3dotgg:
Due to the incredibly early state of the docs, I think it's best to hold off on translation efforts until we get closer to leaving beta. We've already caused 5 conflicts lol.

Thank you for the contribution, keep an eye out for when we're a bit closer.

PR #43, opened 2023-05-11
ADDED issue template for bug report

Caketime:
ADDED basic code of conduct
image
image
image

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG-REPORT.yml
+++ b/.github/ISSUE_TEMPLATE/BUG-REPORT.yml
@@ -1,0 +1,59 @@
+name: Bug Report
+description: File a bug report
+title: "[Bug]: "
+labels: ["unconfirmed_bug"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for taking the time to fill out this bug report!
+  - type: textarea
+    id: what-happened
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/BUG-REPORT.yml
```diff
@@ -0,0 +1,47 @@
+name: Bug Report
+description: File a bug report
+title: "[Bug]: "
+labels: ["bug"]
```
juliusmarminge, line 4: unconfirmed_bug is probably more used and then we switch to bug once its reproduced

Thats how we do in t3-oss and trpc atleast

Caketime:
Force pushed an update so it stays a single commit.

Changes: Label "bug" -> "uncofirmed_bug"

markflorkowski (changes requested):
I'd also like to have "Reproduction Repository" and "Reproduction Steps" fields. They can be optional. 

Otherwise looks good!

Caketime:
- Added optional "Reproduction Repository" and "Reproduction Steps" fields

- Updated the "what did you expect to happen" description from "What did you expect to happen?" to "Please describe what you expected to happen."

markflorkowski (approved):

Issue #37, opened 2023-05-11
Multiple file upload with uploadButton


kj-1809:
Apparently , it's not possible to add multiple files while using uploadButton.

Gaurav-r-a-j:
To upload multiple files using uploadbutton or uploaddropzone you need to set maxFilecount here 


export const ourFileRouter = {
  // Define as many FileRoutes as you like, each with a unique routeSlug
  imageUploader: f({ image: { maxFileSize: "4MB"  , maxFileCount:10} }) // setting maxfilecoutn to allow mulitple file upload
    // Set permissions and file types for this FileRoute
    .middleware(async ({ req }) => {
      // This code runs on your server before upload
      const user = await auth(req);

      // If you throw, the user will not be able to upload
      if (!user) throw new UploadThingError("Unauthorized");

      // Whatever is returned here is accessible in onUploadComplete as `metadata`
      return { userId: user.id };
    })
    .onUploadComplete(async ({ metadata, file }) => {
      // This code RUNS ON YOUR SERVER after upload
      console.log("Upload complete for userId:", metadata.userId);

      console.log("file url", file.url);

      // !!! Whatever is returned here is sent to the clientside `onClientUploadComplete` callback
      return { uploadedBy: metadata.userId };
    }),
} satisfies FileRouter;

PR #48, opened 2023-05-11
fix(uploadButton): accept mimeTypes join & allow multiple files to pass

OrJDev:
This will allow multiple files to be passed when using the uploadButton comp (let the user select more than 1 file optionally)
this also fixes the mimeTypes for uploadButton (it was joined with "," when it should have been ", " therefore it didn't work and you could pass any file you want)

closes https://github.com/pingdotgg/uploadthing/issues/37

### DIFF
--- a/.changeset/olive-doors-kick.md
+++ b/.changeset/olive-doors-kick.md
@@ -1,0 +1,5 @@
+---
+"@uploadthing/react": patch
+---
+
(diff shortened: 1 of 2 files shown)


markflorkowski (approved):

markflorkowski:
I fixed the merge conflict, and default to single file, as well as indicate in the button text if multiple files are supported.

PR #49, opened 2023-05-11
[UPL-37] Better Error Flow

markflorkowski:
- better errrors from the client
- catch errror in hook
- add onUploadError to components
- update examples
- update docs

### DIFF
--- a/packages/react/src/component.tsx
+++ b/packages/react/src/component.tsx
@@ -17,11 +17,13 @@ type EndpointHelper<TRouter extends void | FileRouter> = void extends TRouter
 export function UploadButton<TRouter extends void | FileRouter = void>(props: {
   endpoint: EndpointHelper<TRouter>;
   onClientUploadComplete?: () => void;
+  onUploadError?: (error: Error) => void;
 }) {
   const { startUpload, isUploading, permittedFileInfo } =
     useUploadThing<string>({
       endpoint: props.endpoint as string,
       onClientUploadComplete: props.onClientUploadComplete,
+      onUploadError: props.onUploadError,
     });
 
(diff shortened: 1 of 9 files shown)


packages/react/src/component.tsx
```diff
@@ -17,11 +17,13 @@ type EndpointHelper<TRouter extends void | FileRouter> = void extends TRouter
 export function UploadButton<TRouter extends void | FileRouter = void>(props: {
   endpoint: EndpointHelper<TRouter>;
   onClientUploadComplete?: () => void;
+  onUploadError?: (error: Error) => void;
 }) {
   const { startUpload, isUploading, permittedFileInfo } =
     useUploadThing<string>({
```
t3dotgg, line 23: Can we just pass `props` here?

packages/uploadthing/src/internal/handler.ts
```diff
@@ -67,13 +67,17 @@ const withExponentialBackoff = async <T>(
     backoffMs = Math.min(MAXIMUM_BACKOFF_MS, backoffMs * 2);
     backoffFuzzMs = Math.floor(Math.random() * 500);
 
-    console.error(`[UT] Call unsuccessful after ${tries} tries. Retrying in ${Math.floor(backoffMs / 1000)} seconds...`)
+    console.error(
```
t3dotgg, line 70: We should probably not log this til 3 or 4 tries in, not necessary for this pass

t3dotgg (approved):

packages/react/src/component.tsx
markflorkowski: No, because EndpointHelper type not resolving to string. Maybe this type can be refined to make this work?

PR #54, opened 2023-05-12
file size improvements

didar-dev:
- Added input validation using a regular expression to ensure the `size` parameter is in the correct format.
- Modified the function to return an `Error` object when encountering invalid size formats or units, allowing for graceful error handling.
- Extracted the conversion factors into a separate `conversionFactors` object for improved readability and maintainability.
- Updated variable naming for clarity and consistency.
- Allowed for decimal numbers in the file size and implemented case-insensitive unit matching.
- cast back to int in the return (Thanks @markflorkowski )

### DIFF
--- a/.changeset/rotten-numbers-tan.md
+++ b/.changeset/rotten-numbers-tan.md
@@ -1,0 +1,5 @@
+---
+"uploadthing": patch
+---
+
+Better parsing for `.fileSize()` in the router. thanks @didar-dev!
--- a/packages/uploadthing/src/internal/handler.ts
+++ b/packages/uploadthing/src/internal/handler.ts
@@ -1,4 +1,4 @@
-import type { AnyRuntime, FileRouter, FileSize, SizeUnit } from "../types";
+import type { AnyRuntime, FileRouter, FileSize } from "../types";
 import type { NextApiRequest, NextApiResponse } from "next";
 
 const UPLOADTHING_VERSION = require("../../package.json").version;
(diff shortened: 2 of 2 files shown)


markflorkowski (changes requested):
I like where this is going, I think I would prefer something like this though:

```ts
const UNITS = ["B", "KB", "MB", "GB"] as const;
type SizeUnit = (typeof UNITS)[number];

export const fileSizeToBytes = (input: string) => {
  const regex = new RegExp(`^(\\d+)(\\.\\d+)?\\s*(${UNITS.join("|")})$`, "i");
  const match = input.match(regex);

  if (!match) {
    return new Error("Invalid file size format");
  }

  const sizeValue = parseFloat(match[1]);
  const sizeUnit = match[3].toUpperCase() as SizeUnit;

  if (!UNITS.includes(sizeUnit)) {
    throw new Error("Invalid file size unit");
  }

  return sizeValue * Math.pow(1024, UNITS.indexOf(sizeUnit));
};
```

markflorkowski:
Probably should also cast back to int in the return, as we should not have fractional bytes.

didar-dev:
Thank you @markflorkowski, I have made the requested changes and addressed the suggested improvements. The code now includes the use of the `UNITS` constant and the updated return statement casting the result to an integer to ensure whole bytes are returned. I have thoroughly tested the code, and it is now ready for merging. Please review the changes, and if everything looks good, I kindly request your approval to proceed with the merge

markflorkowski (approved):

PR #82, opened 2023-05-21
Export useUploadThing hook

donnes:
To be able to create our "own" `UploadButton` or `UploadDropzone` component, the hook `useUploadThing` is needed.


t3dotgg:
No it is not, you can use the generateReactHelpers hook to access it as intended :)

PR #86, opened 2023-05-22
add : how to allow multi-file upload

kj-1809:


markflorkowski (changes requested):
Please add this to all docs examples (pagedir/appdir/solid), then I think it should be good to merge 👍

t3dotgg:
@markflorkowski I'd prefer to not clog up the examples with this and put it in the api docs instead, can be convinced either way tho

markflorkowski:
Yeah, this would be good as part of the generic api docs

t3dotgg:
Closing in favor of me working on better api docs

PR #97, opened 2023-05-25
[Fix] Update to buildRequestHandler secret

ayatofrench:
## Change Log
- Updated `buildRequestHandler` 
  - Now uses `config.uploadthingSecret` instead of `config.uploadthingId`
  - Throws an error if secret does not exist
  - Renamed `upSecret` to `preferredOrEnvSecret`

### DIFF
--- a/packages/uploadthing/src/internal/handler.ts
+++ b/packages/uploadthing/src/internal/handler.ts
@@ -154,7 +154,14 @@ export const buildRequestHandler = <
     res?: TRuntime extends "pages" ? NextApiResponse : undefined;
   }) => {
     const { router, config } = opts;
-    const upSecret = config?.uploadthingId ?? process.env.UPLOADTHING_SECRET;
+    const preferredOrEnvSecret =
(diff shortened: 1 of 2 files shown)


packages/uploadthing/src/internal/handler.ts
```diff
@@ -154,10 +154,18 @@ export const buildRequestHandler = <
...
+      config?.uploadthingSecret ?? process.env.UPLOADTHING_SECRET;
     const { uploadthingHook, slug, req, res, actionType } = input;
+
     if (!slug) throw new Error("we need a slug");
+
+    if (!preferredOrEnvSecret) {
+      throw new Error(
+        `Please set your preferred secret in ${slug} router's config or set UPLOADTHING_SECRET in your env file`
+      );
+    }
```
ayatofrench, line 167: Not sure if there is a use case for this, but I did some slight rearranging here to get a slightly better error message just in case someone wanted to have separate secret configs for their various routers. If this is not desired I can definitely revert this!

t3dotgg (approved):
This is fantastic, thank you

PR #99, opened 2023-05-25
Removed @ts-expect-error for middleware in upload-builder.ts

Maki325:
Use a single function instead of interpreting from multiple signatures and use arguments from the `Args` type so it's still customizable.

It's also possible to use the code below, but it would give an error to the user if they wrote this `.middleware((req) => {...})` so they would need to write `.middleware((req, _res) => {...})` if they didn't need the response when using `next-legacy`.

```typescript
type MiddlewareFn<
  TOutput extends Record<string, unknown>,
  TRuntime extends string
> = Args<TRuntime>["res"] extends never
  ? (req: Args<TRuntime>["req"]) => MaybePromise<TOutput>
  : (
      req: Args<TRuntime>["req"],
      res: Args<TRuntime>["res"]
    ) => MaybePromise<TOutput>;
```

But the one plus would be it doesn't allow the `.middleware((req, _res) => {...})` when using `next` or `server`, while in the one committed `_res` is of type `never` so it's not really usable.

### DIFF
--- a/packages/uploadthing/src/types.ts
+++ b/packages/uploadthing/src/types.ts
@@ -39,25 +39,26 @@ type ResolverOptions<TParams extends AnyParams> = {
   file: UploadedFile;
 };
 
-export type ReqMiddlewareFn<TOutput extends Record<string, unknown>> = (
-  req: Request
-) => MaybePromise<TOutput>;
-export type NextReqMiddlewareFn<TOutput extends Record<string, unknown>> = (
-  req: NextRequest
-) => MaybePromise<TOutput>;
-export type NextApiMiddlewareFn<TOutput extends Record<string, unknown>> = (
-  req: NextApiRequest,
-  res: NextApiResponse
-) => MaybePromise<TOutput>;
+export type ReqMiddlewareFn<TOutput extends Record<string, unknown>> =
+  MiddlewareFn<TOutput, "web">;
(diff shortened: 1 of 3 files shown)


t3dotgg (approved):
This looks like an improvement to me. cc @juliusmarminge who is smarter than me

juliusmarminge (approved):
Yea this is much cleaner. Small nit comment just cause i got kinda confused when first reading it 😅 

Let's just confirm the tests passes ay

juliusmarminge:
There seems to be some edge cases that isn't covered here.

Check out the tests locally (especially the upload-builder.test.ts) to see what's wrong.

One of the faults is that it's now okay to pull the `res` even if it's never. Maybe not a big deal? But we're also always expecting 2 arguments when callign the middleware now

Maki325:
I've changed to the other `MiddlewareFn` type I spoke about above. And all the tests pass now.
I did test the old one yesterday, but seems sleepiness got the better of me, as I've misunderstood the number of tests cases to the number of tests passed.
Also, I've rearranged the types and renamed the `Args` type to `MiddlewareFn`.

`never` is still used, but now it signals that the function only has one argument not two.
I've tried to remove `never`, but then `res` is implicitly `any`, I think, and doesn't work with `extends` so it can't figure out that it should only be one argument and not two.

PR #101, opened 2023-05-26
fix: solves requests and responses types and adds typesafety 🧙‍

miguelvelasquezdev:
- Fix type error in packages/uploadthing/src/internal/handler.ts

Screenshot 2023-05-27 at 7 25 35 PM

- add typesafety to requests and possible responses

---

- [x] Funcitonal tests

### DIFF
--- a/packages/uploadthing/src/internal/handler.ts
+++ b/packages/uploadthing/src/internal/handler.ts
@@ -132,8 +132,8 @@ export const buildRequestHandler = <
     uploadthingHook?: string;
     slug?: string;
     actionType?: string;
-    req: Partial<Request> & { json: Request["json"] };
-    res?: TRuntime extends "pages" ? NextApiResponse : undefined;
+    req: RequestAdapter<TRuntime>;
+    res?: ResponseAdapter<TRuntime>;
   }) => {
     const { router, config } = opts;
     const preferredOrEnvSecret =
@@ -155,7 +155,7 @@ export const buildRequestHandler = <
 
     const reqBody = (await req.json()) as {
       file: UploadedFile;
-      files: unknown;
+      files: string[];
       metadata: Record<string, unknown>;
       input?: Json;
     };
--- a/packages/uploadthing/src/internal/handler.ts
+++ b/packages/uploadthing/src/internal/handler.ts
@@ -132,8 +132,12 @@ export const buildRequestHandler = <
     uploadthingHook?: string;
     slug?: string;
     actionType?: string;
-    req: RequestAdapter<TRuntime>;
-    res?: ResponseAdapter<TRuntime>;
+    req: TRuntime extends "pages"
+      ? NextApiRequest & RequestBase
+      : TRuntime extends "app"
+      ? NextRequest
+      : Partial<Request> & RequestBase;
+    res?: TRuntime extends "pages" ? NextApiResponse : undefined;
   }) => {
     const { router, config } = opts;
     const preferredOrEnvSecret =
--- a/packages/uploadthing/src/internal/handler.ts
+++ b/packages/uploadthing/src/internal/handler.ts
@@ -155,7 +155,7 @@ export const buildRequestHandler = <
 
     const reqBody = (await req.json()) as {
       file: UploadedFile;
-      files: string[];
+      files: unknown;
       metadata: Record<string, unknown>;
       input?: Json;
     };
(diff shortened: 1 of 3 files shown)


juliusmarminge:
Hey, thanks for this. We just merged a different PR for this thst simplifierad the types a bit. Can you merge that in and resolve the conflicts?

juliusmarminge:
Tests failing

packages/uploadthing/src/types.ts
```diff
@@ -52,29 +52,29 @@ type ResolverOptions<TParams extends AnyParams> = {
...
+      req: RequestAdapter<TRuntime>,
+      res: ResponseAdapter<TRuntime>
     ) => MaybePromise<TOutput>;
 
-export type ReqMiddlewareFn<TOutput extends Record<string, unknown>> =
-  MiddlewareFn<TOutput, "web">;
-export type NextReqMiddlewareFn<TOutput extends Record<string, unknown>> =
-  MiddlewareFn<TOutput, "app">;
-export type NextApiMiddlewareFn<TOutput extends Record<string, unknown>> =
-  MiddlewareFn<TOutput, "pages">;
```
juliusmarminge, line 76: not sure how many are using these but it's technically a breaking change removing these exports

packages/uploadthing/src/types.ts
```diff
@@ -52,29 +52,29 @@ type ResolverOptions<TParams extends AnyParams> = {
...
-  ? (req: MiddlewareFnArgs<TRuntime>["req"]) => MaybePromise<TOutput>
+  TRuntime extends string
+> = ResponseAdapter<TRuntime> extends undefined
+  ? (req: RequestAdapter<TRuntime>) => MaybePromise<TOutput>
   : (
-      req: MiddlewareFnArgs<TRuntime>["req"],
-      res: MiddlewareFnArgs<TRuntime>["res"],
+      req: RequestAdapter<TRuntime>,
+      res: ResponseAdapter<TRuntime>
     ) => MaybePromise<TOutput>;
```
juliusmarminge, line 76: im not convinced this is better

packages/uploadthing/src/internal/handler.ts
```diff
@@ -230,7 +230,7 @@ export const buildRequestHandler = <
 
     const reqBody = (await req.json()) as {
       file: UploadedFile;
-      files: unknown;
+      files: string[];
```
juliusmarminge, line 233: unknown is more accurate

packages/uploadthing/src/internal/handler.ts
```diff
@@ -253,8 +253,7 @@ export const buildRequestHandler = <
     }
 
     try {
-      const { files } = reqBody as { files: string[] };
-      // @ts-expect-error TODO: Fix this
+      const { files } = reqBody;
```
juliusmarminge, line 256: files should still be unknown here really... we validate it further down (L260).

I'm guessing the previous `as` here is to support the destructuring but maybe just skip that?

```ts
try {
  // ...
  if (!Array.isArray(reqBody.files) || !reqBody.files.every((f) => typeof f === "string"))
        throw new Error("Need file array");
  // ✅ string[]
```

packages/uploadthing/src/types.ts
miguelvelasquezdev: it is exported, but it's not used anywhere...

packages/uploadthing/src/internal/handler.ts
miguelvelasquezdev: so the `as` is unnecessary?

packages/uploadthing/src/types.ts
miguelvelasquezdev: Ohh, i got it

packages/uploadthing/src/types.ts
miguelvelasquezdev: I made changes in the last commit, maybe it's better to read

packages/uploadthing/src/upload-builder.test.ts
```diff
@@ -62,7 +62,7 @@ it("passes `Request` by default", () => {
   const f = createBuilder();
 
   f(["image"]).middleware((req) => {
-    expectTypeOf(req).toMatchTypeOf<Request>();
+    expectTypeOf(req).toMatchTypeOf<Partial<Request> & RequestBase>();
```
juliusmarminge, line 65: I don't like if this is exposed to the user - they should just see a `Request` / `NextRequest` or `NextApiRequest` etc

PR #103, opened 2023-05-26
chore: housekeeping 2 - emit ts-sourcemaps & monorepo config stuff

juliusmarminge:
- this allows jump-to-definition for the internal packages directly from the example apps - previously we just got to the /dist/*.d.ts file

https://github.com/pingdotgg/uploadthing/assets/51714798/4ab9ab24-8e17-4675-8ff0-5d1cab24338b

- adds a standard prettier config along with import sorter
- adds a shared tsup config
- uses the shared tsconfigs more
- standardizes the examples (make the structure the same for app and pages)


packages/config/tsup/index.js
```diff
@@ -0,0 +1,22 @@
...
+    "**/.turbo",
+    "**/dist",
+    "**/node_modules",
+    "**/.DS_STORE",
+    "**/.git",
+  ],
+  async onSuccess() {
+    // emit sourcemap to enable jump to definition
+    execSync("pnpm tsc --project tsconfig.sourcemap.json");
+  },
```
juliusmarminge, line 19: this is the stuff that generates the sourcemaps that allows jump-to-definition

prettier.config.cjs
```diff
@@ -0,0 +1,28 @@
...
+  printWidth: 80,
+  singleQuote: false,
+  semi: true,
+  trailingComma: "all",
+  tabWidth: 2,
+  plugins: [
+    "@ianvs/prettier-plugin-sort-imports",
+    "prettier-plugin-tailwindcss",
+  ],
+  importOrder: [
```
markflorkowski, line 15: Can you add `"<TYPES>"` to the end of the sort order? Mixing type imports with regular imports gets a bit confusing at times.

juliusmarminge: you want separate type imports? so like 

```ts
import { foo } from "foo";
import type { Bar } from "foo";
```

prettier.config.cjs
```diff
@@ -0,0 +1,28 @@
+/** @typedef  {import("prettier").Config} PrettierConfig */
+
+/** @type { PrettierConfig | SortImportsConfig } */
+const config = {
```
markflorkowski, line 4: Thoughts on adding `importOrderSortSpecifiers: true` and `importOrderGroupNamespaceSpecifiers: true`?

Might also want `importOrderTypeScriptVersion`

juliusmarminge: Funny enough I had some of these but prettier complained about not finding these optiosn so i removed it...

this is the config i usually roll: https://github.com/t3-oss/create-t3-turbo/blob/main/prettier.config.cjs

prettier.config.cjs
markflorkowski: Yes, exactly.

prettier.config.cjs
markflorkowski: Oh, It looks like 4.0.0 must have removed stuff then. They definitely still exist in 3.7.2

markflorkowski: yup: 
image

t3dotgg (approved):
Not excited for the merge conflict I'm about to eat on the v4 PR lol

prettier.config.cjs
juliusmarminge: I'll enable separate type imports by eslint instead: 

Uploading CleanShot 2023-05-27 at 11.39.21.mp4…

juliusmarminge:
> Not excited for the merge conflict I'm about to eat on the v4 PR lol

DW I'll resolve it all 😇

.vscode/settings.json
```diff
@@ -7,9 +7,12 @@
   "eslint.rules.customizations": [{ "rule": "*", "severity": "warn" }],
   "typescript.tsdk": "node_modules/typescript/lib",
   "eslint.workingDirectories": [
+    "docs",
     "examples/appdir",
```
markflorkowski, line 11: are wildcards supported here? if so, `examples/*` and `packages/*` so we don't have to worry about this again

docs/src/pages/solid.mdx
```diff
@@ -115,10 +117,13 @@ The `@uploadthing/solid` package includes an "Uploader" component that you can s
 ```tsx copy
 // You need to import our styles for the button to look right. Best to import in the root /layout.tsx but this is fine
 import "@uploadthing/solid/styles.css";
+
 import { type VoidComponent } from "solid-js";
```
markflorkowski, line 121: Is there anything we can do to prefer 
```
import type { ... } from "some-package"
```
over
```
import { type ... } from "some-package"

File: examples/solid/src/routes/api/uploadthing/core.ts [hunk 6]
```diff
@@ -1,7 +1,8 @@
 import { createUploadthing, type FileRouter } from "uploadthing/server";
+
 const f = createUploadthing();
 
-const auth = (req: Request) => ({ id: "fakeId" }); // Fake auth function
+const auth = (req: Request) => Promise.resolve({ id: "fakeId" }); // Fake auth function
```
markflorkowski, line 5: should this function be marked async?

markflorkowski (approved):
Just a few nits.

examples/solid/src/routes/api/uploadthing/core.ts
juliusmarminge: Linter yells since there are no awaits

.vscode/settings.json
juliusmarminge: I'll check

docs/src/pages/solid.mdx
juliusmarminge: I don't know if there is a way to force you to change inline-type to separate ones, but with the current setup the auto fix is set to use separate-types.

Maybe the eslint-plugin-import has an option for this

examples/solid/src/routes/api/uploadthing/core.ts
markflorkowski: ? 
image

docs/src/pages/solid.mdx
markflorkowski: It does :) 
https://github.com/import-js/eslint-plugin-import/blob/HEAD/docs/rules/consistent-type-specifier-style.md#prefer-top-level

examples/solid/src/routes/api/uploadthing/core.ts
juliusmarminge: What I mean is doing 

const auth = async => ({ ... });

Will result in the linter yelling at you cause your async function has no await. Doing the Promise.resolve thing makes it so the function is returning a promise (and thus can be awaited in the middleware like we want) without this lint warning.

Really a nit thing, can also just disable the linter if that's more to your liking

markflorkowski: Oh I see now. My bad. The way you have it is fine 👍

docs/src/pages/solid.mdx
juliusmarminge: https://github.com/pingdotgg/uploadthing/assets/51714798/b48ef707-93f9-4de2-9e15-ebf82265f1eb

now only top-level are allowed

PR #111, opened 2023-05-28
feat: allow custom styling of components

juliusmarminge:
This allows custom styling of the components

Checklist
- [x] Button
- [ ] Dropzone

I'm also experimenting with a new way of including tailwind styles through libs - using the content prop in the config. This should mean we're not generating duplicate styles for tailwind users either i.e. we won't have `ut-bg-blue-500` and `bg-blue-500`.

Non tailwind users should still import the stylesheet in their root layouts

https://github.com/pingdotgg/uploadthing/assets/51714798/827128d4-9d15-4a30-8bbe-8a8e6eae9df5


examples/appdir/tailwind.config.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
     "./src/pages/**/*.{js,ts,jsx,tsx}",
     "./src/components/**/*.{js,ts,jsx,tsx}",
     "./src/app/**/*.{js,ts,jsx,tsx}",
+    "./node_modules/@uploadthing/react/src/components/**",
```
juliusmarminge, line 7: might look a bit hacky but I do genuinely think this is the most optimal way to include tailwind classes from a lib into an app that uses tailwind

packages/react/package.json
```diff
@@ -60,5 +60,8 @@
   },
   "publishConfig": {
     "access": "public"
+  },
+  "dependencies": {
+    "tailwind-merge": "^1.12.0"
```
juliusmarminge, line 65: idk how you feel about this? this is probably the more sketchier thing

markflorkowski: Do you mean the fact that this is the only "real" dep, or that you find `tailwind-merge` sketchy?

juliusmarminge: tailwind-merge is great but always skeptical about bundling deps that aren't core to the package 🤔 making it a peerDep also doesn't really make sense though so i guess it's fine?

markflorkowski: Yeah, definitely doesn't make sense as a peerDep. I mean it is *kinda* core to the package, at least if we consider functional styles as core to the package 🤷

examples/appdir/tailwind.config.js
Mr0Bread: Maybe instead of requiring users to manually add this line, it would be possible to create a function that would automatically append that line by simply accepting config as argument, mutating it inside and then returning back

Something like

```typescript
import { withUt } from 'uploadthing/tailwind'

export default withUt({
  // whatever TW config
})
```

The same with CJS

```javascript
const { withUt } = require('uploadthing/tailwind')

module.exports = withUt({
  // whatever TW config
})
```

juliusmarminge:
Superseded by #217

PR #119, opened 2023-05-31
[fix] Make sure clientside onUploadComplete() is called after serverside

markflorkowski:


packages/uploadthing/src/utils.ts
```diff
@@ -86,3 +87,55 @@ export const getTypeFromFileName = (
...
+    const res = await fetch(queryUrl);
+    const json = (await res.json()) as
+      | { status: "done"; fileData: FileData }
+      | { status: "something else" };
+
+    if (json.status !== "done") return null;
+
+    await callback?.(json);
+  });
+}
```
t3dotgg, line 141: This should probably have an empty line at end right?

PR #122, opened 2023-05-31
fix: prefetch permissions on server to prevent layout flashes

### LINKED ISSUE #105: [Feat] Prefetch upload component permissions in RSC environments
Should be doable with `use` but not sure

juliusmarminge:
Closes #105

This'll blow up if `use` isn't available. Need to decide how to handle this to support everything. Might be able to use conditional exports?

### DIFF
--- a/packages/react/src/useUploadThing.ts
+++ b/packages/react/src/useUploadThing.ts
@@ -1,21 +1,32 @@
-import { useState } from "react";
+import { use, useRef, useState } from "react";
 
-import { DANGEROUS__uploadFiles } from "uploadthing/client";
+import { DANGEROUS__uploadFiles, getUtUrl } from "uploadthing/client";
 import type { ExpandedRouteConfig, FileRouter } from "uploadthing/server";
 
 import { useEvent } from "./utils/useEvent";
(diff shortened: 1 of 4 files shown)


packages/react/src/useUploadThing.ts
```diff
@@ -1,23 +1,43 @@
...
+import { use, useRef, useState } from "react";
 
-import { DANGEROUS__uploadFiles } from "uploadthing/client";
+import { DANGEROUS__uploadFiles, getUtUrl } from "uploadthing/client";
 import type { ExpandedRouteConfig, FileRouter } from "uploadthing/server";
 
 import { useEvent } from "./utils/useEvent";
 import useFetch from "./utils/useFetch";
 
+const fetchEndpointData = async (endpoint: string) => {
```
markflorkowski, line 9: Am I blind, or where is this actually using the endpoint?

juliusmarminge: oh it isn't - endpoint is filtered in the hook. can remove this param

markflorkowski (approved):

PR #126, opened 2023-06-01
fix: patch mime-types to make us edge-compat

juliusmarminge:
Leaving this here for reference:

Tried to upstream this but they refused: https://github.com/jshttp/mime-types/pull/108


patches/mime-types@2.1.35.patch
```diff
@@ -0,0 +1,17 @@
...
+@@ -13,7 +13,11 @@
+  */
+ 
+ var db = require('mime-db')
+-var extname = require('path').extname
++// var extname = require('path').extname
++const extname = function (path) {
++  let index = path.lastIndexOf(".");
++  return index < 0 ? "" : path.substr(index);
++};
```
juliusmarminge, line 14: this might not be the most sophisticated version of `extname` but i doubt it's necessary to be more complex xddd

packages/uploadthing/src/mime-types/index.ts
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * Module dependencies.
+ * @private
+ */
+
+const db = require("./db.json");
+function extname(path: string) {
+  const index = path.lastIndexOf(".");
+  return index < 0 ? "" : path.substring(index);
+}
```
juliusmarminge, line 31: this was `path.extname` before - you decide if we need a more sophisticated version of this

t3dotgg (approved):

OrJDev:
Everything looks good to me, also needs this to get the latest solid version ahead, great one.

packages/uploadthing/src/mime-types/index.ts
juliusmarminge, line 31: the original module had tests that handled dotfiles, adding support for that here too i guess

```suggestion
function extname (path) {
  var index = path.lastIndexOf('.')
  var isDotFile = index === path.lastIndexOf('/') + 1
  if (index <= 0 || isDotFile) {
    return ''
  }
  return path.substring(index)
}
```

markflorkowski:
Potential alternative option if we don't want to deal with patches -- https://www.npmjs.com/package/mime

markflorkowski (approved):

PR #134, opened 2023-06-05
fix: add back support for react 17

### LINKED ISSUE #130: [Bug]: v4.1.0 'use' is not exported from 'react'
### What happened?

Building with Next `pages/` directory with uploadthing v4.1.0 throws this:

```
Attempted import error: 'use' is not exported from 'react' (imported as 'use').

Import trace for requested module:
./node_modules/.pnpm/@uploadthing+react@4.1.0_react-dropzone@14.2.3_react@18.2.0_uploadthing@4.1.0/node_modules/@uploadthing/react/dist/index.mjs
./src/components/form.tsx
```

This is fine with v4.0.0

Not error

### Version

4.1.0

### Reproduction Steps

_No response_

### Reproduction Repository

_No response_

### Relevant log output

_No response_

### Code of Conduct

juliusmarminge:
🤷🏼

Closes #130 (although I couldn't repro it on React ^18 which it looked like OP was on, this fixes the issue all the way down to React 17)

### DIFF
--- a/packages/react/src/useUploadThing.ts
+++ b/packages/react/src/useUploadThing.ts
@@ -1,4 +1,4 @@
-import { use, useRef, useState } from "react";
+import * as React from "react";
 
 import { DANGEROUS__uploadFiles, getUtUrl } from "uploadthing/client";
 import type { ExpandedRouteConfig, FileRouter } from "uploadthing/server";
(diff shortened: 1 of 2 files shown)


packages/react/src/useUploadThing.ts
```diff
@@ -1,4 +1,4 @@
-import { use, useRef, useState } from "react";
```
juliusmarminge, line 1: this could be

```ts
import React, { useRef, useState } from "react";

...

const promiseRef = useRef<...>()

...

if (typeof React.use === "function") ...
```

if you want

markflorkowski: As dumb as it is, I do kind of like this better.

juliusmarminge: Aight changed it up

markflorkowski (approved):

PR #135, opened 2023-06-06
[feat] Allow use of mime-types in file router config

markflorkowski:
NOTE: Requires corresponding infra changes. 

To Do:
- [x] Infra changes 
- [x] Test to make sure this works and types are 
good


packages/uploadthing/src/mime-types/index.ts
```diff
@@ -71,8 +71,8 @@ function populateMaps(
   // source preference (least -> most)
   const preference = ["nginx", "apache", undefined, "iana"];
 
-  Object.keys(db).forEach(function forEachMimeType(type) {
-    const mime = db[type];
+  Object.keys(mimeDB).forEach(function forEachMimeType(type) {
+    const mime = mimeDB[type as MimeType] as unknown as MimeObject;
```
juliusmarminge, line 75: these type casts shouldn't be necessary?

packages/uploadthing/src/mime-types/index.ts
```diff
@@ -87,7 +87,10 @@ function populateMaps(
       const extension = exts[i];
 
       if (types[extension]) {
-        const from = preference.indexOf(db[types[extension]].source);
+        const from = preference.indexOf(
+          (mimeDB[types[extension] as MimeType] as unknown as MimeObject)
```
juliusmarminge, line 91: same as above - is the `mimeDB` not typed?

packages/uploadthing/src/mime-types/index.ts
markflorkowski: It got very angry when I didn't have them. Will play with it some more though

packages/uploadthing/src/utils.ts
```diff
@@ -76,6 +81,12 @@ export const getTypeFromFileName = (
     );
   }
 
+  if (allowedTypes.some((type) => type.includes("/"))) {
```
t3dotgg, line 84: Would be nice to comment what these cases are for future reference

packages/uploadthing/src/types.ts
```diff
@@ -36,13 +44,17 @@ type RouteConfig = {
...
+>;
 
 type PartialRouteConfig = Partial<
-  Record<AllowedFileType, Partial<RouteConfig>>
+  Record<AllowedFileType | MimeType, Partial<RouteConfig>>
 >;
 
-export type FileRouterInputConfig = AllowedFileType[] | PartialRouteConfig;
+export type FileRouterInputConfig =
+  | (AllowedFileType | MimeType)[]
```
t3dotgg, line 56: We should prolly define this as like `FileRouterInputKey` and re-use it

t3dotgg (approved):

PR #136, opened 2023-06-06
implement a shared pacakge that can be used for both infra and OSS 

juliusmarminge:
Closes UPL-51


juliusmarminge:
Too late to deal with TS2742 errors..................

juliusmarminge:
gaaahahahahahaha why does nothing just woooork
 
CleanShot 2023-06-09 at 00 39 59@2x

.github/workflows/ci.yaml
```diff
@@ -42,11 +46,8 @@ jobs:
       - name: Install deps (with cache)
         run: pnpm install
 
-      - name: Typecheck
-        run: pnpm typecheck
+      - name: Build, lint, test, and typecheck
```
markflorkowski, line 49: Idk if this should be in this PR, but fine 😅

.gitignore
```diff
@@ -25,7 +25,6 @@ yarn-error.log*
 .pnpm-debug.log*
 
 # local env files
-.env
```
markflorkowski, line 28: why?

examples/pagedir/.env
```diff
@@ -0,0 +1 @@
+UPLOADTHING_URL="http://localhost:3010"
```
markflorkowski, line 1: Ah, this is why you removed `.env` from the gitignore. 

I think this should probably be an `example.env`, since you don't actually want this most of the time, no?

packages/react/src/component.tsx
```diff
@@ -3,13 +3,14 @@ import { useCallback, useRef, useState } from "react";
 import type { FileWithPath } from "react-dropzone";
 import { useDropzone } from "react-dropzone";
 
+import { ExpandedRouteConfig } from "@uploadthing/shared";
```
markflorkowski, line 6: This should still be a type import, no?

packages/uploadthing/server.ts
```diff
@@ -1,6 +1,5 @@
 import { createBuilder } from "./src/upload-builder";
 
-export * from "./src/types";
-export * as utapi from "./src/helper-functions";
```
markflorkowski, line 4: I may be blind, but where did this export go?

.github/workflows/ci.yaml
juliusmarminge: I tihnk it's leftovers from when i tested a bunch of stuff and got build errors on the examples - I can remove it and make a new one

.gitignore
juliusmarminge: Cause the pagedir example is running on port 3010 and to get it to work we must override the UPLOADTHING_URL and i thought it could be commited? It's also leftovers from when i tested out the 100s of errors during the example builds though so can revert it in here

examples/pagedir/.env
juliusmarminge: Sure

packages/react/src/component.tsx
juliusmarminge: Yep - I'm not sure why the linter isn't yelling 👀

packages/uploadthing/server.ts
juliusmarminge: Ah i exported a bunch of stuff when i got the TS2742 errors and must have removed too many when reverting 😓

packages/react/src/component.tsx
```diff
@@ -199,11 +196,6 @@ export const UploadDropzone = <
 
   const ready = fileTypes.length > 0;
 
-  const getUploadButtonText = (fileTypes: string[]) => {
-    if (!(fileTypes.length > 0)) return "Loading...";
-    return `Choose File${multiple ? `(s)` : ``}`;
-  };
```
juliusmarminge, line 204: unused

packages/react/src/client-generator.test.ts
```diff
@@ -4,15 +4,6 @@ import { createUploadthing } from "uploadthing/server";
...
 import { generateReactHelpers } from "./useUploadThing";
 
-const badReqMock = {
-  headers: {
-    get(key: string) {
-      if (key === "header1") return "woohoo";
-      return null;
-    },
-  },
-} as unknown as Request;
```
juliusmarminge, line 14: unsued

.github/workflows/ci.yaml
markflorkowski: Don't worry about it

markflorkowski (approved):

PR #141, opened 2023-06-07
fix(react-UploadDropzone): clear files onClientUploadComplete

### LINKED ISSUE #140: [Bug]: UploadDropzone does not clear files after completed uplad
### What happened?

After uploading files via the UploadDropzone component, it keeps the `Upload` button on screen and the files in the component state.

The button should disappear  & clear file state

### Version

4.1.0

### Reproduction Steps

Any use of `uploadDropzone` will show this bug. Id give a full example, but I'm just making this issue so my PR for the fix has something to attach to for documentation.

### Reproduction Repository

_No response_

### Relevant log output

_No response_

### Code of Conduct

GentikSolm:
# Pull Request

## Linked GH Issues

Fixes #140

## Summary of Change

Clears `files` state onUploadComplete for react-dropzone

### DIFF
--- a/.changeset/few-snakes-itch.md
+++ b/.changeset/few-snakes-itch.md
@@ -1,0 +1,5 @@
+---
+"@uploadthing/react": patch
+---
+
+fix(react-UploadDropzone): clear files onClientUploadComplete for dropzone and upload button
--- a/packages/react/src/component.tsx
+++ b/packages/react/src/component.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useState } from "react";
+import { useCallback, useRef, useState } from "react";
 import type { FileWithPath } from "react-dropzone";
 import { useDropzone } from "react-dropzone";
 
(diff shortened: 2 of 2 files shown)


markflorkowski (commented):
I realized that the upload button suffers from a similar issue. Do you mind making corresponding changes to that component? You'll need to add state for the files.

GentikSolm:
> I realized that the upload button suffers from a similar issue. Do you mind making corresponding changes to that component? You'll need to add state for the files.

I just tested uploadButton and it seems to be fine. The upload button uploads the second you hit "select" on your systems file explorer using the `onChange` prop, so it doesnt actually ever cache the files. 

If you want i can give it the same behavior as the drop zone, where after the user selects files it will show how many you are about to upload, and wait for the user to press the button again. If we went that route, it would most likely make sense to add a 'cancel' icon / button somewhere. With dropzone, we can change which files we are uploading by just hitting the dropzone again and selecting different files. The uploadButton directly opens the selector, so users would most likely want to be able to switch which files they wanted to upload if they miss-clicked and would need a button to clear the staged files.

markflorkowski:
I just tested again, try uploading the same file twice. "

It seems like `onChange()` does not fire the second time since the value (native, not react) of the input is not cleared, and thus does not change when selecting the same file again.

GentikSolm:
Ah, got it. Working on it.

markflorkowski (approved):

PR #142, opened 2023-06-08
[fix] Make PDF uploads work.

markflorkowski:


packages/uploadthing/src/utils.ts
```diff
@@ -75,7 +75,11 @@ export const getTypeFromFileName = (
     );
   }
 
-  const type = mimeType.split("/")[0] as AllowedFileType;
+  const type = (
```
t3dotgg, line 78: This feels like it should be a function or at least commented what it's doing

t3dotgg (approved):

PR #155, opened 2023-06-11
fix: make sure running info only prints on server & solid server code doesn't leak

OrJDev:
This makes sure that the: `[UT] UploadThing dev server is now running!` will only be printed on the server, i noticed it also prints on the client side on the solid adapter, that shouldn't happen

### DIFF
--- a/packages/solid/src/index.tsx
+++ b/packages/solid/src/index.tsx
@@ -1,19 +1,4 @@
 import "../styles.css";
 
-import { createNextRouteHandler } from "uploadthing/next";
-
 export * from "./useUploadThing";
 export * from "./component";
-
-export const createServerHandler = (
-  ...args: Parameters<typeof createNextRouteHandler>
-) => {
-  const handler = createNextRouteHandler(...args);
-  const thisHandler =
-    (method: keyof typeof handler) => (event: { request: Request } | Request) =>
-      handler[method](event instanceof Request ? event : event.request);
-  return {
-    GET: thisHandler("GET"),
-    POST: thisHandler("POST"),
-  };
-};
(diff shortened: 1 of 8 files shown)


t3dotgg (approved):

packages/solid/src/index.tsx
```diff
@@ -1,19 +1,4 @@
...
-) => {
-  const handler = createNextRouteHandler(...args);
-  const thisHandler =
-    (method: keyof typeof handler) => (event: { request: Request } | Request) =>
-      handler[method](event instanceof Request ? event : event.request);
-  return {
-    GET: thisHandler("GET"),
-    POST: thisHandler("POST"),
-  };
-};
```
juliusmarminge, line 19: this is a breaking change

juliusmarminge (changes requested):
is this to be considered a security vulnerabilities so it's okay to make a breaking change in a patch release?

OrJDev:
> is this to be considered a security vulnerabilities so it's okay to make a breaking change in a patch release?

I think it should be a minor considering this fact

markflorkowski:
Yeah, I think breaking changes should always be at least a minor. Versioning is hard 😅

packages/uploadthing/src/next/core/approuter.ts
```diff
@@ -53,3 +53,16 @@ export const createNextRouteHandler = <TRouter extends FileRouter>(
 
   return { GET, POST };
 };
+
```
t3dotgg, line 56: Is this the best place to export this from? 🤔

OrJDev: technically, since both solid & app router are based on Request / Response api, yes, either that or make a new file called `node-handler.ts`

juliusmarminge: these work on edge so node isn't the best name.

OrJDev: changed it to `./src/internal/edge`

OrJDev:
Everything should be fine now

markflorkowski (approved):

juliusmarminge (approved):

PR #156, opened 2023-06-11
fix: implement f.router to prevent usage of satisfies

OrJDev:
Instead of using the `satisfies` keyword, we can now do this:

```ts
import { createUploadthing } from "uploadthing/next";

const f = createUploadthing();

export const uploadRouter = f.router({
  videoAndImage: f({
    image: {
      maxFileSize: "4MB",
      maxFileCount: 4,
    },
    video: {
      maxFileSize: "16MB",
    },
  })
    .middleware(() => ({}))
    .onUploadComplete((data) => {
      console.log("upload completed", data);
    }),

  withMdwr: f({
    image: {
      maxFileCount: 2,
      maxFileSize: "1MB",
    },
  })
    .middleware((req) => {
      const h = req.headers.get("someProperty");

      if (!h) throw new Error("someProperty is required");

      return {
        someProperty: h,
        otherProperty: "hello" as const,
      };
    })
    .onUploadComplete(({ metadata, file }) => {
      console.log("uploaded with the following metadata:", metadata);
      metadata.someProperty;
      //       ^?
      metadata.otherProperty;
      //       ^?

      console.log("files successfully uploaded:", file);
      file;
      // ^?
    }),

  withoutMdwr: f({
    image: {
      maxFileCount: 2,
      maxFileSize: "16MB",
    },
  })
    .middleware(() => {
      return { testMetadata: "lol" };
    })
    .onUploadComplete(({ metadata, file }) => {
      console.log("uploaded with the following metadata:", metadata);
      metadata;
      // ^?

      console.log("files successfully uploaded:", file);
      file;
      // ^?
    }),
});

export type OurFileRouter = typeof uploadRouter;
```

This is using a proxy under the hood.


juliusmarminge:
Why though? The router holds no information, everything is in the endpoints themselves. 

Fun fact is we want to get rid of the `t.router()` in tRPC since it doesn't do anything useful really

OrJDev:
> Why though? The router holds no information, everything is in the endpoints themselves. 
> 
> 
> 
> Fun fact is we want to get rid of the `t.router()` in tRPC since it doesn't do anything useful really

Its just a different way of writing stuff, no one forces you to use f.router it's optional, also if you don't use typescript v5 like most of people, you still get type safety work, i think its just nice having in the code base, shouldn't break anything

juliusmarminge:
> Its just a different way of writing stuff, no one forces you to use f.router it's optional, also if you don't use typescript v5 like most of people, you still get type safety work, i think its just nice having in the code base, shouldn't break anything

Supporting lower versions is nice, but it comes at a cost of maintaining more api surface and more docs etc. `satisfies` is supported in 4.9 and we're on 5.1 today - might be a bit too small window but we also haven't set a peerDep that we should adhere too (although we probably should)

OrJDev:
> > Its just a different way of writing stuff, no one forces you to use f.router it's optional, also if you don't use typescript v5 like most of people, you still get type safety work, i think its just nice having in the code base, shouldn't break anything
> 
> 
> 
> Supporting lower versions is nice, but it comes at a cost of maintaining more api surface and more docs etc. `satisfies` is supported in 4.9 and we're on 5.1 today - might be a bit too small window but we also haven't set a peerDep that we should adhere too (although we probably should)
> 
> 
> 
> imo if we add this, it should be a callout that you can do f.router if you can't use satisfies, and keep using satisfied in the docs and examples

Sure as long as it exists there, will update this pr latest

OrJDev:
> > Its just a different way of writing stuff, no one forces you to use f.router it's optional, also if you don't use typescript v5 like most of people, you still get type safety work, i think its just nice having in the code base, shouldn't break anything
> 
> Supporting lower versions is nice, but it comes at a cost of maintaining more api surface and more docs etc. `satisfies` is supported in 4.9 and we're on 5.1 today - might be a bit too small window but we also haven't set a peerDep that we should adhere too (although we probably should)
> 
> imo if we add this, it should be a callout that you can do f.router if you can't use satisfies, and keep using satisfied in the docs and examples

Done, made it clear that you need to use f.router when `satisfies` is not a thing.

OrJDev:
Resolved all conflicts and got everything up to date, hopefully we can get this merged asap as it is important to have in the core imo.

docs/src/pages/solid.mdx
```diff
@@ -121,6 +121,35 @@ export const uploadRouter = {
 export type OurFileRouter = typeof uploadRouter;
 ```
 
+#### Not using TypeScript V4.9 or above?
+
+Not using `v4.9` or above, meaning you won't get access to the `satisfies`
+keyword, therefore you should be using `f.router` to define a router.
+
+```ts copy filename="src/server/uploadthing.ts"
+import { createUploadthing } from "uploadthing/next";
```
juliusmarminge, line 130: /solid?

juliusmarminge (commented):
LGTM. Might want to add 1 or 2 tests (happy/sad path) to the builder just to make sure nothing breaks now or in the future

docs/src/pages/solid.mdx
OrJDev: The import is right, it is coming from uploadthing/next

OrJDev:
> LGTM. Might want to add 1 or 2 tests (happy/sad path) to the builder just to make sure nothing breaks now or in the future

Added tests, think should be good now.

markflorkowski:
fixed merge conflicts. I should know better than to use the web editor ever... Sorry about that 😅

docs/src/pages/nextjs/appdir.mdx
```diff
@@ -51,6 +51,35 @@ export const ourFileRouter = {
 export type OurFileRouter = typeof ourFileRouter;
 ```
 
+#### Not using TypeScript V4.9 or above?
+
+Not using `v4.9` or above, meaning you won't get access to the `satisfies`
+keyword, therefore you should be using `f.router` to define a router.
```
juliusmarminge, line 57:
```suggestion
If you're not using TypeScript 4.9 or above, you can use the `f.router` helper to define a router and get full typesafety:
```

docs/src/pages/nextjs/pagedir.mdx
```diff
@@ -69,6 +69,35 @@ export const ourFileRouter = {
 export type OurFileRouter = typeof ourFileRouter;
 ```
 
+#### Not using TypeScript V4.9 or above?
+
+Not using `v4.9` or above, meaning you won't get access to the `satisfies`
+keyword, therefore you should be using `f.router` to define a router.
```
juliusmarminge, line 75:
```suggestion
If you're not using TypeScript 4.9 or above, you can use the `f.router` helper to define a router and get full typesafety:
```

juliusmarminge (commented):
[As mentioned here](https://github.com/pingdotgg/uploadthing/pull/280#discussion_r1306619592), we might wanna refactor the docs a bit to be more focused and less code duplication, this heading shouldn't be needed in more than one place...

docs/src/pages/nextjs/appdir.mdx
```diff
@@ -51,6 +51,34 @@ export const ourFileRouter = {
 export type OurFileRouter = typeof ourFileRouter;
 ```
 
+#### Not using TypeScript V4.9 or above?
```
juliusmarminge, line 54: add `Since version xxx` note

.changeset/serious-bikes-raise.md
```diff
@@ -0,0 +1,5 @@
+---
+"uploadthing": patch
+---
+
+feat: f.router function to define uploadThing routers
```
juliusmarminge, line 5:
```suggestion
feat: introcude `f.router` helper to define routers without the `satisfies` keyword
```

juliusmarminge (approved):

t3dotgg:
Hate to push back when it's this far along, but I'm really not sure we need this. tRPC no longer has anything like this correct? 🤔

juliusmarminge:
> tRPC no longer has anything like this correct? 🤔

We do, although we could drop it since no info is stored on the router. All procedures have all the info they need.

I will say the further we wait, the less important this is. We're at 5.2 now and satisfies was introduced in 4.9, and i think it's fair to assume most projects using Uploadthing are on the newer side?

t3dotgg:
Being that `satisfies` was introduced in the last major, I am going to close this in favor of pushing for a single, centralized path with `satisfies`. I am hopeful we don't regret this.

Thank you @OrJDev for the hard work regardless, sorry I was so indecisive on this 😓

PR #157, opened 2023-06-11
feat: allow user input along with the files

juliusmarminge:
Closes UPL-50

> Note: I based this off of `shared-pkg` so the diff will be messy. Merge #136 before reviewing this one

This allows the user to pass in client-side metadata along with the files to their endpoints


packages/react/src/component.tsx
```diff
@@ -67,6 +63,13 @@ const allowedContentTextLabelGenerator = (
   return capitalizeStart(INTERNAL_doFormatting(config));
 };
 
+type inferEndpointInput<
+  TRouter extends FileRouter,
+  TEndpoint extends keyof TRouter,
+> = TRouter[TEndpoint]["_def"]["_input"] extends UnsetMarker
+  ? undefined
+  : TRouter[TEndpoint]["_def"]["_input"];
```
juliusmarminge, line 71: move to uploadthing/server (or shared)

packages/react/src/component.tsx
```diff
@@ -75,27 +78,36 @@ const allowedContentTextLabelGenerator = (
...
+  props: {
+    endpoint: TEndpoint;
+    onClientUploadComplete?: (
+      res?: Awaited<ReturnType<typeof DANGEROUS__uploadFiles>>,
+    ) => void;
+    onUploadError?: (error: Error) => void;
+  } & (inferEndpointInput<TRouter, TEndpoint> extends undefined
+    ? {}
+    : { input: inferEndpointInput<TRouter, TEndpoint> }),
+) {
```
juliusmarminge, line 94: probable extract the type out here to a shared one that each component can extend from - duplicating it 3 times now...

examples/appdir/src/server/uploadthing.ts
```diff
@@ -18,13 +20,28 @@ export const uploadRouter = {
       console.log("upload completed", data);
     }),
 
+  withInput: f(["image"])
+    .input(
+      z.object({
+        foo: z.string(),
+        now: z.date(),
```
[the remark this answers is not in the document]
markflorkowski: I can't really think of a reason a user would need to pass a date here, so I am ok with banning it. 

If we can think of one or two legit uses for it, then serialization makes sense.

packages/uploadthing/src/internal/types.ts
```diff
@@ -21,58 +21,77 @@ type ResolverOptions<TParams extends AnyParams> = {
...
+
+// Don't want to use Zod cause it's an optional dependency
+export type ParseFn<TType> = (input: unknown) => MaybePromise<TType>;
+export type ParserZodEsque<TInput, TParsedInput extends Json> = {
+  _input: TInput;
+  _output: TParsedInput; // if using .transform etc
+  parse: ParseFn<TParsedInput>;
+};
+
+// In case we add support for more parsers later
```
markflorkowski, line 35: clever. I like this.

markflorkowski (commented):
Something we may want to consider is calling `input` `metadata`, and changing the existing `metadata` to something else. I have not fully thought this through, I just find `input` to be a bit confusing.

examples/appdir/src/server/uploadthing.ts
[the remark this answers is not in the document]
juliusmarminge: let's start without then and if users requests it we add serialization of the type (or superjson for the actual values)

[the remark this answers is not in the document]
markflorkowski: sgtm!

packages/react/package.json
```diff
@@ -59,7 +59,8 @@
     "typescript": "5.1.3",
     "uploadthing": "4.1.2",
     "vitest": "^0.30.1",
-    "wait-on": "^7.0.1"
+    "wait-on": "^7.0.1",
+    "zod": "^3.21.4"
```
juliusmarminge, line 63: testing dep

t3dotgg (approved):
This is very good

packages/react/src/component.tsx
```diff
@@ -265,26 +268,3 @@ export const UploadDropzone = <
...
-      </div>
-      <div className="flex flex-col items-center justify-center gap-4">
-        <span className="text-center text-4xl font-bold">
-          {`...or using a dropzone:`}
-        </span>
-        <UploadDropzone<TRouter> {...props} />
-      </div>
-    </>
-  );
-};
```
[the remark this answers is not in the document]
markflorkowski: Huh, yeah... not sure why that existed

packages/solid/src/component.tsx
juliusmarminge, line 1: there are things in here as well as the hook that can be separated out to the shared package and used for both solid and react.

not doing that now though - just a small TODO

packages/react/src/component.tsx
```diff
@@ -264,45 +272,62 @@ export const UploadDropzone = <
...
+}
 
 export function generateComponents<TRouter extends FileRouter>() {
   return {
-    UploadButton: (
-      props: React.ComponentProps<typeof UploadButton<TRouter>>,
-    ) => {
-      return <UploadButton {...props} />;
+    UploadButton: (props: UploadthingComponentProps<TRouter>) => {
+      // @ts-expect-error - this is validated above
```
juliusmarminge, line 321: Definetely don't like that there is 5 `@ts-expect-error` in this file but it does still work as it should:

CleanShot 2023-06-15 at 22 43 35

Would love to fix this though

OrJDev:
This looks good to me, think API could be improved but it is a good start point, related https://github.com/pingdotgg/uploadthing/pull/166

juliusmarminge:
> This looks good to me, think API could be improved but it is a good start point and works perfectly fine, related https://github.com/pingdotgg/uploadthing/pull/166

Interested to see what improvements?

markflorkowski (approved):

PR #166, opened 2023-06-16
input: add an input for foo

OrJDev:


markflorkowski:
There is already an example on the `/sink` page for this. We want to try to keep the "main" examples pretty basic 👍

PR #167, opened 2023-06-16
Upload Button styling capability

Mr0Bread:
Inspired by Clerk `appearance` prop and adjusted to current internal styling approach of UT. Open for suggestions to improve this one

Main points:
1. UT component uses tailwind classes with custom prefix. I noticed that simply adding default TW classes at the end of `className` prop will not always work as it will fail to overwrite the class. In case custom classes applied by user cannot override UT's default ones, user can override them in granular way, preserving default styling
2. Added option to apply inline styles to any of the elements. In case of button, callback can be provided to apply inline styles based on current state
3. Custom loader element can be passed instead of default one

Possible additional features:
1. Ability to specify custom button text
2. Add additional places where user can specify styles based on state

I'm also not sure if structure of appearance object is good enough, but I tried to achieve following:
1. Not expose too much information at once to not make TS autocomplete oversaturated. className configuration of each element is accessible through `classes` key of every element. In case user does not need those and interact with inline styles only, I don't see any reason to present those
2. All elements can be accessed from the first level

### DIFF
--- a/packages/react/src/component.tsx
+++ b/packages/react/src/component.tsx
@@ -115,7 +115,9 @@ export function UploadButton<TRouter extends void | FileRouter = void>(props: {
         bg?: string;
         cursor?: string;
       };
-      inlineStyle?: CSSProperties | ((ready: boolean) => CSSProperties | undefined);
+      inlineStyle?:
+        | CSSProperties
+        | ((ready: boolean) => CSSProperties | undefined);
     };
     loader?: JSX.Element;
     buttonContent?: {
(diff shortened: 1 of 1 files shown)


Mr0Bread:
PR does not contain any addition to docs as such API is not confirmed

Mr0Bread:
@markflorkowski please, take a look

markflorkowski:
Sorry we did not end up merging this. 

We want to think a bit more about how we want to support styling with the best DX possible. Personally, I found Clerk's method to be very flexible, but also far too complicated most of the time. 

I appreciate you jumping on this though and prompting me to think more about what we are looking for in this feature 😊

Issue #87, opened 2023-05-22
CSS from library affects whole app


Mr0Bread:
Since library's CSS is being applied to global elements, it interferes with other styles. I've been using shadcn/ui with uploadthing and importing uploadthing's CSS removed background from button
![image](https://github.com/pingdotgg/uploadthing/assets/55754540/0c5723b0-2fb0-43ae-9794-e1ea92c0c079)
![image](https://github.com/pingdotgg/uploadthing/assets/55754540/4477c7ff-1aff-420e-a579-cac933dc0d37)

Bug or feature? Am I missing something?

markflorkowski:
This definitely seems like a bug. The uploadthing components should be using `ut-` prefixed classnames that don't interfere:

<img width="359" alt="image" src="https://github.com/pingdotgg/uploadthing/assets/11494384/57f04ad7-2f1c-41ea-883c-2ebf8e3c2ff1">

<img width="516" alt="image" src="https://github.com/pingdotgg/uploadthing/assets/11494384/77d895a6-8ca7-47c4-b6f0-ce5bfeb76c6b">

Which version of the package are you using?

Mr0Bread:
@markflorkowski 3.0.4 Both uploadthing and @uploadthing/react

Mr0Bread:
I thought that it might be because I imported `@uploadthing/react/styles.css` after my own css file, but the order does not matter

Mr0Bread:
I guess simply noting this in the docs will help other users to not face this issue. It's definitely not a problem with your package. Btw, yesterday tried uploadthing and it's great. Thanks for your hard work. Waiting for a moment when you will be open for translations contributions ;D

markflorkowski:
Added a note in the FAQ page of the docs 👍

CamTheGoblin:
If you run into this issue and your using the @layer base in your globals.css (like is done when using shadcn/ui: https://ui.shadcn.com/docs/installation) then you can make sure your @layer base is after the @tailwind base, but before the @tailwind components and @tailwind utilities. this seemed to stop the uploadthing styles from overriding everything else for me. (it seemed to specifically be overriding things that were covered in the @layer base)
```
@tailwind base;
@layer base {
   ...
}
@tailwind components;
@tailwind utilities;
 ```

cliffordkleinsr:
[...]

PR #171, opened 2023-06-17
Removed base layer from stylesheet and added missing `ut-` prefix

Mr0Bread:
- It emits global level styles that may affect whole application. It may create conflicts with app's styling (like it does with shadcn/ui #87 )
- Noticed that some internal classes were not using ut prefix
- Applied additional class to `p` tags in components to remove default margins to preserve default styling of components


markflorkowski:
closing in favor of `preflight: false` strategy in #186

PR #176, opened 2023-06-20
feat: upload progress

### LINKED ISSUE #143: [feat] upload progress
I can't find anything in the docs about getting the upload progress. I want to show upload progress on my website so the user can see it working properly

GentikSolm:
# Pull Request

## Linked GH Issues

Closes #143

## Summary of Change

Adds `onUploadProgress` option to `useUploadThing`
Adds `onUploadProgress` prop to `UploadDropzone` and `UploadButton`

https://github.com/pingdotgg/uploadthing/assets/34988548/19fc3678-9549-42ea-9152-f58f7ad19515

NOTE:
The upload progress does not currently account for UT waiting on the `await pollForFileData(presigned.key);`.
If we want it to, i can fudge the percentage math so that it leaves the last ~10% for the `await pollFor...` Another solution would be extra fields on the upload progress, that detail the exact loading state, like "files uploading" -> "Awaiting server metadata" etc. I think this may be a bit overkill, but up for discussion. 

TODO:
- [x] Update Docs
- [x] Add solid impl


packages/react/src/component.tsx
```diff
@@ -181,18 +210,24 @@ export function UploadDropzone<TRouter extends FileRouter>(
...
         setFiles([]);
         if ($props.onClientUploadComplete) {
           $props.onClientUploadComplete(res);
         }
       },
       onUploadError: $props.onUploadError,
-    },
-  );
+    });
+  const progress: number = useMemo(() => {
```
t3dotgg, line 223: This is gonna trigger a _shitload_ of re-renders, minimum 10 per item being uploaded. Probably best to have an optional callback for tracking? Maybe a specific hook w/ progress? 🤔

markflorkowski (approved):
2 re-renders without passing onUploadProgress() seems reasonable to me, testing locally is not working for me at the moment though, so I wasn't able to confirm the re-render counts. I trust you though 👍

markflorkowski:
My dev environment was just broken, nothing specific to your PR, no worries.

t3dotgg (approved):
Huge lift, looks great!

markflorkowski (approved):

PR #178, opened 2023-06-20
fix: infinite dev server

juliusmarminge:
the mimetypes package causes infinite rebuilds when running dev

we had these issues a while back and i think it's due to the top-level source files causing some form of cycles.

i refactored the repo to not do this and could now also remove the entire `ignoreWatch` thing we used as a workaround last time.

the refactoring might be a bit opinionated so open for suggestions


markflorkowski (approved):

t3dotgg (approved):
This is gonna be a painful bandaid rip but if it works fine I'm down. Make sure all the package exports behave as expected

Also we should probably get current PRs merged before this since the conflicts will suck

PR #186, opened 2023-06-26
fix: disable preflight and add reset styles inline

GentikSolm:
# Pull Request

## Linked GH Issues
N/A

## Summary of Change
Disable tailwind preflight
Reset styles only where needed
Fix style drift between react and solidjs uploadDropzone


packages/solid/src/component.tsx
```diff
@@ -200,16 +200,17 @@ export const UploadDropzone = <TRouter extends FileRouter>(
         </svg>
         <div class="ut-mt-4 ut-flex ut-text-sm ut-leading-6 ut-text-gray-600">
           <label
-            for="file-upload"
-            class="ut-relative ut-cursor-pointer ut-font-semibold ut-text-blue-600 focus-within:ut-outline-none focus-within:ut-ring-2 focus-within:ut-ring-blue-600 focus-within:ut-ring-offset-2 hover:ut-text-blue-500"
+            html-for="file-upload"
+            class="ut-relative ut-cursor-pointer ut-font-semibold  focus-within:ut-outline-none focus-within:ut-ring-2 focus-within:ut-ring-blue-600 focus-within:ut-ring-offset-2 hover:ut-text-blue-500 ut-text-blue-600"
```
markflorkowski, line 204: huh, wonder why the order changed here.

markflorkowski (approved):
This seems reasonable to me👍

requested a review from @juliusmarminge as well

.prettierignore
```diff
@@ -0,0 +1,35 @@
...
+.pnp
+.pnp.js
+coverage
+
+.changeset
+
+*.md
+*.mdx
+*.yml
+*.yaml
```
juliusmarminge, line 11: Why?

.prettierignore
```diff
@@ -0,0 +1,35 @@
...
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+.pnpm-debug.log*
+
+.env
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local
```
juliusmarminge, line 30: Again, why? (These arent that necessary but still annoying when autosave doesnt work and you're left wondering why)

packages/solid/tailwind.config.cjs
```diff
@@ -4,6 +4,9 @@ const colors = require("tailwindcss/colors");
 module.exports = {
   content: ["./src/**/*.{js,ts,jsx,tsx,mdx}"],
   prefix: "ut-",
+  corePlugins: {
+    preflight: false,
+  },
```
juliusmarminge, line 9: Whats the difference between doing this here rather than not including the `@tailwind base` in the css file?

.prettierignore
GentikSolm: me and @markflorkowski found prettier was aggressively formatting all these. I can add these back in if we want

.prettierignore
GentikSolm: I just copied over `.gitignore` which was previously being used as the `--ignore-file` for `format` in `package.json`

.prettierignore
markflorkowski: now that I thought of it a bit more, maybe just ignore `.github` and `.changeset` and don't restrict by filetypes

packages/solid/tailwind.config.cjs
GentikSolm: Did some digging, looks like `@tailwindcss/forms` is also doing some global stuff that ill need to rework. Looking at tailwind's documentation, setting `preflight: false` seems to be the preferred way to turn off basic style resets.

GentikSolm: Did some further research, `@tailwindcss/forms` isnt actually styling anything in our library, since every `input` is hidden and of `type='file'` which forms doesnt touch regardless. Removed it as a dep.

juliusmarminge (approved):

PR #189, opened 2023-06-27
feat: Handle failed uploads (UPL-60)

markflorkowski:
### To do:
- [x] infra -- add endpoint for marking failed uploads
- [x] merge #195 
- [x] server uploads
- [x] test


juliusmarminge (commented):
Server uploads?

packages/uploadthing/src/sdk/utils.ts
```diff
@@ -96,6 +105,21 @@ export const uploadFilesInternal = async (
       });
 
       if (!s3res.ok) {
+        // tell uploadthing infra server that upload failed
+        await fetch(
+          createRequestUrl({
+            url: generateUploadThingURL("/"),
```
juliusmarminge, line 111: probably want a pathname to this url no?

this will just create a url like `https://uploadthing.com/?actionType=failure&slug=__server-uploaded`, no?

packages/uploadthing/src/internal/handler.ts
```diff
@@ -223,7 +223,7 @@ export const buildRequestHandler = <
       return { status: 200 };
     }
 
-    if (!actionType || actionType !== "upload") {
+    if (!actionType || ["upload", "failure"].includes(actionType)) {
```
juliusmarminge, line 226:
```suggestion
      if (!actionType || !["upload", "failure"].includes(actionType)) {
```

?

packages/uploadthing/src/sdk/utils.ts
juliusmarminge: or are you trying to first hit the users backend which will then hit uploadthings server? seems a bit unnecessary?

markflorkowski: image

packages/uploadthing/src/client.ts
```diff
@@ -183,9 +187,25 @@ export const DANGEROUS__uploadFiles = async <TRouter extends FileRouter>(
...
+          slug: String(opts.endpoint),
+          actionType: "failure",
+        }),
+        {
+          method: "POST",
+          body: JSON.stringify({
+            fileKey: fields.key,
+          }),
+        },
+      );
```
juliusmarminge, line 203: afaik this calls your server first, and then that will call uploadthing infra? is the plan to add an `onUploadError` hook in the file route config which will make this step necessary, or can we just send the fetch directly to uploadthing?

juliusmarminge: would it maybe also make sense to throw here instead, and then in an `.allSettled` check for errors and send a single request to the server, instead of 1 per file?

markflorkowski: Yeah, I am not sure why I went the route of going through your server first tbh...

markflorkowski: Right, I did this because we don't have the API key on client.

packages/uploadthing/src/internal/handler.ts
```diff
@@ -206,15 +206,15 @@ export const buildRequestHandler = <
...
 
-    const reqBody = (await req.json()) as {
-      file: UploadedFile;
-      files: unknown;
-      metadata: Record<string, unknown>;
-      input?: Json;
-    };
-
-    if (uploadthingHook === "callback") {
+    if (uploadthingHook && uploadthingHook === "callback") {
```
juliusmarminge, line 209:
```suggestion
    if (uploadthingHook === "callback")
```

packages/uploadthing/src/internal/handler.ts
```diff
@@ -223,7 +223,7 @@ export const buildRequestHandler = <
       return { status: 200 };
     }
 
-    if (!actionType || actionType !== "upload") {
+    if (!actionType || !["upload", "failure"].includes(actionType)) {
```
juliusmarminge, line 226: Could we perhaps move this array out, mark it as const and then infer the type from it and use that in `createRequestUrl`

juliusmarminge: Also the error below is outdated

packages/uploadthing/src/internal/handler.ts
```diff
@@ -233,119 +233,169 @@ export const buildRequestHandler = <
     }
 
     try {
-      const { files, input: userInput } = reqBody as {
-        files: string[];
-        input: Json;
-      };
+      if (actionType === "upload") {
```
juliusmarminge, line 236: I'd prefer a `switch` over `else if` but not sure what we use elsewhere

packages/uploadthing/src/internal/handler.ts
```diff
@@ -233,119 +233,169 @@ export const buildRequestHandler = <
...
+        } catch (error) {
+          console.error(error);
+          return new UploadThingError({
+            code: "BAD_REQUEST",
+            message: "Invalid input",
+            cause: error,
+          });
+        }
+
+        const metadata = await uploadable._def.middleware({
```
juliusmarminge, line 257: This can throw errors

packages/uploadthing/src/internal/handler.ts
```diff
@@ -233,119 +233,169 @@ export const buildRequestHandler = <
...
+          error: new UploadThingError({
+            code: "UPLOAD_FAILED",
+            message: `Upload failed for ${fileKey}`,
+          }),
+          fileKey,
+        });
+
+        return { status: 200 };
+      } else {
+        return { message: "invalid actionType", status: 500 };
```
juliusmarminge, line 397:
```suggestion
        return new UploadThingError({ ... })
```

packages/uploadthing/src/internal/handler.ts
markflorkowski: so it should be wrapped in a try-catch?

juliusmarminge: Like it was before, and then return the UploadThing error

markflorkowski: The whole switch is wrapped in a try catch that does this? 
image

juliusmarminge: Ah, mobile diff view is hard to parse 😅

That doesnt quite make sense though does it? There are more things that can fail in that big block which would cause a sub-optimal error message?

markflorkowski: Yeah... I should really go through and clean up this logic....

packages/uploadthing/src/client.ts
```diff
@@ -33,9 +37,12 @@ function fetchWithProgress(
...
 
-const createRequestPermsUrl = (config: { url?: string; slug: string }) => {
-  const queryParams = `?actionType=upload&slug=${config.slug}`;
-
+const createRequestUrl = (config: {
+  url?: string;
+  slug: string;
+  actionType: ActionType;
+}) => {
+  const queryParams = `?actionType=${config.actionType}&slug=${config.slug}`;
```
juliusmarminge, line 45: If we're going nit I'd say use some `URL` or `URLSearchParams` here instead of a raw string but thats very nit

packages/uploadthing/src/internal/handler.ts
```diff
@@ -223,135 +224,209 @@ export const buildRequestHandler = <
...
-    if (!actionType || actionType !== "upload") {
+    if (!actionType || !VALID_ACTION_TYPES.includes(actionType)) {
       // This would either be someone spamming or the AWS webhook
       return new UploadThingError({
         code: "BAD_REQUEST",
         cause: `Invalid action type ${actionType}`,
-        message: `Expected "upload" but got "${actionType}"`,
+        message: `Expected ${VALID_ACTION_TYPES.map((x) => `"${x}"`)
+          .join(", ")
+          .replace(/,(?!.*,)/, " or")} but got "${"a"}"`,
```
juliusmarminge, line 234: Damn that regex replace is hardcore :)

packages/uploadthing/src/internal/handler.ts
```diff
@@ -223,135 +224,209 @@ export const buildRequestHandler = <
...
+        case "failure": {
+          const { fileKey } = (await req.json()) as {
+            fileKey: string;
+          };
+
+          // Tell uploadthing to mark the upload as failed
+          const uploadthingApiResponse = await fetch(
+            generateUploadThingURL("/api/failureCallback"),
+            {
+              method: "POST",
```
juliusmarminge, line 373: Formatting?

packages/uploadthing/src/internal/handler.ts
```diff
@@ -223,135 +224,209 @@ export const buildRequestHandler = <
...
+              return new UploadThingError({
+                code: "URL_GENERATION_FAILED",
+                message: "Unable to get presigned urls",
+                cause,
+              });
+            }
+          }
+
+          // Run the onUploadError callback
+          uploadable._def.onUploadError({
```
juliusmarminge, line 406: Have we documented if this can throw or not? Might be good to tell people what this function should do. If this throws and isnt wrapped in try-catch errors might be weird?

juliusmarminge (commented):
I'll review properly when i get back on Saturday but looks good!

packages/uploadthing/src/internal/handler.ts
markflorkowski: Every other way I tried of doing it seemed like too much code hahah

packages/uploadthing/src/internal/handler.ts
markflorkowski: ?

packages/uploadthing/src/client.ts
```diff
@@ -90,9 +103,10 @@ export const DANGEROUS__uploadFiles = async <TRouter extends FileRouter>(
 ) => {
   // Get presigned URL for S3 upload
   const s3ConnectionRes = await fetch(
-    createRequestPermsUrl({
+    createRequestUrl({
```
t3dotgg, line 106: nit: might be worth naming this "createS3RequestUrl" or something to make it clear what it's calling specifically

packages/uploadthing/src/internal/handler.ts
```diff
@@ -223,135 +224,222 @@ export const buildRequestHandler = <
...
+                "Content-Type": "application/json",
+                "x-uploadthing-api-key": preferredOrEnvSecret,
+                "x-uploadthing-version": UPLOADTHING_VERSION,
+              },
+            },
+          );
+
+          if (!uploadthingApiResponse.ok) {
+            console.error("[UT] failed to mark upload as failed");
+            try {
```
t3dotgg, line 387: We do this a few times, would it be worth making a `safeParseJson` helper?

t3dotgg (approved):

packages/uploadthing/src/internal/handler.ts
juliusmarminge: never mind, looked like indentation was incorrect on mobile 😅

packages/uploadthing/src/internal/handler.ts
```diff
@@ -223,135 +224,222 @@ export const buildRequestHandler = <
...
         code: "BAD_REQUEST",
         cause: `Invalid action type ${actionType}`,
-        message: `Expected "upload" but got "${actionType}"`,
+        message: `Expected ${VALID_ACTION_TYPES.map((x) => `"${x}"`)
+          .join(", ")
+          .replace(/,(?!.*,)/, " or")} but got "${"a"}"`,
       });
     }
 
     try {
```
juliusmarminge, line 238: i don't quite like this outer try-catch block. what is it that can throw that we're not catching explicitely?

packages/uploadthing/src/internal/handler.ts
```diff
@@ -223,135 +224,222 @@ export const buildRequestHandler = <
...
+            return new UploadThingError({
+              code: "INTERNAL_SERVER_ERROR",
+              message: "Failed to run onUploadError callback",
+              cause: error,
+            });
+          }
+
+          return { status: 200 };
+        }
+        default: {
```
juliusmarminge, line 429: should make sure we get typeerror if we forget handling a case in the future

```suggestion
        }
        actionType satisfies never;
        default: {
```

packages/uploadthing/src/internal/types.ts
```diff
@@ -47,6 +47,7 @@ export interface AnyParams {
   _metadata: any; // imaginary field used to bind metadata return type to an Upload resolver
   _runtime: any;
   _errorShape: any;
+  _error: any; // used for onUploadError
```
juliusmarminge, line 50: maybe `_errorFn` to properly differentiate it from `_errorShape`?

pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '6.1'
```
juliusmarminge, line 1: 👀

juliusmarminge (approved):

t3dotgg (approved):

markflorkowski:
image

t3dotgg (approved):

PR #192, opened 2023-06-28
fix(types): bizzare typescript bug

GentikSolm:
Seems to be a typescript bug, where typescript is either getting confused, or doing some sort of optimization too agressivly. I highly recomend double checking to make sure this works for others as well.
image
image

### DIFF
--- a/packages/shared/src/types.ts
+++ b/packages/shared/src/types.ts
@@ -35,7 +35,7 @@ type RouteConfig = {
 export type FileRouterInputKey = AllowedFileType | MimeType;
 
 export type ExpandedRouteConfig = Partial<
-  Record<FileRouterInputKey, RouteConfig>
+  Record<AllowedFileType | MimeType, RouteConfig>
 >;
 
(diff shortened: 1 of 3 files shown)


packages/shared/src/types.ts
```diff
@@ -35,7 +35,7 @@ type RouteConfig = {
 export type FileRouterInputKey = AllowedFileType | MimeType;
 
 export type ExpandedRouteConfig = Partial<
-  Record<FileRouterInputKey, RouteConfig>
+  {[key in FileRouterInputKey]: RouteConfig}
 >;
```
juliusmarminge, line 39: Nit but i think this is more kind on the compiler

```suggestion
export type ExpandedRouteConfig = {
  [key in FileRouterInputKey]?: RouteConfig
};

juliusmarminge: nvm that caused the same error wtf... ts is weeeeird

juliusmarminge (approved):

PR #195, opened 2023-06-28
feat: errorFormatter

### LINKED ISSUE #198: [Question]: How to get details of an error when uploading client side?
### What happened?

When attempting to upload files larger than the defined maxFileSize, an error is thrown. However, client side, the error on `onUploadError`  doesn't provide any specific details about the error itself, making it difficult to understand what went wrong.

## Current Behavior

When logging the error to the console, the following output is displayed:image

image

The error outputs above don't provide any specific information regarding the cause of the error.

It would be helpful if the error message included more details about the specific error. Currently, the details of the error can be found in an XML format within the response of the request made to https://uploadthing-prod.s3.us-west-2.amazonaws.com/.

image

image

In the provided example, the response error code "EntityTooLarge" indicates that the maxFileSize was exceeded. However, this information is currently not accessible to the client.

## Desired Solution

I would like to explore options for the client to receive more detailed information about the error that occurred during file upload. This will help in troubleshooting and providing meaningful feedback to users.

Details about the error accessible client side.

### Version

    "@uploadthing/react": "^5.0.0",
    "@uploadthing/shared": "^5.0.0",
    "uploadthing": "^5.0.0",

### Reproduction Steps

_No response_

### Reproduction Repository

_No response_

### Relevant log output

_No response_

### Code of Conduct

juliusmarminge:
Closes UPL-54
Closes #198

Might be related to #189

### DIFF
--- a/packages/shared/src/error.ts
+++ b/packages/shared/src/error.ts
@@ -24,7 +24,7 @@ function messageFromUnknown(cause: unknown) {
 
 export class UploadThingError extends Error {
   public readonly cause?: Error;
-  public readonly code;
+  public readonly code: ErrorCode;
 
   constructor(opts: {
     code: keyof typeof ERROR_CODES;
--- a/packages/shared/src/error.ts
+++ b/packages/shared/src/error.ts
@@ -1,11 +1,14 @@
 const ERROR_CODES = {
   INTERNAL_SERVER_ERROR: 500,
+  INTERNAL_CLIENT_ERROR: 500,
   FAILED_TO_SIGN: 500,
   FAILED_TO_UPLOAD: 500,
   BAD_REQUEST: 400,
   MISSING_ENV: 500,
   FILE_LIMIT_EXCEEDED: 500,
 };
+
+type ErrorCode = keyof typeof ERROR_CODES;
 
 function messageFromUnknown(cause: unknown) {
   if (typeof cause === "string") {
@@ -35,8 +38,26 @@ export class UploadThingError extends Error {
 
     this.code = opts.code;
   }
+
+  public static async from(response: Response) {
+    const message = await response.text();
+    return new UploadThingError({
+      message,
+      code: getErrorCodeFromStatusCode(response.status),
+      cause: response,
+    });
+  }
 }
 
 export function getStatusCodeFromError(error: UploadThingError) {
   return ERROR_CODES[error.code] ?? 500;
+}
+
+function getErrorCodeFromStatusCode(statusCode: number): ErrorCode {
+  for (const [code, status] of Object.entries(ERROR_CODES)) {
+    if (status === statusCode) {
+      return code as ErrorCode;
+    }
+  }
+  return "INTERNAL_SERVER_ERROR";
 }
--- a/packages/shared/src/error.ts
+++ b/packages/shared/src/error.ts
@@ -47,7 +47,8 @@ export class UploadThingError extends Error {
     if (opts.cause instanceof Error) {
       this.cause = opts.cause;
     } else if (opts.cause instanceof Response) {
-      // TODO: handle this
+      // TODO: Handle this
+      // this.cause = ;
     } else if (typeof opts.cause === "string") {
       this.cause = new Error(opts.cause);
     } else {
--- a/packages/shared/src/error.ts
+++ b/packages/shared/src/error.ts
@@ -1,3 +1,5 @@
+import type { Json } from "./types";
+
 const ERROR_CODES = {
   BAD_REQUEST: 400,
   NOT_FOUND: 404,
@@ -30,22 +32,23 @@ function messageFromUnknown(cause: unknown, fallback?: string) {
   return fallback ?? "An unknown error occurred";
 }
 
-export class UploadThingError extends Error {
+export class UploadThingError<
+  TShape extends Json = { message: string },
+> extends Error {
   public readonly cause?: Error;
   public readonly code: ErrorCode;
-  public readonly data?: any;
+  public readonly data?: TShape;
 
   constructor(opts: {
     code: keyof typeof ERROR_CODES;
     message?: string;
     cause?: unknown;
-    data?: any;
+    data?: TShape;
   }) {
     const message = opts.message ?? messageFromUnknown(opts.cause, opts.code);
 
     super(message);
     this.code = opts.code;
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
     this.data = opts.data;
 
     if (opts.cause instanceof Error) {
@@ -62,21 +65,24 @@ export class UploadThingError extends Error {
   }
 
   public static async fromResponse(response: Response) {
-    const message = await response.text();
-    let shape: unknown = undefined;
-    try {
-      shape = JSON.parse(message);
-    } catch {}
+    const json = (await response.json()) as Json;
+    const message =
+      json &&
+      typeof json === "object" &&
+      "message" in json &&
+      typeof json.message === "string"
+        ? json.message
+        : undefined;
     return new UploadThingError({
       message,
       code: getErrorTypeFromStatusCode(response.status),
       cause: response,
-      data: shape,
+      data: json,
     });
   }
 }
 
-export function getStatusCodeFromError(error: UploadThingError) {
+export function getStatusCodeFromError(error: UploadThingError<any>) {
   return ERROR_CODES[error.code] ?? 500;
 }
 
--- a/packages/shared/src/error.ts
+++ b/packages/shared/src/error.ts
@@ -1,25 +1,33 @@
 const ERROR_CODES = {
+  BAD_REQUEST: 400,
+  NOT_FOUND: 404,
+
   INTERNAL_SERVER_ERROR: 500,
   INTERNAL_CLIENT_ERROR: 500,
-  FAILED_TO_SIGN: 500,
-  FAILED_TO_UPLOAD: 500,
-  BAD_REQUEST: 400,
+  URL_GENERATION_FAILED: 500,
+  UPLOAD_FAILED: 500,
   MISSING_ENV: 500,
   FILE_LIMIT_EXCEEDED: 500,
-};
+} as const;
 
 type ErrorCode = keyof typeof ERROR_CODES;
 
-function messageFromUnknown(cause: unknown) {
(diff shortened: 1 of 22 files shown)


packages/shared/src/error.ts
```diff
@@ -0,0 +1,63 @@
+const ERROR_CODES = {
+  INTERNAL_SERVER_ERROR: 500,
+  INTERNAL_CLIENT_ERROR: 500,
+  FAILED_TO_SIGN: 500,
```
markflorkowski, line 4: `URL_GENERATION_FAILED` is probably more accurate

packages/shared/src/error.ts
```diff
@@ -0,0 +1,63 @@
...
+  FAILED_TO_UPLOAD: 500,
```
markflorkowski, line 5: I feel like I would prefer `<SOMETHING>_FAILED` over `FAILED_TO_<SOMETHING>`. 

I am not sure if my opinion here is bad practice though.

packages/shared/src/error.ts
```diff
@@ -0,0 +1,63 @@
...
+  BAD_REQUEST: 400,
+  MISSING_ENV: 500,
+  FILE_LIMIT_EXCEEDED: 500,
+};
+
+type ErrorCode = keyof typeof ERROR_CODES;
+
+function messageFromUnknown(cause: unknown) {
```
markflorkowski, line 13: explicitly return `undefined` or a generic message if one can't be resolved?

packages/shared/src/error.ts
```diff
@@ -0,0 +1,63 @@
...
+      cause: response,
+    });
+  }
+}
+
+export function getStatusCodeFromError(error: UploadThingError) {
+  return ERROR_CODES[error.code] ?? 500;
+}
+
+function getErrorCodeFromStatusCode(statusCode: number): ErrorCode {
```
markflorkowski, line 56: nit. `ErrorType` instead of `ErrorCode` -- imo code implies a numeric code.

packages/shared/src/error.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    this.code = opts.code;
+  }
+
+  public static async fromResponse(response: Response) {
+    const message = await response.text();
+    return new UploadThingError({
+      message,
+      code: getErrorCodeFromStatusCode(response.status),
+      cause: response,
```
markflorkowski, line 47: I am pretty sure `messageFromUnknown()` can't resolve this?

markflorkowski: There are a few places we are passing a fetch response as a cause, we should probably handle it?

packages/uploadthing/src/internal/handler.ts
```diff
@@ -45,7 +46,11 @@ const fileCountLimitHit = (
     const limit = routeConfig[key as FileRouterInputKey]?.maxFileCount;
     if (!limit) {
       console.error(routeConfig, key);
-      throw new Error("invalid config during file count");
+      throw new UploadThingError({
+        code: "BAD_REQUEST",
+        message: "Invalid config during file count",
+        // cause: More info here that can be hidden from the client if chosen
```
markflorkowski, line 52: would the cause here be the route config?

or 
```
expected `number`, got `undefined`
```

packages/uploadthing/src/internal/handler.ts
```diff
@@ -139,17 +148,26 @@ export const buildRequestHandler = <
...
+        message: "No slug provided",
+      });
 
     if (!preferredOrEnvSecret) {
-      throw new Error(
-        `Please set your preferred secret in ${slug} router's config or set UPLOADTHING_SECRET in your env file`,
-      );
+      return new UploadThingError({
+        code: "BAD_REQUEST",
+        message: "No secret provided",
```
markflorkowski, line 160: Should message and cause be flipped here?

packages/uploadthing/src/internal/handler.ts
```diff
@@ -139,17 +148,26 @@ export const buildRequestHandler = <
...
+        cause: `Please set your preferred secret in ${slug} router's config or set UPLOADTHING_SECRET in your env file`,
+      });
     }
 
     const uploadable = router[slug];
     if (!uploadable) {
-      return { status: 404 };
+      return new UploadThingError({
+        code: "BAD_REQUEST",
+        message: "No uploadable found for slug",
```
markflorkowski, line 169: what is an `uploadable`? Do we define that to users anywhere?

packages/uploadthing/src/internal/handler.ts
```diff
@@ -174,7 +192,11 @@ export const buildRequestHandler = <
       // This would either be someone spamming
       // or the AWS webhook
 
-      return { status: 404 };
+      return new UploadThingError({
+        code: "BAD_REQUEST",
+        message: "Invalid action type",
```
markflorkowski, line 197: Again, should message and cause be flipped here?

packages/uploadthing/src/internal/handler.ts
```diff
@@ -211,7 +241,11 @@ export const buildRequestHandler = <
 
       const limitHit = fileCountLimitHit(files, parsedConfig);
 
-      if (limitHit) throw new Error("Too many files");
+      if (limitHit)
+        return new UploadThingError({
+          code: "BAD_REQUEST",
+          message: "File limit exceeded",
```
markflorkowski, line 247: More details would be nice. 
```
expected <= ${parsedConfig.fileCount} files, received ${files.length}
```

packages/uploadthing/src/internal/handler.ts
```diff
@@ -262,8 +303,10 @@ export const buildRequestHandler = <
     } catch (e) {
       console.error("[UT] middleware failed to run");
       console.error(e);
-
-      return { status: 400, message: (e as Error).toString() };
+      return new UploadThingError({
+        code: "BAD_REQUEST",
+        cause: e,
```
markflorkowski, line 308: message?

packages/uploadthing/src/internal/next/approuter.ts
```diff
@@ -20,16 +24,27 @@ export const createNextRouteHandler = <TRouter extends FileRouter>(
...
+    if (response instanceof UploadThingError) {
+      const formattedError = errorFormatter(response);
+      return new Response(JSON.stringify(formattedError), {
+        status: getStatusCodeFromError(response),
+        headers: {
+          "x-uploadthing-version": UPLOADTHING_VERSION,
+        },
+      });
+    }
+    if (response.status !== 200) {
```
markflorkowski, line 37: `if (response.status !== 200 || !response.message)` ? 

Since this logic is actually different from what it was previously. Or we can consider the way that it was checking before incorrect, and this is better.

packages/shared/src/error.ts
juliusmarminge: > I am pretty sure messageFromUnknown() can't resolve this?
Since we're passing a message we don't need to worry about messageFromUnknown

> There are a few places we are passing a fetch response as a cause, we should probably handle it?
Yes - I have a feeling a type is wrong which is leaking unknown somewhere

markflorkowski: > Since we're passing a message we don't need to worry about messageFromUnknown

Right 🤦

packages/uploadthing/src/internal/next/pagerouter.ts
juliusmarminge, line 1: [Lines 33-41] we're sort of doing double checks now (both here and in the requestHandler). should we let these cases fall through and rely on the handler to handle all of the bad states?

CleanShot 2023-06-29 at 00 03 18@2x

markflorkowski: As in broaden the types that the handler accepts, then move this logic there?

juliusmarminge: Yea? Bad idea?

juliusmarminge: Alternatively narrow it and let each top level handler handle all cases, I feel like the in-between state that we have rn is a bit confusing?

markflorkowski: My initial reaction was that the handler shouldn't accept incorrect types, but perhaps if it just accepts `unknown`, and we actually validate the shape inside, that is better.

packages/shared/src/error.ts
```diff
@@ -0,0 +1,80 @@
...
+    cause?: unknown;
+  }) {
+    const message = opts.message ?? messageFromUnknown(opts.cause, opts.code);
+
+    super(message);
+    this.code = opts.code;
+
+    if (opts.cause instanceof Error) {
+      this.cause = opts.cause;
+    } else if (opts.cause instanceof Response) {
```
markflorkowski, line 49: How do we feel about

```ts
else if (opts.cause instanceof Response) {
      this.cause = new Error(`Response ${opts.cause.status} ${opts.cause.statusText}`);
}
```

juliusmarminge: Yea.

Alternatively we'd just have to parse the response before creating the error in the first place, so that we could actually use the error message from the server.

This is just for the `cause` though. When the response is the error, we have the `UploadthingError.fromResponse()` we can use

OrJDev:
LGTM

packages/uploadthing/src/internal/handler.ts
```diff
@@ -183,13 +183,6 @@ export const buildRequestHandler = <
         cause: `Expected uploadthingHook to be of type 'string', got '${typeof uploadthingHook}'`,
       });
     }
-    if (typeof req.body !== "string") {
-      return new UploadThingError({
-        code: "BAD_REQUEST",
-        message: "Request body must be a JSON string",
-        cause: `Expected request body to be of type 'string', got '${typeof req.body}'`,
-      });
-    }
```
juliusmarminge: we never use the `req.body` - not sure why this check was there in the first place @markflorkowski ?

markflorkowski: We do use the body, just via `.json()` -- this check is there to make sure it exists I guess. Probably should attempt parsing to make sure it is valid JSON before throwing though. 

image

juliusmarminge: Aight Yea cause this threw in cases it shouldnt have so i nuked it

markflorkowski (approved):

PR #200, opened 2023-06-29
fix(docs): sign up link in Getting Started page

rijk:


docs/src/pages/getting-started.mdx
```diff
@@ -50,7 +50,7 @@ UPLOADTHING_SECRET=... # A secret key for your app (starts with sk_live_)
 </Callout>
 
 <Callout type="warning">
-  If you don't already have a uploadthing secret key, [sign up](/create-account) and create one from
+  If you don't already have a uploadthing secret key, [sign up](https://uploadthing.com/create-account) and create one from
```
juliusmarminge, line 53: This page 404s - it should be /sign-in

juliusmarminge (changes requested):

docs/src/pages/getting-started.mdx
rijk: Ah. That explains #201 :) 

It's weird, the link pointed me to https://docs.uploadthing.com/create-account, and then I figured just to remove the subdomain and I _thought_ that pointed me to the right page. But maybe I did something else.

juliusmarminge (approved):

PR #203, opened 2023-06-29
Edited App Dir Docs

Aayush-Rajagopalan:
added "use client" to the `src/utils/uploadthing.ts` doc since server components cannot be added to client components


juliusmarminge:
This is not needed since we inject the use client directive on our react entrypoints for you.

See our example too https://github.com/pingdotgg/uploadthing/blob/main/examples/appdir/src/utils/uploadthing.ts

juliusmarminge:
https://github.com/pingdotgg/uploadthing/blob/f82f4a8b89f1685a7081e9365f69f370f56d7274/packages/react/tsup.config.ts#L9

PR #204, opened 2023-06-29
docs: destructure `req` in middleware

danielroe:
Not super familiar with SolidStart but as-is this gives a type error.

### DIFF
--- a/docs/src/pages/solid.mdx
+++ b/docs/src/pages/solid.mdx
@@ -88,7 +88,7 @@ export const uploadRouter = {
       maxFileSize: "1MB",
     },
   })
-    .middleware((req) => {
+    .middleware(({ req }) => {
       const h = req.headers.get("someProperty");
 
       if (!h) throw new Error("someProperty is required");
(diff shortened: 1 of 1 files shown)


juliusmarminge:
i thought destructuring will actually consume the signal? idk, signals are weird and i'll never see how they provide a better dx 🤷🏼

danielroe:
Then should it be `req.req`?

juliusmarminge:
> Then should it be `req.req`?

Oh right. Yes.

`req` here must be old. It's an object now with the request inside of it (since 5.0.0). I think we refer to it as `opts` in the rest

```ts
.middleware(opts => opts.req)
```

docs/src/pages/solid.mdx
```diff
@@ -88,7 +88,7 @@ export const uploadRouter = {
       maxFileSize: "1MB",
     },
   })
-    .middleware((req) => {
+    .middleware(({ req }) => {
       const h = req.headers.get("someProperty");
```
juliusmarminge, line 92:
```suggestion
    .middleware((opts) => {
      const h = opts.req.headers.get("someProperty");
```

juliusmarminge (approved):

PR #205, opened 2023-06-29
fix(changeset): add changeset for react and solid packages

juliusmarminge:
should fix the failing workflow

will force a release of the react and solid pacakges needed in order to fix the declaration bug that the current changeset indicates

CleanShot 2023-06-29 at 22 24 18

alternatively we can change the current changeset to manually bump these packages if we don't want them to always be locked to one another

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -5,7 +5,12 @@
     { "repo": "pingdotgg/uploadthing" }
   ],
   "commit": false,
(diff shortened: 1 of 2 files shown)


markflorkowski:
I don't think the react and solid packages should be locked to each other, or am I misunderstanding?

markflorkowski (approved):

PR #207, opened 2023-06-30
fix(shared): Swap order of cases in getUploadthingUrl()

markflorkowski:
We should default to the `UPLOADTHING_URL` if it is set, always.


jakeleventhal (approved):
LGTM!

Tested and confirmed issue was resolved

juliusmarminge (approved):
Makes sense to me

PR #214, opened 2023-07-11
feat: server uploads

### LINKED ISSUE #168: Question: Using on server
Thanks for a wonderful library and service. For my usecase, I'm currently generating images on the backend (using some AI stuff), which I'm then pushing to S3, i.e. the following:

```typescript
import { S3Client, PutObjectCommand } from "@aws-sdk/client-s3";

/**
 * Uploads a given remote image URL to S3 under a given key
 * @param url - URL of image to upload to S3
 */
export const uploadAvatar = async (url: string, key: string) => {
  // Get image from replicate
  const res = await fetch(url);
  const blob = await res.arrayBuffer();
  // Upload to S3
  const s3Client = new S3Client({});
  const uploadCommand = new PutObjectCommand({
    Bucket: process.env.AWS_S3_BUCKET_NAME,
    Key: key,
    Body: new Uint8Array(blob),
  });
  const expectedUrl = `https://${process.env.AWS_S3_BUCKET_NAME}.s3.${process.env.AWS_REGION}.amazonaws.com/${key}`;
  const result = await s3Client.send(uploadCommand);
  return result.$metadata.httpStatusCode == 200 ? expectedUrl : null;
};
```
Is there any easy way this library can be used like this on the backend? I kinda want to get rid of the aws-sdk library completely from my codebase

juliusmarminge:
Closes #168

### DIFF
--- a/docs/src/pages/api-reference/server.mdx
+++ b/docs/src/pages/api-reference/server.mdx
@@ -149,6 +149,46 @@ accessible here
 The UploadThing API Helper, for use ON YOUR SERVER. It's basically just a REST
 API but better
 
+### `uploadFiles`
+
+`uploadFiles` let's you upload files directly from your server, without having
+to define file routes. When uploadthing files from the server, no presigned URLs
+are used, your files are uploaded directly to S3 via our servers.
+
+This can be useful for uploading files from sever actions,
+
+```tsx
+import { utapi } from "uploadthing/server";
+
+async function uploadFiles(formData: FormData) {
+  "use server";
+  const files = fd.getAll("files");
+  const response = await utapi.uploadFiles(files);
+  //    ^? { key: string, url: string }[]
+}
+
+function MyForm() {
+  return (
+    <form action={uploadFiles}>
+      <input name="files" type="file" multiple />
+      <button type="submit">Upload</button>
+    </form>
+  );
+}
+```
+
+| Arguments | Type                       | Required | Description                       |
+| :-------- | :------------------------- | :------- | :-------------------------------- |
+| files     | FileEsque \|\| FileEsque[] | Yes      | The file(s) you want to upload    |
+| metadata  | Json                       | No       | Metadata to attach to the file(s) |
+
+```ts
+// Edge / Node < 20 friendle File interface
+interface FileEsque extends Blob {
+  name: string;
+}
+```
+
 ### `deleteFiles`
 
 `deleteFiles` takes in a fileKey or an array of fileKeys and deletes them from
--- a/docs/src/pages/api-reference/server.mdx
+++ b/docs/src/pages/api-reference/server.mdx
@@ -190,6 +190,39 @@ interface FileEsque extends Blob {
 }
 ```
 
+### `uploadFileFromUrl`
+
+`uploadFileFromUrl` is similar to `uploadFiles` as it let's you upload files
+directly from your server, without having to define file routes.
+
+When uploading files from URL, the file is first downloaded on **your** server,
+before being uploaded to S3 via our servers.
+
+```tsx
+import { utapi } from "uploadthing/server";
+
+async function uploadFile(formData: FormData) {
+  "use server";
+  const url = fd.get("url");
+  const response = await utapi.uploadFileFromUrl(url);
+  //    ^? { key: string, url: string }
+}
+
+function MyForm() {
+  return (
+    <form action={uploadFile}>
+      <input name="url" />
+      <button type="submit">Upload</button>
+    </form>
+  );
+}
+```
+
+| Arguments | Type            | Required | Description                            |
+| :-------- | :-------------- | :------- | :------------------------------------- |
+| url       | URL \|\| string | Yes      | The URL of the file you wish to upload |
+| metadata  | Json            | No       | Metadata to attach to the file(s)      |
+
 ### `deleteFiles`
 
 `deleteFiles` takes in a fileKey or an array of fileKeys and deletes them from
--- a/docs/src/pages/api-reference/server.mdx
+++ b/docs/src/pages/api-reference/server.mdx
@@ -184,7 +184,7 @@ function MyForm() {
 | metadata  | Json                       | No       | Metadata to attach to the file(s) |
 
 ```ts
-// Edge / Node < 20 friendle File interface
+// Edge / Node < 20 friendly File interface
 interface FileEsque extends Blob {
   name: string;
 }
--- a/docs/src/pages/api-reference/server.mdx
+++ b/docs/src/pages/api-reference/server.mdx
@@ -262,3 +262,25 @@ const multipleUrls = await utapi.getFileUrls([
 | Arguments | Type                 | Required | Description                                      |
 | :-------- | :------------------- | :------- | :----------------------------------------------- |
 | input     | string \|\| string[] | Yes      | The fileKey or fileKeys you want to get URLs for |
+
+### `listFiles`
+
+`listFiles` takes no input and returns an array of objects containing file ids
+and keys for all files that have been uploaded to the application your API key
+corresponds to.
+
+```ts
+import { utapi } from "uploadthing/server";
+
+const files = await utapi.listFiles();
+
+console.log(files);
+/*
+ * [
+ *  {
+ *    id: "2e0fdb64-9957-4262-8e45-f372ba903ac8",
+ *    key: "2e0fdb64-9957-4262-8e45-f372ba903ac8_image.jpg"
+ *  }
+ * ]
+ */
+```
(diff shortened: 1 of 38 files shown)


examples/appdir/next.config.js
```diff
@@ -2,6 +2,7 @@
 const nextConfig = {
   reactStrictMode: true,
   transpilePackages: ["uploadthing", "@uploadthing/react"],
+  experimental: { serverActions: true },
```
markflorkowski, line 5: We may want to include a note in the docs about this?

packages/config/eslint/index.js
```diff
@@ -29,6 +29,10 @@ const config = {
         disallowTypeAnnotations: false,
       },
     ],
+    "@typescript-eslint/no-misused-promises": [
+      2,
```
markflorkowski, line 33: should this be a number, or `warn` / `error` / etc?

markflorkowski (approved):

packages/config/eslint/index.js
juliusmarminge: 2 == 'error' - i can change it for verbosity

examples/appdir/next.config.js
juliusmarminge: Docs added.

docs/src/pages/api-reference/server.mdx
```diff
@@ -149,6 +149,80 @@ accessible here
 The UploadThing API Helper, for use ON YOUR SERVER. It's basically just a REST
 API but better
 
+### `uploadFiles`
+
+`uploadFiles` let's you upload files directly from your server, without having
+to define file routes. When uploadthing files from the server, no presigned URLs
+are used, your files are uploaded directly to S3 via our servers.
```
markflorkowski, line 156: > your files are uploaded directly to S3 via our servers

- "directly" is not really correct here
- S3 bit is unnecessary, the backing storage doesn't matter to the end user. 

I think this could be changed to:

"your files are uploaded via our servers"

docs/src/pages/api-reference/server.mdx
```diff
@@ -149,6 +149,80 @@ accessible here
...
+}
+```
+
+### `uploadFileFromUrl`
+
+`uploadFileFromUrl` is similar to `uploadFiles` as it let's you upload files
+directly from your server, without having to define file routes.
+
+When uploading files from URL, the file is first downloaded on **your** server,
+before being uploaded to S3 via our servers.
```
markflorkowski, line 199: same as above.

packages/uploadthing/src/sdk/index.ts
```diff
@@ -14,6 +16,61 @@ function getApiKeyOrThrow() {
...
+  const blob = await fileResponse.blob();
+
+  const formData = new FormData();
+  formData.append("files", blob, filename);
+  formData.append("metadata", JSON.stringify(metadata));
+
+  return uploadFilesInternal(formData, {
+    apiKey: getApiKeyOrThrow(),
+    utVersion: UPLOADTHING_VERSION,
+  }).then((files) => files[0]);
```
markflorkowski, line 71: Should this function also take an array or urls?

juliusmarminge: It can, just didnt think it was really a use case here. For direct file uploads it makes sense since you can selektiv multiple files but I didnt think the same really applied for URLs.

But i can definetely add it. Should the return type be an array if and only if an array is passed in?

markflorkowski: I can't think of a specific use case either, it was just thinking so that the API was similar for both uploadFiles and uploadFilesViaUrl.

> Should the return type be an array if and only if an array is passed in?

This makes sense to me

juliusmarminge: Done

docs/src/pages/api-reference/server.mdx
```diff
@@ -149,6 +149,84 @@ accessible here
 The UploadThing API Helper, for use ON YOUR SERVER. It's basically just a REST
 API but better
 
+### `uploadFiles`
+
+> Added in `v5.3`
+
+`uploadFiles` let's you upload files directly from your server, without having
+to define file routes. When uploadthing files from the server, no presigned URLs
```
markflorkowski, line 157: this is incorrect now.

docs/src/pages/api-reference/server.mdx
```diff
@@ -149,6 +149,84 @@ accessible here
...
+```
+
+### `uploadFilesFromUrl`
+
+> Added in `v5.3`
+
+`uploadFilesFromUrl` is similar to `uploadFiles` as it let's you upload files
+directly from your server, without having to define file routes.
+
+When uploading files from URL, the file is first downloaded on **your** server,
```
markflorkowski, line 202: may want to put this in a callout block -- since it could be costly for the user.

docs/src/pages/api-reference/server.mdx
```diff
@@ -149,6 +149,87 @@ accessible here
 The UploadThing API Helper, for use ON YOUR SERVER. It's basically just a REST
 API but better
 
+### `uploadFiles`
+
+> Added in `v5.3`
+
+`uploadFiles` let's you upload files without needing to define file routes. This
+can be useful for uploading files from sever actions, or your own API routes.
```
t3dotgg, line 157:
```suggestion
Upload files directly from your server **without** using the file router. Useful for server-side file processing, uploading from a server action, and much more.
```

docs/src/pages/api-reference/server.mdx
```diff
@@ -149,6 +149,87 @@ accessible here
...
+
+```tsx
+import { utapi } from "uploadthing/server";
+
+async function uploadFiles(formData: FormData) {
```
t3dotgg, line 162: Would be cool to have an example reading from FS, this is fine for now tho 🙏

docs/src/pages/api-reference/server.mdx
```diff
@@ -149,6 +149,87 @@ accessible here
...
+  name: string;
+}
+```
+
+### `uploadFilesFromUrl`
+
+> Added in `v5.3`
+
+`uploadFilesFromUrl` is similar to `uploadFiles` as it let's you upload files
+directly from your server, without having to define file routes.
```
t3dotgg, line 200:
```suggestion
Have a file hosted somewhere else you want to upload on UploadThing? This is the function you're looking for
```

t3dotgg: (also "lets" instead of "let's", generally the 's is for ownership)

docs/src/pages/api-reference/server.mdx
```diff
@@ -149,6 +149,87 @@ accessible here
...
+
+<Callout type="warning">
+  When uploading files from URL, the file is first downloaded on **your**
+  server, before presigned URLs are created and the file is uploaded to the
+  storage provider.
+</Callout>
+
+```tsx
+import { utapi } from "uploadthing/server";
```
t3dotgg, line 209: IMO this example would be stronger if you made it simpler, i.e.

```tsx
const fileUrl = "https://test.com/some.png";

const uploadedFile = await utapi.uploadFileFromUrl(fileUrl);
```

t3dotgg (approved):
Looks good minus docs changes, can also be a follow-up. So hyped to have this in finally 🙏

packages/uploadthing/src/sdk/utils.ts
```diff
@@ -0,0 +1,86 @@
...
+      }
+    | { error: string };
+
+  if (!res.ok || "error" in json) {
+    const message = "error" in json ? json.error : "Unknown error";
+    throw new Error(message);
+  }
+
+  // Upload each file to S3
+  const uploads = await Promise.all(
```
t3dotgg, line 52: allSettled might be better here and we can handle the error cases w/ a more elegant return maybe?

t3dotgg (commented):
Looks good minus docs changes, can also be a follow-up. So hyped to have this in finally 🙏

docs/src/pages/api-reference/server.mdx
```diff
@@ -149,6 +149,87 @@ accessible here
...
+}
+
+function MyForm() {
+  return (
+    <form action={uploadFile}>
+      <input name="url" />
+      <button type="submit">Upload</button>
+    </form>
+  );
+}
```
juliusmarminge, line 225:
```suggestion
const fileUrl = "https://test.com/some.png";

const uploadedFile = await utapi.uploadFileFromUrl(fileUrl);
```

markflorkowski (approved):

packages/uploadthing/src/sdk/utils.ts
```diff
@@ -0,0 +1,110 @@
...
+      await pollForFileData(key);
+
+      return {
+        key,
+        url: fileUrl,
+      };
+    }),
+  );
+
+  return uploads.map((upload) => {
```
t3dotgg, line 101: Ooo I do like this

t3dotgg (approved):

PR #217, opened 2023-07-13
feat: custom styling and content

Mr0Bread:
I took @juliusmarminge 's idea as a base to include UT's classes into user's TW config and made my own version of styling functionality

Created utility function `withUt` to make process of setting up UT with TailwindCSS easier for the user

I also decided to come further and reduced the amount of elements that builds up both UploadButton and UploadDropzone preserving their looks. I was able to get rid of such things like `labelContainer` and `label` and left with `label` only

Content and styling customisations were placed into separate props. Content and styling can be configured independently.

In case of styling, user can provide either string, CSSProperties object or callback returning one of the previously mentioned options
In case of content, options are: `ReactNode` or callback returning `ReactNode`

@markflorkowski I started working on this one bit before than you commented on my previous PR. Even if this one will not end up being merged, I hope it will inspire you at least in some way in finding the best approach

Examples:
image
image


juliusmarminge:
Firstly, I Appreciate the work you've put into this. However, I see some issues with this type of implementation.

1. It introduces a loooot of code. The components were so simple before and now it feels like it 10x'd in complexity.

2. As Mark touched on in your previous PR, I've always felt Clerks way of customizing to be far to flexible and complex. If I wanna go that far and customize every single thing, I'll just make my own component using the hooks.

What I'm looking for when customizing a built in component is the general theming of it, so I can make it match my brand somewhat. If I wanna have it perfect, I'll use the hook and build the component with my existing design system. 

That's the implementation I started with way back, just accepting a className prop that would let you style the colors and text styles of the button and apply other styles to the container element.

Anyway, that's just my thoughts. Not sure what direction Theo or Mark has thought of going.

Mr0Bread:
@juliusmarminge Thank you for your reply. Your vision totally makes sense.

Maybe we could bring together the best out of two worlds.

What I mean is we take from here:

- withUT function. Simple stuff nice to have
- Reduced number of elements. This one is important if we want to make component generally themable through single `className`. In my implementation, I removed `span` from inside of a button, but it is still wrapped into container. In addition, there is allowed content text. Even if we apply className directly to button, user won't be able to do anything about allowed content text

In case of dropzone the question seems to be a bit more tricky since it has more elements and there is no like main single visual element to style in comparison to button

juliusmarminge:
> Maybe we could bring together the best out of two worlds.

Sounds good to me. The button should only require a single className prop.

For the dropzone, I'm not so sure. Maybe then `appearance` is the best, and then have a few elements you can apply classes to (or inline-styles too perhaps) 🤔

Mr0Bread:
> The button should only require a single className prop.

As far as I understand, `className` prop will be applied to `label` which represents button. There are two other elements that will be left inaccessible: `div > p`. The ones for allowed content. It won't be even possible to change the text color of it

Firstly, that I would suggest is removing `p` node and rendering text right inside of `div`. Something like that
```jsx
<div
  className="h-[1.25rem] m-0 text-xs leading-5 text-gray-600"
>
  { fileTypes && allowedContentTextLabelGenerator(permittedFileInfo?.config) }
</div>
```

Secondly, we can either add [custom variant](https://tailwindcss.com/docs/plugins#static-variants) to tailwind config through `withUt` function or simply leave it to user to use [arbitrary variants](https://tailwindcss.com/docs/adding-custom-styles#arbitrary-variants)

In first case we declare simple plugin that adds custom variant. For example, `allowed-content` that maps to `& + div`. Such selector will select next sibling(allowed content) and will allow to apply any custom styles to it. From user's perspective it will look like this
```jsx
<UploadButton
  className="text-red-300 allowed-content:text-orange-400"
/>
```

Allowed content element already has applied styles but since variant uses additional selector, specificity of it should be enough to override default styles

In second case styling allowed content would look like that from user's perspective
```jsx
<UploadButton
  className="text-red-300 [&+div]:text-orange-400"
/>
```

As you understand, second approach does not require custom plugin to be registered but in this case user has to understand arbitrary variants as well as component's layout (even though it's rather simple)

---

Regarding dropzone

It's possible to simplify the layout. We can remove `p` node and render allowed content text right inside of div. Also I see that upload `button` is wrapped into `div`. I believe we can render `button` straight away, without additional container

The less nodes there are, less appearance properties we need to support

---

Regarding `content` and `render` props

In case of `render`, as far as I can guess, you would like to not have it and simply rely on user's ability to build their own components using UT's hooks. Am I getting it correctly?

What about content customisation with `content` prop? What if user wants to simply change text of button from `Upload file(s)` to, say, `Upload` or `Upload document`? Is it good idea to force user to build custom component to simply change the text?

---

Regarding the capability to pass callbacks

What is your stance on this one? 

I implemented those with the purpose of changing state-specific styles and contents. For example, color of disabled button or color of text while button is in uploading state. Without those we need some workarounds

What can I think of right now in terms of styling at least is that it's possible to apply data attributes to elements based on their state and then use special [`data` class](https://tailwindcss.com/docs/hover-focus-and-other-states#data-attributes). 
For instance, button may have `data-ut-state` attribute that can be in three different states: readying, ready and uploading. To style both the default and loading states of upload button in dropzone user would do something like this:
```jsx
<UploadDropzone
  appearance={{
    button: "text-orange-400 data-[ut-state=uploading]:text-orange-400/50"
  }}
/>
```
To make it even easier for user, we can introduce custom classes so that user does not need to write arbitrary variant every time. For example, `data-ut-readying`, `data-ut-ready` and `data-ut-uploading`
```jsx
<UploadDropzone
  appearance={{
    button: "text-orange-400 data-ut-uploading:text-orange-400/50"
  }}
/>
```

Or we can go even further and create custom variants to remove `data-` prefix from classes. Then it would look like the following
```jsx
<UploadDropzone
  appearance={{
    button: "text-orange-400 ut-uploading:text-orange-400/50"
  }}
/>
```

In terms of content I don't see any workaround yet

juliusmarminge:
I think you've captured a tonne of valuable insights here, and my 'single className prop' seems to be outdated since the component has gotten more complex since I did my initial PR for this.

Some key takeaways:

- simplify the layout, the fewer nodes to style the easier it is to both maintain the components as well as from a user standpoint grok what they need to do

- for the remaining items, an `appearance` object might be the best option actually? Although your "workarounds" sounds pretty cool as well.

- different states, is there ever a case where the component is disabled but not uploading? If not you could just use the `disabled:` variant. If yes, data attributes are pretty widely used (e.g. Radix)

- customizing content, maybe allow a prop like
  ```ts
  content: ReactNode | (permittedFileInfo) => ReactNode
  ```

Mr0Bread:
Hey, @juliusmarminge 

I made improvements to UploadButton and UploadDropze

I simplified layout of both of them to what I believe is minimum

So, I added a `className` prop and applied it to the outermost element - container. I think applying it to the outermost element makes most sense since CSS is about applying styles to element and it's descendant

To provide a way of styling the descendants in TW way, I added new variants: `ut-button` , `ut-allowed-content`, `ut-label` and `ut-upload-icon`. They allow to apply styles to descending elements of container. `ut-label` and `ut-upload-icon` are specific to UploadDropzone. Those variants select node by additional data attribute. It's done to have single `ut-button` for both UploadDropzone and UploadButton where button itself is represented using different HTML tags: `button` and `label` respectively

Additionally, to provide a way of applying styling based on state, I added data attribute to all the elements and extended TW config with three new data classes: `data-ut-readying`, `data-ut-ready` and `data-ut-uploading`.

The new variants and data classes can be combined so every element can be styled based on current state of component
```jsx
className="border rounded p-3 ut-button:bg-red-400 ut-allowed-content:text-lg ut-button:data-ut-readying:bg-orange-300"
```

If user does not extend their TW config with new helper function, they can use arbitrary variants and data classes. If user does not use TW at all, default CSS with classes is absolutely capable of doing the same thing.

Please, let me know what you think of such approach. Basically, in terms of styling through classes, it makes `appearance` prop mostly useless. Only thing is that `appearance` also allows to apply inline styles but it's a question whether it needs to be supported at all in UT or not. Also, callbacks of `appearance` prop has more access to internal state of component giving more freedom but it's a question if such freedom is needed in the first place

---

Regarding what you said about content prop.

It's not possible to make it like you suggested if we want to allow user to customise both button text and allowed content text since those two are separate nodes. Let me know if I'm missing something

Mr0Bread:
By the way, UploadButton has three different states while button from UploadDropzone only two of them since it is being rendered only after at least one file had been added. I added `disable` prop to button of UploadDropzone but also applied new data attribute for consistency

examples/appdir/tailwind.config.js
```diff
@@ -1,5 +1,7 @@
+const { withUt } = require("./node_modules/@uploadthing/react/src/tw");
```
juliusmarminge, line 1: this should probably be a proper export

Mr0Bread: I guess you meant import

And I tried but it looks like TW config does not support conventional imports with absolute paths. Only require with relative import works for me

juliusmarminge: Well even just `require("uploadthing/tw")`. It should be supported since you can require plugins that way?

Mr0Bread: When trying `require("@uploadthing/react/tw")` getting following error:
image
Even though package.json has this subpath defined
image
I checked Tailwind's [package.json](https://github.com/tailwindlabs/tailwindcss/blob/master/package.json) and as far as I can see they do not specify `exports` property at all.

 I will try to move Tailwind utility to separate package `@uploadthing/tw` (that's a good thing anyway) and play around with build configuration to make it work properly

And what do you mean by

> And `import` for ts configs would be great

Didn't get what you want me to do

Mr0Bread: I was able to achieve proper import experience in tailwind.config
image

packages/react/src/component.tsx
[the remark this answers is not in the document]
Mr0Bread: Yep, sounds legit. Will do

packages/react/src/component.tsx
```diff
@@ -150,42 +266,95 @@ export function UploadButton<TRouter extends FileRouter>(
...
   return (
-    <div className="ut-flex ut-flex-col ut-items-center ut-justify-center ut-gap-1">
+    <div
+      className={twMerge(
+        "flex flex-col items-center justify-center gap-1",
+        $props.className,
+        styleFieldToClassName($props.appearance?.container, styleFieldArg),
+      )}
+      style={styleFieldToCssObject($props.appearance?.container, styleFieldArg)}
+      data-ut-state={getUtState()}
```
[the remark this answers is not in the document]
Mr0Bread: Yeah, I guess we don't, will adjust

packages/react/src/components/button.tsx
```diff
@@ -0,0 +1,178 @@
...
+    );
+
+    const { fileTypes, multiple } = generatePermittedFileTypes(
+        permittedFileInfo?.config,
+    );
+
+    const ready = fileTypes.length > 0;
+
+    const getUploadButtonText = (fileTypes: string[]) => {
+        if (!(fileTypes.length > 0)) return "Loading...";
```
juliusmarminge, line 82: gotta love this logic

Mr0Bread: FYI, it was like that before me, but I can refactor ;D

examples/appdir/tailwind.config.js
juliusmarminge: I fixed the bundling issues in https://github.com/pingdotgg/uploadthing/pull/217/commits/ead7636a7d381cb941dbc784821e200dae1e9cfe

Mr0Bread:
@juliusmarminge I see that you pushed some stuff. I pulled latest branch and the error with tw config and importing file is back for me locally

Btw, why not having separate package for tw config? Yeah, it's small, but I don't think we should have separate `withUt` inside of each framework/library specific package (like `@uploadthing/react` in current case)

examples/appdir/tailwind.config.js
Mr0Bread: Thanks for that

juliusmarminge:
> Btw, why not having separate package for tw config? Yeah, it's small, but I don't think we should have separate withUt inside of each framework/library specific package (like @uploadthing/react in current case)

Mostly to keep the amount of stuff users need to install to a minimum.

packages/react/src/components/button.tsx
```diff
@@ -43,136 +49,120 @@ export type UploadButtonProps<TRouter extends FileRouter> =
...
-                <input
-                    {...getInputProps()}
-                />
-                {isUploading
-                    ? spinnerFieldToElement(
-                        $props.appearance?.buttonSpinner,
-                        styleFieldArg,
-                    ) || <Spinner />
-                    : contentFieldToContent($props.content?.button, styleFieldArg) ||
-                    getUploadButtonText(fileTypes)}
```
juliusmarminge: i simplified this and removed the `buttonSpinner` prop. I don't think it makes sense to still control the content even if the user has overridden it.

juliusmarminge:
We need some docs for this.

- the anatomy, what element is which? perhaps some illustration?
  CleanShot 2023-07-20 at 00 08 13@2x

- tabbed view for tailwind / no tailwind
  - with tailwind: 
    - the withUt helper in config
    - all styles can be applied as class names with the variants (ut-button: ut-allowed-content etc) and states (data-[state=uploading]: etc)
  - without: 
    - **mandatory** import of our stylesheet
    - inline css objects

- ...

Up for doing it @Mr0Bread ?

packages/react/src/components/button.tsx
Mr0Bread: Fair enough, thank you

Mr0Bread:
> we could always have an internal package with the source code and then the ui lib packages can export from that package, but i don't think it's the best UX to have them install 3 packages just to use uploadthing

If internal package will be a dependency of ui lib packages, then the user won't need to install tw package separately. It will come out of the box

On the other hand `uploadthing/tailwind` is definitely a better entrypoint. I would say that I like this one more. One file, one new export point, ui lib agnostic

Mr0Bread:
> Up for doing it @Mr0Bread ? I think @markflorkowski is on board with this approach, but perhaps a confirmatino before putting time into 

Yep, for sure. Can start working on it once approach is approved

markflorkowski:
Can confirm, I am on board with this approach 👍

Mr0Bread:
@juliusmarminge Should the entrypoint to docs about custom styling be a top-level link or should it be placed somewhere else? Even though it's currently implemented for React components only, in the future it won't be React-only feature
image

Mr0Bread:
First version of docs have been committed

Basically, structure is following:
- Anatomy of components with screenshots
- Overall documentation of theming props: their simplified interfaces, how they work, where apply values
- `How to theme` section: contains three subsections regarding TailwindCSS, custom classes and inline styles. Each contain relevant information and basic examples
- `Content customisation` section is dedicated to `content` prop. Includes basic information, interfaces, basic example

@juliusmarminge Please, take a look. Probably, you will find what can be improved

docs/src/pages/theming.mdx
```diff
@@ -0,0 +1,430 @@
...
+# Theming
+
+## Components anatomy
+
+### UploadButton
+
+Simplified component structure:
+
+```jsx
+<div data-state={getUtState()} className={className}>
```
juliusmarminge, line 12: What is `getUtState`? Not something users should care about

docs/src/pages/theming.mdx
```diff
@@ -0,0 +1,430 @@
...
+  </button>
+</div>
+```
+
+<Callout type="info">
+  UploadDropzone consists of five themable elements: container, upload icon,
+  label, button and allowed content.
+</Callout>
+
+<Callout type="warning">
```
juliusmarminge, line 73: It is not quite clear to me what this means, maybe we should refactor the components to make them have the same anatomy?

docs/src/pages/theming.mdx
```diff
@@ -0,0 +1,430 @@
...
+
+```ts
+type UploadButtonProps = {
+  /* rest of props */
+  appearance?: {
+    container?: StyleField;
+    button?: StyleField;
+    allowedContent?: StyleField;
+  };
+};
```
juliusmarminge, line 118: Might be clearer to have `interface StyleField {}` and some short inline comments

docs/src/pages/theming.mdx
```diff
@@ -0,0 +1,430 @@
...
+    allowedContent?: StyleField;
+    button?: StyleField;
+  };
+};
+```
+
+The types of callback arguments are following:
+
+```ts
+type ButtonCallbackArguments = {
```
juliusmarminge, line 135: Do we really need this callback? Is the data attributes not enough?

docs/src/pages/theming.mdx
```diff
@@ -0,0 +1,430 @@
...
+  isUploading: boolean;
+  uploadProgress: number;
+  fileTypes: string[];
+  isDragActive: boolean;
+};
+```
+
+## How to theme
+
+### With TailwindCSS
```
juliusmarminge, line 153: Might make sense to use `<Tabs>` here. There's already quite a lot of text and scrolling necessary to get where you wanna be.

docs/src/pages/theming.mdx
```diff
@@ -0,0 +1,430 @@
...
+
+#### Configuring TailwindCSS
+
+To leverage best developer experience, we strongly recommend extending your
+TailwindCSS config using our utility function `withUt`
+
+<Callout type="warning">
```
juliusmarminge, line 160: Could be rephrased a bit. Don't think it needs to be a warning either

docs/src/pages/theming.mdx
```diff
@@ -0,0 +1,430 @@
...
+
+```js filename="tailwind.config.js"
+// @ts-check
+const { withUt } = require("@uploadthing/react/tw");
+
+/** @type {import('tailwindcss').Config} */
+module.exports = withUt({
+  // your config goes here
+});
+```
```
juliusmarminge, line 183: A language picker in top right would be nice instead of showing both, although out of scope for this PR

docs/src/pages/theming.mdx
```diff
@@ -0,0 +1,430 @@
...
+
+<Callout type="warning">
```
juliusmarminge, line 185: This shouldn't be necessary if we rephrase the one above I think

docs/src/pages/theming.mdx
```diff
@@ -0,0 +1,430 @@
...
+#### `className` prop
+
+`className` prop accepts any classes so you can pass there anything you like.
+When it comes to TailwindCSS, to make styling easier, `withUt` adds few utility
+classes and variants for you to leverage.
+
+Tailwind variants that allow applying styles to internal elements and their
+arbitrary variant alternatives:
+
+| Custom variant       |                        Arbitrary variant |
```
juliusmarminge, line 205: Why should users care about this table?

docs/src/pages/theming.mdx
```diff
@@ -0,0 +1,430 @@
...
+These variants and classes can be used in conjunction with each other to make
+component match your design in the exact way you want.
+
+```jsx
+<UploadButton
+    className="bg-blue-500 ut-button:bg-cyan-500 ut-button:ut-readying:bg-cyan-500/50"
+               ^           ^ applied to button   ^ applied to button when loading config
+               | applied to container
+
+/>
```
juliusmarminge, line 230: Very nice. We could also render this component beneath to show the result

docs/src/pages/theming.mdx
```diff
@@ -0,0 +1,430 @@
...
+#### `appearance` prop
+
+If you need, you can pass TailwindCSS classes directly to specific elements of
+component or provide a callback that will be called with current state of the
+component and will return a string
+
+```jsx
+<UploadButton
+  appearance={{
+    button({ ready, isUploading }) {
```
juliusmarminge, line 250: Why should I use the callback here instead of tailwind variants?

juliusmarminge (commented):
Some brief thoughts

docs/src/pages/theming.mdx
Mr0Bread: Then probably makes sense to replace it with inline comment saying something like `current state` or `readying, ready or uploading`(to specify values that can be here)

docs/src/pages/theming.mdx
Mr0Bread: Okay, will add

docs/src/pages/theming.mdx
Mr0Bread: While data attributes are enough when it comes to styling through classes, callback makes it possible to apply state-specific styles with inline styles

docs/src/pages/theming.mdx
Mr0Bread: I intentionally didn't use tabs to make use of headings that builds up navigation on the right. It makes it possible to quickly find what you need, click on it and page will scroll to the heading. 

Do you think it's an overkill in this case and tabs should be enough?

docs/src/pages/theming.mdx
Mr0Bread: I thought it might be useful for particular users to know what arbitrary variants can be used instead of custom classes and variants. Though thinking about it now, I guess components anatomy should be enough to understand how to target children of component's container

docs/src/pages/theming.mdx
Mr0Bread: Makes sense, will do

docs/src/pages/theming.mdx
Mr0Bread: Regarding what that means: Visually buttons in both UploadButton and UploadDropzone are the similar. On the markup level they are defined using different tags. In case of `UploadButton` it's `label` because it's wrapping hidden file input. To make styling through data attributes(and custom data classes from TW specifically) work in the same way among both `UploadButton` and `UploadDropzone` I introduced `data-ut-element` data attribute. So if `UploadButton` has some styles applied to button element, for example, `ut-button:text-cyan-500`, it can be just copy pasted to `UploadDropzone` and those styles will be applied to the button of `UploadDropzone`. Hope my explanation is clear in this case

Regarding anatomy refactoring: I guess we kind of can do that. We either need to replace `label` with `button` inside of `UploadButton` or `button` with `label` inside of `UploadDropzone`. Maybe there is also another option I'm missing. What do you think?

docs/src/pages/theming.mdx
Mr0Bread: How would you rephrase it? Give me at least some direction

docs/src/pages/theming.mdx
Mr0Bread: Doesn't look like it's something available in Nextra out of the box. If you know open source project that implemented such thing in a proper way, I could make use of their code ;D

docs/src/pages/theming.mdx
Mr0Bread: Would screenshots be enough or you want to render the component within the markdown?

docs/src/pages/theming.mdx
Mr0Bread: Well, it would definitely make more sense for a user to leverage TW variants instead of callback for state-specific styling. I guess I can remove usage of callback in this particular example to avoid promotion of not the best practices

docs/src/pages/theming.mdx
markflorkowski: I would probably pull it out of the callout, and add it to the above paragraph as something like:

> This utility function adds additional classes and variants used to style the uploadthing components. If you choose not to use it, be sure to import the default stylesheet into your app.

docs/src/pages/theming.mdx
juliusmarminge: > Would screenshots be enough or you want to render the component within the markdown?

Adding @uploadthing/react as a dep and just render it would probably be the best.

docs/src/pages/theming.mdx
```diff
@@ -247,11 +242,6 @@ component and will return a string
 ```jsx
 <UploadButton
   appearance={{
-    button({ ready, isUploading }) {
-      return `${ready ? "bg-green-500" : "bg-red-500"} ${
-        isUploading ? "cursor-not-allowed" : ""
-      } after:bg-orange-400`;
-    },
```
juliusmarminge: i didn't mean remove this part, just migrate it to use the tailwind variants

```ts
    button: "after:bg-orange-400 bg-red-500 ut-ready:bg-green-500 ut-uploading:cursor-not-allowed"
```

docs/src/pages/theming.mdx
Mr0Bread: @juliusmarminge But doesn't FE component require API endpoint to be present to pull configuration from? I mean I'm not sure If I can define API endpoint in Nextra

juliusmarminge: Nextra is just Nextjs, so you could.

Mr0Bread:
@juliusmarminge I just pushed several updates to docs. As you may notice, I simply written all visual examples manually. It allowed me to show different states of components in static way.

I also tried to improve visual examples so they look kinda ok (not ugly at least). You might want to look yourself when you got time for that

docs/src/pages/theming.mdx
```diff
@@ -0,0 +1,956 @@
...
+    disabled={isUploading}
+  >
+    /button content goes here/
+  </button>
+</div>
+```
+
+<Callout type="info">
+  UploadDropzone consists of five themable elements: container, upload icon,
+  label, button and allowed content.
```
juliusmarminge, line 71:
```suggestion
  UploadDropzone consists of five themable elements: `container`, `upload icon`,
  `label`, `button` and `allowed content`.
```

docs/src/pages/theming.mdx
```diff
@@ -0,0 +1,956 @@
...
+  </label>
+  <div data-state={/* "ready" | "readying" | "uploading" */} data-ut-element="allowed-content">
+    /allowed content text goes here/
+  </div>
+</div>
+```
+
+<Callout type="info">
+  UploadButton consists of three themable elements: container, button and
+  allowed content.
```
juliusmarminge, line 26:
```suggestion
  UploadDropzone consists of five themable elements: `container`, `upload icon`,
  `label`, `button` and `allowed content`.
```

docs/src/pages/theming.mdx
```diff
@@ -0,0 +1,956 @@
...
+  style={{ marginTop: "1rem", borderRadius: "0.5rem" }}
+/>
+
+## Theming props
+
+### `className`
+
+Both UploadButton and UploadDropzone accept `className` prop. It allows you to
+pass any additional classes to the component. All classes that are being passed
+through this prop are going to be applied to the outermost element - container
```
juliusmarminge, line 95:
```suggestion
Both UploadButton and UploadDropzone accept `className` prop. It allows you to
pass any additional classes to the component. All classes that are being passed
through this prop are going to be applied to the outermost element - container.
```

docs/src/pages/theming.mdx
```diff
@@ -0,0 +1,977 @@
...
+    <input className="sr-only" disabled />
+    <p>Ready</p>
+  </label>
+  <div
+    className="h-[1.25rem] text-xs leading-5 text-white"
+    data-ut-element="allowed-content"
+  >
+    <p>Allowed content</p>
+  </div>
+</div>
```
juliusmarminge, line 256: Can we not setup a mock endpoint & router and use the components from the react lib? Will make it more future proof

juliusmarminge (commented):
Looking reallllly good. The custom html buttons are a bit sketchy imo, would be better with the proper components.

Everything else is looking solid to me!

docs/src/pages/theming.mdx
Mr0Bread: How to handle demonstration of state-specific styles in this case? I mean `readying` state is available only during quite short amount of time after component mount. State `uploading` requires actually uploading files and also visible for short amount of time

juliusmarminge: hmm - we could add a `__internal-state` prop or something to control the state (doesn't even have to be exposed as a type) and use that?

```mdx
<UploadButton __internal_state="ready" endpoint="foobar" />
<UploadButton __internal_state="readying" endpoint="foobar" />
<UploadButton __internal_state="uploading" endpoint="foobar" />
```

Mr0Bread: I was thinking about something like that as well. I also think that if that state is used we actually should make components non interactive. I mean that we don't actually want users be able to upload anything. Also we don't need to actually fetch config and therefore we don't need exposing an API route. What do you think?

Mr0Bread:
I also want to know about theming functionality for SolidJS package. This PR contains improvements for React only. Should Solid package be covered in this PR or another one?

juliusmarminge:
I'd say just add flags enough to make the documentation render the components with the correct states, no need to overdo it.

Also, don't worry about tests or the Solid pacakge for now. They aren't versioned together so it's fine to just release the React version and do the Solid one as a follow-up

Mr0Bread:
@juliusmarminge Hey, I reverted changes to useUploadthing hook and removed isEnabled flag. I also decided not to add new  api route since it is redundant. Components on `Theming` page are fully controlled and therefore they do not require any config to be displayed properly. They perform requests but they fail (since endpoint does not exist)

docs/package.json
```diff
@@ -19,6 +20,10 @@
   },
   "devDependencies": {
     "@types/react": "18.2.8",
-    "@types/react-dom": "18.2.4"
+    "@types/react-dom": "18.2.4",
+    "autoprefixer": "10.4.14",
+    "postcss": "8.4.22",
+    "tailwind-merge": "^1.13.2",
```
juliusmarminge, line 26: This is a prod dependency, although I can't find where you're using it so should probably just be removed

examples/appdir/tailwind.config.ts
```diff
@@ -1,5 +1,7 @@
+import { withUt } from "@uploadthing/react/tw";
+
 /** @type {import('tailwindcss').Config} */
```
juliusmarminge, line 3: Typehint not necessary with the helper function 😉

packages/react/package.json
```diff
@@ -12,6 +12,11 @@
     "./hooks": {
       "types": "./dist/hooks.d.ts",
       "import": "./dist/hooks.mjs"
+    },
+    "./tw": {
```
juliusmarminge, line 16: Did we wanna move this export to the main package?

packages/react/src/components/shared.tsx
```diff
@@ -0,0 +1,81 @@
...
+    }
+};
+
+export const allowedContentTextLabelGenerator = (
+    config?: ExpandedRouteConfig,
+): string => {
+    return capitalizeStart(INTERNAL_doFormatting(config));
+};
+
+export const progressHeights: { [key: number]: string } = {
```
juliusmarminge, line 53: widths?

packages/react/src/tw.ts
```diff
@@ -0,0 +1,30 @@
+import type { Config } from "tailwindcss";
+import plugin from "tailwindcss/plugin";
+
+export function withUt(twConfig: Config) {
+  // content can be an object. should add logic for that as well
```
juliusmarminge, line 5: Never seen this 👀👀

packages/react/src/types.ts
```diff
@@ -0,0 +1,23 @@
...
+  inferErrorShape,
+} from "uploadthing/server";
+
+export type UploadthingComponentProps<TRouter extends FileRouter> = {
+  [TEndpoint in keyof TRouter]: {
+    endpoint: TEndpoint;
+
+    onUploadProgress?: (progress: number) => void;
+    onClientUploadComplete?: (
+      res?: Awaited<ReturnType<UploadFileType<TRouter>>>,
```
juliusmarminge, line 15: This is wrong, should sync your fork to get the latest here

pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.1'
+lockfileVersion: '6.0'
```
juliusmarminge, line 1: 👀👀👀

juliusmarminge (commented):
Looks mostly good. Great job on this!

cc @markflorkowski wanna take a look??

examples/appdir/src/app/page.tsx
```diff
@@ -20,6 +20,31 @@ export default function Home() {
...
+          //   button({ ready, isUploading }) {
+          //     return `${ready ? "bg-green-500" : "bg-red-500"} ${isUploading ? "cursor-not-allowed" : ""
+          //       } after:bg-orange-400`;
+          //   },
+          //   container: "p-4 border rounded border-2 border-cyan-300",
+          //   allowedContent:
+          //     "h-8 flex flex-col items-center justify-center bg-red-100 p-3 rounded",
+          // }}
+
+          // Uncomment this to see custom content in action
```
markflorkowski, line 34: Not commented 😅

packages/react/package.json
markflorkowski: Yes, I think we should. Since it is not react-specific.

packages/react/src/components/shared.tsx
markflorkowski: amounts?

packages/react/src/utils/styles.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export const styleFieldToClassName = <T>(
+  styleField: StyleField<T> | undefined,
+  args: T,
+) => {
+  if (typeof styleField === "string") return styleField;
+  if (typeof styleField === "function") {
+    const result = styleField(args);
+
+    if (typeof result === "string") return result;
```
markflorkowski, line 19: may want to error/warn if we don't get a string at this point?

packages/react/src/utils/styles.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export const styleFieldToCssObject = <T>(
+  styleField: StyleField<T> | undefined,
+  args: T,
+) => {
+  if (typeof styleField === "object") return styleField;
+  if (typeof styleField === "function") {
+    const result = styleField(args);
+
+    if (typeof result === "object") return result;
```
markflorkowski, line 33: same as above.

packages/react/src/utils/styles.ts
```diff
@@ -0,0 +1,50 @@
...
+export const contentFieldToContent = <T>(
+  contentField: ContentField<T> | undefined,
+  arg: T,
+) => {
+  if (!contentField) return null;
+  if (typeof contentField !== "function") return contentField;
+  if (typeof contentField === "function") {
+    const result = contentField(arg);
+
+    return result;
```
markflorkowski, line 48: no check on this one?

markflorkowski (commented):
Agreed this is looking great, I left a few comments, but really mostly nits. 

I want to digest this for a bit and give it a re-read tomorrow as well.

Thanks again for the hard work on this!

packages/react/src/components/shared.tsx
Mr0Bread: Since it maps to width property of `after` pseudo-element, I would say `progressWidths` is more suitable

packages/react/src/utils/styles.ts
Mr0Bread: The callback always return `ReactNode`. In case of StyleField's callbacks can return either string or `CSSProperties` object

packages/react/src/utils/styles.ts
Mr0Bread: Considering current logic it will always create warnings. Let me explain. 

The functions `styleFieldToClassName` and `styleFieldToCssObject` are used to extract proper value in proper places
Properties of `appearance` prop accept either string, object or callback that may return either string or object. 

To properly handle it, I created this functions to either extract wanted value of return null
image

Hope it makes sense for you

Mr0Bread:
Hi @markflorkowski @juliusmarminge . Pushed improvements to PR

Rebased my commits onto latest `main`
`withUt` is now part of main package `uploadthing`
Made small improvements as per your feedback
Also left couple of answers to your comments

Please, take a look

docs/src/pages/theming.mdx
```diff
@@ -0,0 +1,731 @@
...
+additional classes and variants used to style our components.
+
+In addition, it also automatically sets the `content` option to include all the
+necessary classes that the components uses. This allows you to avoid having
+duplicated styles in your bundle. Therefore, when using `withUt`, you should not
+import our stylesheet into your app. If you choose not to use `withUt`, you have
+to import default stylesheet to make components look right.
+
+```ts filename="tailwind.config.ts"
+import { withUt } from "@uploadthing/react/tw";
```
juliusmarminge, line 168: wrong import

docs/src/pages/theming.mdx
```diff
@@ -0,0 +1,731 @@
...
+
+/** @type {import('tailwindcss').Config} */
+export default withUt({
+  // your config goes here
+});
+```
+
+```js filename="tailwind.config.js"
+// @ts-check
+const { withUt } = require("@uploadthing/react/tw");
```
juliusmarminge, line 178: same as above

juliusmarminge:
Some nits:

CleanShot 2023-08-12 at 22 02 14@2x

CleanShot 2023-08-12 at 22 01 38@2x

Mr0Bread:
@juliusmarminge Fixed things you mentioned

markflorkowski (approved):

markflorkowski:
Nit: contrast on example is too low:
image

juliusmarminge:
Hmm it doesn't look like `dark:` works for MDX components in the docs, but I tried making it somewhat work for both modes... not the easiest though..

markflorkowski:
Also just noticed, the logo is messed up in light mode on the docs. 
image

juliusmarminge:
ok fixed the darkmode

juliusmarminge (approved):

docs/src/pages/theming.mdx
```diff
@@ -0,0 +1,762 @@
+import { Callout } from "nextra-theme-docs";
+
+import { UploadButton, UploadDropzone } from "@uploadthing/react";
+
+# Theming
```
t3dotgg, line 5: Only thought here - it'd be nice to have a minimal example up top and a bit more context. Not your responsibility, will put it on my TODO

docs/src/pages/theming.mdx
```diff
@@ -0,0 +1,762 @@
...
+</Callout>
+
+<Callout type="warning">
+  Note: While in UploadButton the button element is defined using `label`, in
+  UploadDropzone it is defined using `button`. As an abstraction layer, button
+  element in these two components have special data attribute applied:
+  `data-ut-element="button"`.
+</Callout>
+
+<Image
```
t3dotgg, line 82: These diagrams are so good

docs/src/pages/theming.mdx
```diff
@@ -0,0 +1,762 @@
...
+import { withUt } from "uploadthing/tw";
+
+export default withUt({
+  // your config goes here
+});
+```
+
+```js filename="tailwind.config.js"
+// @ts-check
+const { withUt } = require("uploadthing/tw");
```
t3dotgg, line 178: Surprisingly elegant solution

t3dotgg (approved):
This is so good holy shit. Incredible work.

PR #219, opened 2023-07-16
fix: pagedir not using absolute url

juliusmarminge:
fixes https://discord.com/channels/966627436387266600/1102510616326967306/1130116966162645073


markflorkowski (approved):

markflorkowski (approved):
This is much better. I was trying to think of a cleaner way of doing what you initially implemented, but you beat me to it.

PR #225, opened 2023-07-17
feat: Sveltekit support

AlanAcDz:
Added support for Svelte and Sveltekit

- Created Sveltekit handle hook for API route handler in uploadthing package
- Created package for Svelte support with:
  - Custom "createUploader" function to pass typed props to components
  - Upload Button with progress indicator
  - Upload Dropzone with progress indicator
- Created example Sveltekit app to showcase components
- Updated docs to include Sveltekit stuff


AlanAcDz:
Ok so I'm pretty sure this should pass the CI check now. My bad, this is my first contribution to OS hehe

.prettierignore
```diff
@@ -24,3 +24,6 @@ yarn-error.log*
 dist
 .solid
 pnpm-lock.yaml
+
+.svelte-kit
+.vscode
```
markflorkowski, line 29: any particular reason for this?

examples/sveltekit/.gitignore
```diff
@@ -0,0 +1,10 @@
+.DS_Store
```
markflorkowski, line 1: Is there any reason why this needs a separate `.gitignore` from the one at the root of the repo?

packages/config/eslint/index.js
```diff
@@ -8,7 +8,7 @@ const config = {
     "plugin:@typescript-eslint/recommended-requiring-type-checking",
   ],
   parser: "@typescript-eslint/parser",
-  plugins: ["@typescript-eslint", "import"],
+  plugins: ["@typescript-eslint", "import", "svelte3"],
```
markflorkowski, line 11: We may want to break down the eslint config as to not include rules/plugins that are not used in every package. 

https://eslint.org/docs/latest/use/configure/configuration-files#cascading-and-hierarchy

cc @juliusmarminge @t3dotgg

packages/svelte/.gitignore
```diff
@@ -0,0 +1,11 @@
+.DS_Store
```
markflorkowski, line 1: Same as above, is this necessary to be separate from the `.gitignore` at the root of the repo?

packages/svelte/src/lib/component/UploadDropzone.svelte
```diff
@@ -0,0 +1,128 @@
...
+  import { classNames, generateClientDropzoneAccept } from "uploadthing/client";
+  import type { FileRouter } from "uploadthing/server";
+
+  import { INTERNAL_uploadthingHookGen } from "../useUploadThing";
+  import Dropzone from "./Dropzone.svelte";
+  import {
+    allowedContentTextLabelGenerator,
+    generatePermittedFileTypes,
+    progressHeights,
+    type UploadthingComponentProps,
```
markflorkowski, line 13: separate type imports preferred. I thought I had a lint rule for this 🤔

packages/svelte/src/lib/component/Dropzone.svelte
```diff
@@ -0,0 +1,338 @@
...
+    dragEnter: Event;
+    dragLeave: Event;
+    dragOver: Event;
+    dropAccepted: { acceptedFiles: InputFile[]; event: Event };
+    dropRejected: { rejectReasons: FileRejectReason[]; event: Event };
+    fileDialogCancel: void;
+  }
+
+  // default props
+  export let disabled = false;
```
markflorkowski, line 35: Are these intentionally exported? Could they be an object instead of a doze exports?

packages/svelte/src/routes/+page.svelte
```diff
@@ -0,0 +1,3 @@
+<h1>Welcome to your library project</h1>
```
markflorkowski, line 1: Is this file required?

.prettierignore
AlanAcDz: The .svelte-kit folder is auto generated by SvelteKit and should not be checked by prettier. I guess the .vscode one is unnecessary

examples/sveltekit/.gitignore
AlanAcDz: Not at all, will remove it

packages/svelte/src/lib/component/UploadDropzone.svelte
AlanAcDz: Let me check if there's an issue with linting in svelte files

packages/svelte/src/lib/component/Dropzone.svelte
AlanAcDz: Sure, I'll make them an object

packages/svelte/src/routes/+page.svelte
AlanAcDz: Honestly not sure, I guess not. I'll look into it

packages/config/eslint/index.js
juliusmarminge: Yes we can have multiple presets and then packages extends only what it needs:

https://github.com/t3-oss/create-t3-turbo/tree/main/packages/config/eslint

https://github.com/t3-oss/create-t3-turbo/blob/f73087c2422fd855c0e76ea69c66dcd6fed1e719/apps/nextjs/package.json#L46-L53

https://github.com/t3-oss/create-t3-turbo/blob/f73087c2422fd855c0e76ea69c66dcd6fed1e719/packages/api/package.json#L26-L31

I can do that tonight

packages/svelte/src/lib/component/UploadDropzone.svelte
AlanAcDz: Ok so I had to change `eslint-plugin-svelte3` cause it isn't compatible with svelte 4, the new one is `eslint-plugin-svelte`.

But, I had to put the svelte eslint overrides in the `.eslintrc.cjs` file, cause it does not work if I put it in the eslint config package.

Also, the new plugin does not detect generics in svelte components, so I had to put a bunch of eslint-disable comments, which is... unfortunate. There's already an [issue](https://github.com/sveltejs/svelte-eslint-parser/issues/306) for it

markflorkowski: Still looks like this is not separated :/

AlanAcDz: I'm not sure you're looking at the latest changes though 😅

Captura de pantalla 2023-07-27 124708

markflorkowski: Huh, I guess Github is diffing incorrectly 😭

docs/src/pages/sveltekit.mdx
```diff
@@ -0,0 +1,257 @@
...
+# Getting Started with SvelteKit
+
+Make sure you
+[install the packages and get your env variables first](/getting-started)
+
+## Package Setup
+
+<Steps>
+
+### Install the packages
```
t3dotgg, line 12: Increasingly feel like I need to make a shared component for this setup that takes the `additionalPackages` as props. Not for this PR, just a note for myself for the future

docs/src/pages/sveltekit.mdx
```diff
@@ -0,0 +1,257 @@
...
+
+### Add env variables
+
+```bash copy
+UPLOADTHING_SECRET=... # A secret key for your app (starts with sk_live_)
+UPLOADTHING_APP_ID=... # Your app id
+```
+
+<Callout type="warning">
+  If you don't already have a uploadthing secret key, [sign up](/create-account)
```
t3dotgg, line 46: `an` UploadThing secret key

docs/src/pages/sveltekit.mdx
```diff
@@ -0,0 +1,257 @@
...
+});
+```
+
+If you need multiple handle hooks in your application take a look at the
+[sequence](https://kit.svelte.dev/docs/modules#sveltejs-kit-hooks-sequence)
+helper function from SvelteKit.
+
+### Creating The UploadThing Components (optional)
+
+Generating components let's you pass your generic `FileRouter` once and then
```
t3dotgg, line 176: lets*

packages/svelte/src/lib/component/Uploader.svelte
```diff
@@ -0,0 +1,26 @@
+<script lang="ts" generics="TRouter extends FileRouter">
+  // This eslint comment is because eslint-plugin-svelte does not support component generics as of yet:
```
t3dotgg, line 2: This is incredibly sad but I appreciate you for calling it out

t3dotgg (approved):
Looks good to me! Would love to get a more svelte-y person to review if possible

AlanAcDz:
Awesome! I was actually in the process of updating the eslint setup and adding support for theming to the components, I just need to update the docs now (and check twice for typos haha)

packages/svelte/src/lib/component/Uploader.svelte
AlanAcDz: Yeah it's a bummer, but I did find a workaround for it :)

AlanAcDz:
This should have the latest changes from the main packages now. ESlint issues have been fixed and docs have been updated.
I apologize if this requires a new review, I was mostly done updating things when it got approved 😅

docs/src/pages/sveltekit.mdx
```diff
@@ -0,0 +1,331 @@
...
+
+<Callout type="warning">
+  If you don't already have an UploadThing secret key, [sign
+  up](/create-account) and create one from the
+  [dashboard!](https://uploadthing.com/dashboard)
+</Callout>
+
+### Define process in vite.config
+
+UploadThing internals use process.env which is not defined in vite, so it is
```
juliusmarminge, line 53: This can be removed now

docs/src/pages/sveltekit.mdx
```diff
@@ -0,0 +1,331 @@
...
+- (Optional) `middleware` to authenticate and tag requests
+- `onUploadComplete` callback for when uploads are completed
+
+```ts copy filename="src/lib/server/uploadthing.ts"
+import { createUploadthing } from "uploadthing/sveltekit";
+import type { FileRouter } from "uploadthing/sveltekit";
+
+const f = createUploadthing();
+
+export const uploadRouter = {
```
juliusmarminge, line 89: Can you make this router match the other adapters? We have a specific page for fileroutes which shows more detail so the introduction/quickstart only needs the basic

packages/svelte/src/lib/component/shared.ts
```diff
@@ -0,0 +1,108 @@
...
+} from "@uploadthing/shared";
+import type { UploadFileResponse } from "uploadthing/client";
+import type {
+  ErrorMessage,
+  FileRouter,
+  inferEndpointInput,
+  inferErrorShape,
+} from "uploadthing/server";
+
+export const generatePermittedFileTypes = (config?: ExpandedRouteConfig) => {
```
juliusmarminge, line 13: Fyi, These helper functions will be exported from uplodthing/client once the solid theming pr is merged

packages/svelte/src/lib/component/shared.ts
```diff
@@ -0,0 +1,108 @@
...
+ * })
+ */
+export function createUploader<TRouter extends FileRouter>(
+  props: FileRouter extends TRouter
+    ? ErrorMessage<"You forgot to pass the generic">
+    : UploadthingComponentProps<TRouter>,
+) {
+  // Cast back to UploadthingComponentProps<TRouter> to get the correct type
+  // since the ErrorMessage messes it up otherwise
+  return props as UploadthingComponentProps<TRouter>;
```
juliusmarminge, line 107: Is it intentional to just return props here?

packages/uploadthing/src/sveltekit.ts
```diff
@@ -0,0 +1,33 @@
...
+      if (request.method === "POST") {
+        return POST(request);
+      }
+      if (request.method === "GET") {
+        return GET();
+      }
+    }
+    return resolve(event);
+  };
+};
```
juliusmarminge, line 33: Can't sveltekit just use the /server entrypoint? Trying to avoid too many entrypoints, especially when they're all using Request/Response

juliusmarminge: I made this change for now. If there is a good reason not to, we can revert

packages/svelte/src/lib/component/shared.ts
juliusmarminge: merged now

packages/svelte/src/lib/component/shared.ts
AlanAcDz: Yes. You cannot just call a Svelte component with a generic type like you can in React, the type must be passed indirectly via props. This function is just a convenience helper to type the uploader props

packages/uploadthing/src/sveltekit.ts
AlanAcDz: I do think having the handle hook is useful, but maybe I can export it from the svelte package instead

AlanAcDz: I went ahead and created the handle hook in the svelte package, although I had to expose the `RouterWithConfig` type, let me know if that's ok

juliusmarminge: > I do think having the handle hook is useful

Why? I'm not saying it's bad I'm just trying to understand what the benefit is?

> I went ahead and created the handle hook in the svelte package

That doesnt make much sense to me, the client side libs doesnt need server code. We can expose a svelte-kit entrypoint if there is a reason for it, but to me it looks like svelte is request/response based so I don't see a reason you can't use that?

I'm also very lost on the difference between server hooks and api routes so might be completely off

AlanAcDz: A server hook in sveltekit acts like a middleware, it can be used for stuff like authentication for example. In this case my idea was to use a server hook to handle the uploadthing api routes, instead of creating a whole extra file to create an actual api route. I've seen other sveltekit libraries do a similar thing, like trpc-sveltekit. Maybe calling it useful was not the right word, it's more like a nice thing to have.

We could also just create the api routes file like with other frameworks, if you think the server hook thing is confusing/not worth it.

AlanAcDz: Decided to let go of the handle hook and just use an API route 😅

sby051:
May I suggest naming the package `sveltekit` specifically instead of just `svelte`? This package is intended for `SvelteKit`-specific use, not with a configuration such as `Vite + Svelte`, so this naming may reduce unwanted confusion of which type of Svelte project configuration it can be used in. 

This naming pattern would also apply to function, variable and export namings, replacing any instance of `Svelte` and `svelte` etc. with `SvelteKit` and `svelteKit` respectively.

AlanAcDz:
> May I suggest naming the package `sveltekit` specifically instead of just `svelte`? This package is intended for `SvelteKit`-specific use, not with a configuration such as `Vite + Svelte`, so this naming may reduce unwanted confusion of which type of Svelte project configuration it can be used in.
> 
> This naming pattern would also apply to function, variable and export namings, replacing any instance of `Svelte` and `svelte` etc. with `SvelteKit` and `svelteKit` respectively.

That's an interesting thought, however, you _could_ technically use the client specific parts of this package (dropzone, upload button, etc) in a Vite + Svelte project if you want, and use whatever other backend that has an uploadthing adapter

AlanAcDz:
Ok who's idea was it to put the "Close with comment" button with no confirmation besides the "Comment" button hahaha

juliusmarminge:
@AlanAcDz would you be up to getting this up to date?

juliusmarminge:
I also agree that package should be named `svelte` since client parts should be usable in any client based svelte app no matter what backend adapter you're using

sbyware:
> I also agree that package should be named `svelte` since client parts should be usable in any client based svelte app no matter what backend adapter you're using

This is fair! No worries at all.

AlanAcDz:
> @AlanAcDz would you be up to getting this up to date?
Sure!

juliusmarminge:
Rebase to main. We've ditched the community repo idea

AlanAcDz:
Is it normal for the checks to time out? I cannot shake the feeling that I messed something up during the rebase 🙃

juliusmarminge:
> Is it normal for the checks to time out? I cannot shake the feeling that I messed something up during the rebase 🙃

I think that was a bug in Bun 1.0.31 - but there's a newer version out now which seems to work

juliusmarminge:
@AlanAcDz 330+ files changed doesn't seem right. is your base branch up to date?

markflorkowski:
> @AlanAcDz 330+ files changed doesn't seem right. is your base branch up to date?

I think this is based off the community branch. I think we may have decided against that route? IIRC we were just going to have a folder inside of the regular main branch? cc @t3dotgg

AlanAcDz:
Oh so that was it, changing the base fixed the 330+ files issue

examples/minimal-sveltekit/vite.config.ts
```diff
@@ -0,0 +1,11 @@
+import { sveltekit } from "@sveltejs/kit/vite";
+import { defineConfig } from "vite";
+
+export default defineConfig({
+  plugins: [sveltekit()],
+  // Uploadthing internals use process.env which is not defined in vite, this is a workaround
+  // https://github.com/vitejs/vite/issues/1973
+  define: {
+    "process.env": {},
+  },
```
juliusmarminge, line 10: this shouldn't be necessary anymore

packages/svelte/package.json
```diff
@@ -0,0 +1,66 @@
...
+    "!dist/**/*.test.*",
+    "!dist/**/*.spec.*"
+  ],
+  "peerDependencies": {
+    "svelte": "^4.0.0",
+    "uploadthing": "workspace:6.5.2"
+  },
+  "dependencies": {
+    "@uploadthing/shared": "workspace:6.3.4",
+    "attr-accept": "^2.2.2",
```
juliusmarminge, line 36: shouldn't be necessary if you use our internal dropzone package as base

packages/svelte/src/app.d.ts
juliusmarminge, line 1: needed for libs too?

packages/svelte/src/app.html
juliusmarminge, line 1: same as above, needed for libs?

packages/svelte/src/lib/component/Dropzone.svelte
```diff
@@ -0,0 +1,364 @@
+<script lang="ts">
```
juliusmarminge, line 1: can you refactor this to use our `@uploadthing/dropzone/core` ?

packages/svelte/src/lib/useUploadThing.ts
```diff
@@ -0,0 +1,151 @@
...
+import type {
+  GenerateTypedHelpersOptions,
+  UploadthingComponentProps,
+  UseUploadthingProps,
+} from "./types";
+import { createFetch } from "./utils/createFetch";
+
+declare const globalThis: {
+  __UPLOADTHING?: EndpointMetadata;
+};
```
juliusmarminge, line 22: do you ever inject this?

packages/svelte/src/lib/utils/dropzone.ts
juliusmarminge, line 1: same here, refactor to use `@uploadthing/dropzone/core`

packages/svelte/src/lib/utils/styles.ts
juliusmarminge, line 1: isn't the required helpers exported from `uploadthing/client`?

packages/svelte/src/app.d.ts
AlanAcDz: It's not, I'll remove it

packages/svelte/src/app.html
AlanAcDz: Unfortunately it looks like it is 🤔

packages/svelte/src/lib/component/Dropzone.svelte
AlanAcDz: Sure!

packages/svelte/src/lib/component/UploadButton.svelte
```diff
@@ -0,0 +1,263 @@
...
+    uploadProgress: number;
+    fileTypes: string[];
+  };
+
+  type UploadButtonAppearance = {
+    container?: StyleField<ButtonStyleFieldCallbackArgs>;
+    button?: StyleField<ButtonStyleFieldCallbackArgs>;
+    allowedContent?: StyleField<ButtonStyleFieldCallbackArgs>;
+    clearBtn?: StyleField<ButtonStyleFieldCallbackArgs>;
+  };
```
juliusmarminge, line 47: I think we may need to adjust the types here to adjust for Svelte?

CleanShot 2024-03-21 at 21 58 27@2x

packages/svelte/src/lib/component/UploadButton.svelte
```diff
@@ -0,0 +1,263 @@
...
+
```
juliusmarminge, line 48: the `content` prop seems to be missing?

juliusmarminge: or is that what the `<slot>`'s are for? might need some docs for that if so

AlanAcDz: Yes that's what the `<slot>`'s are for, I thought I added that in the docs 🤔 might need to update them

docs/src/pages/getting-started/svelte.mdx
```diff
@@ -1 +1,311 @@
...
+
+## Theming Svelte components
+
+Check out our [theming guide](/theming) to know the basics about
+[components anatomy](/theming#components-anatomy) and
+[theming props](/theming#theming-props) as they still apply to
+`@uploadthing/svelte`.
+
+However, [content customization](/theming#content-customization) makes use of
+Svelte's named slots and slot props.
```
juliusmarminge, line 276: Nice! Include a link back here in the /theming#content-customization docs too:

> If you're using svelte, checkout the svelte docs...

examples/minimal-sveltekit/src/routes/+page.svelte
```diff
@@ -0,0 +1,50 @@
+<script lang="ts">
+  import { createUploader, useUploadThing } from "$lib/utils/uploadthing";
+
+  import { Uploader } from "@uploadthing/svelte";
+
+  import "@fontsource-variable/inter";
+
+  const uploader = createUploader({
+    endpoint: "videoAndImage",
```
juliusmarminge, line 9: Nit - but to keep consistency in the APIs, shouldn't endpoint be a positional argument here?

AlanAcDz:
I think there are some issues that are making the CI checks to fail but are not related to this PR 🤔. 
There's this issue when running `bun run test`, taken from the gh action logs:

```
Error: Playwright Test did not expect test() to be called here.
Most common reasons include:
- You are calling test() in a configuration file.
- You are calling test() in a file that is imported by the configuration file.
- You have two different versions of @playwright/test. This usually happens
  when one of the dependencies in your package.json depends on @playwright/test.
 ❯ TestTypeImpl._currentSuite node_modules/playwright/lib/common/testType.js:71:13
 ❯ TestTypeImpl._createTest node_modules/playwright/lib/common/testType.js:80:24
 ❯ Module.<anonymous> node_modules/playwright/lib/transform/transform.js:257:12
 ❯ examples/backend-adapters/test/smoke.test.ts:7:1
      5| const testFile = (name: string) => path.join(dirname, "test-files", na…
      6| 
      7| test("uploads a single image", async ({ page }) => {
       | ^
      8|   await page.goto("/");
      9|   await page.waitForSelector("text=Hello from Hono!");
```

There are these linting issues in solid's dropzone:

```
/Users/alan/Documents/Dev work/OS Projects/uploadthing/packages/dropzone/src/solid.tsx
   81:11  error  Prefer using an optional chain expression instead, as it's more concise and easier to read  @typescript-eslint/prefer-optional-chain
  238:9   error  Prefer using an optional chain expression instead, as it's more concise and easier to read  @typescript-eslint/prefer-optional-chain

✖ 2 problems (2 errors, 0 warnings)
  2 errors and 0 warnings potentially fixable with the `--fix` option.

error: script "lint" exited with code 1
```

And then I'm running into this issue when running or building the docs, not sure what this one means
```
 ⨯ ./src/pages/index.mdx
TypeError: Cannot use 'in' operator to search for 'themes' in txt
[nextra] Error compiling /Users/alan/Documents/Dev work/OS Projects/uploadthing/docs/src/pages/index.mdx.
```

markflorkowski:
Fixed the linting issue, unsure on the cause of the test issues

@juliusmarminge - mind taking a look when you are awake?

Will continue looking into the docs issue too.

juliusmarminge:
Looks like the vitest workspace got confused so it ran all the tests from the svelte package (including some Playwright ones). Added a vitest config seems to have fixed it

juliusmarminge:
really curious waht the docs issue is about...

juliusmarminge:
@AlanAcDz Published a beta from #735 and running into some issues with the devhook getting a 403. Any ideas?

Repo: https://github.com/juliusmarminge/test-svelte-ut

CleanShot 2024-03-26 at 11 21 45@2x

juliusmarminge:
Seems to be fixed by [adding explicit content-type header to the callback](https://github.com/pingdotgg/uploadthing/pull/225/commits/8bed90d7c00997ac5a87824d5f3a5a6307abf4ab): 
CleanShot 2024-03-26 at 13 01 22

packages/shared/src/component-utils.ts
```diff
@@ -76,18 +76,22 @@ export const allowedContentTextLabelGenerator = (
...
 
-type AnyRuntime = "react" | "solid";
+type AnyRuntime = "react" | "solid" | "svelte";
 type MinCallbackArg = { __runtime: AnyRuntime };
 type inferRuntime<T extends MinCallbackArg> = T["__runtime"] extends "react"
   ? "react"
-  : "solid";
+  : T["__runtime"] extends "solid"
+    ? "solid"
+    : "svelte";
```
juliusmarminge, line 85: nit:

```ts
type inferRuntime<T extends MinCallbackArg> = T["__runtime"] extends "react"
  ? "react"
  : T["__runtime"] extends "solid"
    ? "solid"
    : T["__runtime"] extends "solid"
      ? "svelte"
      : never;
```

packages/shared/src/component-utils.ts
```diff
@@ -76,18 +76,22 @@ export const allowedContentTextLabelGenerator = (
...
 
 type ElementEsque<TRuntime extends AnyRuntime> = TRuntime extends "react"
   ? ReactNode
   : JSX.Element;
 type CSSPropertiesEsque<TRuntime extends AnyRuntime> = TRuntime extends "react"
   ? CSSProperties
-  : JSX.CSSProperties;
+  : TRuntime extends "solid"
+    ? JSX.CSSProperties
+    : string;
```
juliusmarminge, line 94: nit:

```ts
type CSSPropertiesEsque<TRuntime extends AnyRuntime> = TRuntime extends "react"
  ? CSSProperties
  : TRuntime extends "solid"
    ? JSX.CSSProperties
    : TRuntime extends "svelte"
      ? string
      : never;
```

juliusmarminge (approved):
This looks good to me!

Some very small nits, maybe align the example more with the Next.js one (we don't use the `<Uploader />` component there, I personally don't see the value with it ;) )

juliusmarminge (changes requested):
Oh no Dropzone seems broken

Uploading CleanShot 2024-03-27 at 22.51.00.mp4…

juliusmarminge:
Fixed so it's selecting files, but it seems the accept prop doesn't quite work.

We're passing `accept: undefined` to the `createDropzone` function, and it doesn't update when the fileTypes has loaded. Some derived state not being reactive??? I have no clue how svelte's reactivity model works...

juliusmarminge:
Button works, dropzone doesnt. We should only be able to select images and videos here:

https://github.com/pingdotgg/uploadthing/assets/51714798/38f32fd8-4894-4b7f-8289-0b34a16d2699

AlanAcDz:
That's weird 🤔 I'll take a look

AlanAcDz:
So there was a weird edge-case with Svelte's reactivity that was causing the issue, should be fixed now

bmdavis419:
I spent a some time with the SvelteKit example and package and it felt good to me, all 3 examples worked, the code makes sense, I did not see any issues, docs had pretty much everything I needed as well. 100% possible I missed something as I'm not as familiar with the product but from what I tested it looks good.

I've spent a lot of time in the SvelteKit world over the last 6 months and I don't think that there are any major conventions that are followed like in react (something pointed out to me recently is that due to how well sveltekit works with normal JS they typically just use the basic JS lib for most things).

The only thing that felt a little out of place is the "useUploadThing". Its not a problem, it just feels very react brained lol. The callback system is great and I really like how you implemented "useUploadThing" my note would just be to maybe rename it to "uploadThing" kinda like how this lib does it (one of the biggest in Svelte land) https://superforms.rocks/concepts/client-validation.

juliusmarminge:
Taking some inspo from the crew at tanstack:

`createUploadThing` -> svelte, solid
`useUploadThing` -> React and Vue

https://tanstack.com/query/v4/docs/framework/svelte/overview
https://tanstack.com/query/v4/docs/framework/react/quick-start
https://tanstack.com/query/v4/docs/framework/vue/quick-start
https://tanstack.com/query/v4/docs/framework/solid/overview#important-differences-between-solid-query--react-query

juliusmarminge (approved):
Awesome work @AlanAcDz! Sorry it took this long to review 

🚢

examples/minimal-sveltekit/src/lib/utils/uploadthing.ts
```diff
@@ -0,0 +1,6 @@
+import { generateSvelteHelpers } from "@uploadthing/svelte";
```
markflorkowski, line 1: Nit: it's coming from the svelte package, calling this `generateSvelteHelpers` is a bit redundant. 

We do the same thing in the React package though I guess 🤔

packages/svelte/src/lib/create-uploadthing.ts
```diff
@@ -0,0 +1,147 @@
...
+  >(
+    endpoint: TEndpoint,
+    props: Omit<
+      UploadthingComponentProps<TRouter, TEndpoint, TSkipPolling>,
+      "endpoint"
+    >,
+  ) => ({ endpoint, ...props });
+};
+
+export const generateSvelteHelpers = <TRouter extends FileRouter>(
```
markflorkowski, line 134: See previous note about this naming.

packages/svelte/src/lib/types.ts
```diff
@@ -0,0 +1,99 @@
...
+   * @example "https://www.example.com/api/uploadthing"
+   *
+   * If relative, host will be inferred from either the `VERCEL_URL` environment variable or `window.location.origin`
+   *
+   * @default (VERCEL_URL ?? window.location.origin) + "/api/uploadthing"
+   */
+  url?: string | URL;
+}
+
+export type UseUploadthingProps<
```
markflorkowski, line 23: Nit: should name match `createUploadthing`?

markflorkowski (approved):

packages/svelte/src/lib/types.ts
juliusmarminge: Yes mb

examples/minimal-sveltekit/src/lib/utils/uploadthing.ts
markflorkowski: leave as is for now. may rethink this in the future, but it should not be locking for this PR

PR #226, opened 2023-07-17
chore: add server only guard

juliusmarminge:
Some are confused when getting the `"Missing UPLOADTHING_SECRET env variable."` error message when they try to use the utapi on the client. This improves that by adding a server only guard

### DIFF
--- a/packages/uploadthing/src/sdk/index.ts
+++ b/packages/uploadthing/src/sdk/index.ts
@@ -2,7 +2,17 @@ import { generateUploadThingURL } from "@uploadthing/shared";
 
 import { UPLOADTHING_VERSION } from "../constants";
 
-const UT_SECRET = process.env.UPLOADTHING_SECRET;
+function guardServerOnly() {
+  if (typeof window !== "undefined") {
+    throw new Error("The `utapi` should only be used on the client.");
+  }
+}
+
+function guardApiKey() {
+  if (!process.env.UPLOADTHING_SECRET)
+    throw new Error("Missing UPLOADTHING_SECRET env variable.");
+  return process.env.UPLOADTHING_SECRET;
+}
 
 /**
  * Request to delete files from UploadThing storage.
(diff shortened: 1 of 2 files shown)


packages/uploadthing/src/sdk/index.ts
```diff
@@ -2,7 +2,17 @@ import { generateUploadThingURL } from "@uploadthing/shared";
 
 import { UPLOADTHING_VERSION } from "../constants";
 
-const UT_SECRET = process.env.UPLOADTHING_SECRET;
+function guardServerOnly() {
+  if (typeof window !== "undefined") {
+    throw new Error("The `utapi` should only be used on the client.");
```
markflorkowski, line 7: "The `utapi` can only be used on the server."

packages/uploadthing/src/sdk/index.ts
```diff
@@ -13,14 +23,16 @@ const UT_SECRET = process.env.UPLOADTHING_SECRET;
  * await deleteFiles(["2e0fdb64-9957-4262-8e45-f372ba903ac8_image.jpg","1649353b-04ea-48a2-9db7-31de7f562c8d_image2.jpg"])
  */
 export const deleteFiles = async (fileKeys: string[] | string) => {
+  guardServerOnly();
+  const apiKey = guardApiKey();
```
markflorkowski, line 27: Is this used anywhere other than the fetch below?

markflorkowski (changes requested):

markflorkowski:
Good call on this

packages/uploadthing/src/sdk/index.ts
juliusmarminge: Whooops 😅

packages/uploadthing/src/sdk/index.ts
juliusmarminge: Nope, but I didn't wanna move the order of execution so I kept it at the beginning of the function. You wanna just in-line it straight in the fetch call?

markflorkowski: Yeah I was thinking to just inline it

markflorkowski (approved):

markflorkowski:
changeset?

PR #227, opened 2023-07-18
Sveltekit support

miguelvelasquezdev:


markflorkowski:
Looks like a duplicate of #225

Issue #230, opened 2023-07-19
[Feat]: UploadButton Customization


abedshaaban:
I made some designs on figma for the UploadButton with some props to change some text and customize the styling of it. 

# UploadButton Design
![Group 79](https://github.com/pingdotgg/uploadthing/assets/72554214/05aa09ab-e6b6-4208-b189-e43f7b869331)

- `description`: is the text under/ beside the button (example: `Images up to 16MB, max 2`)
- `color`: this is the text color of the `description` (example: `#ffffff` or `rgb(...)` or `hsl`) since this would be passed to the inline style.
- `backgroundColor`: this is the background color for the `description`
- `border`: border over all the conent
- `short`: if true content is `flex-column` else it is `flex-row`
- `btn.color`: the text color of the button
- `btn.backgroundColor`: background color of the button
- `btn.text`: text inside the button (example: `Choose File(s)`)

all these props are optional and have a default value to them so we attain the old UploadButton style.

i would like to upgrade the `UploadButton` and contribute this, but if someone can tell me how start.

markflorkowski:
I like this approach for it's simplicity, but I think #217 will cover all of these needs and more.

syntaxlexx:
Those updates will actually help quite a lot!
Like in this case, I would like to completely override the description text

![image](https://github.com/pingdotgg/uploadthing/assets/19994360/5851686a-6c57-4be8-9ead-da30ce8a5ef2)

markflorkowski:
Ah yeah.... we really didn't consider the case where you have a lot of accepted file types 😅 Definitely should have a way to handle that better.

PR #231, opened 2023-07-21
feat: UploadButton props for styling

abedshaaban:
This pull request is linked to #230 
all the props are not required, and so they have a default value
## code
```ts
     <UploadButton
          { ... }
          description={"Image up to 4MB"}
          color={"rgb(142,	158,	144)"}
          backgroundColor={"whitesmoke"}
          short={false}
          border={true}
          btn={{
            text: "Select",
            color: "black",
            backgroundColor: "#EB9C25",
          }}
      />
```

## result
image

### DIFF
--- a/.changeset/mean-peaches-invite.md
+++ b/.changeset/mean-peaches-invite.md
@@ -1,0 +1,5 @@
+---
+"@uploadthing/react": minor
+---
+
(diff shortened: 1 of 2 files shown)


juliusmarminge:
Hey,

We're moving forward with #217

PR #234, opened 2023-07-23
feat(utapi): Add renameFiles to utapi

markflorkowski:
- add UTAPI function
- update type name, add docs

### DIFF
--- a/docs/src/pages/api-reference/server.mdx
+++ b/docs/src/pages/api-reference/server.mdx
@@ -213,7 +213,7 @@ console.log(files);
 
 ### `renameFiles`
 
-`renameFiles` takes in an `Rename` or an array of `Rename`s and renames the
+`renameFiles` takes in a `Rename` or an array of `Rename`s and renames the
 appropriate files on the server
 
 ```ts
@@ -245,6 +245,6 @@ type Rename = {
 };
 ```
 
-| Arguments | Type                 | Required | Description                                 |
-| :-------- | :------------------- | :------- | :------------------------------------------ |
-| input     | Rename \|\| Rename[] | Yes      | The `Rename` or `Rename`s you want to apply |
+| Arguments | Type                 | Required | Description                                            |
+| :-------- | :------------------- | :------- | :----------------------------------------------------- |
+| renames   | Rename \|\| Rename[] | Yes      | Object specifying what file to rename and the new name |
--- a/docs/src/pages/api-reference/server.mdx
+++ b/docs/src/pages/api-reference/server.mdx
@@ -226,7 +226,7 @@ console.log(files);
 > Added in `v5.3`
 
 `renameFiles` takes in a `Rename` or an array of `Rename`s and renames the
-appropriate files on the server
+appropriate files on the server.
 
 ```ts
 import { utapi } from "uploadthing/server";
@@ -253,7 +253,7 @@ await utapi.renameFiles([
 ```ts
 type Rename = {
   fileKey: string;
-  newName: string;
+  newName: string; // should include file extension
 };
 ```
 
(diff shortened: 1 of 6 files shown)


docs/src/pages/api-reference/server.mdx
```diff
@@ -210,3 +210,41 @@ console.log(files);
...
+```ts
+type Rename = {
+  fileKey: string;
+  newName: string;
+};
+```
+
+| Arguments | Type                 | Required | Description                                |
+| :-------- | :------------------- | :------- | :----------------------------------------- |
+| input     | Rename \|\| Rename[] | Yes      | The fileKey or fileKeys you want to delete |
```
Hacksore, line 250: The description is copypasta'd here

docs/src/pages/api-reference/server.mdx
```diff
@@ -210,3 +210,41 @@ console.log(files);
  * ]
  */
 ```
+
+### `renameFiles`
+
+`renameFiles` takes in an `Rename` or an array of `Rename`s and renames the
```
juliusmarminge, line 216:
```suggestion
`renameFiles` takes in a `Rename` or an array of `Rename`s and renames the
```

docs/src/pages/api-reference/server.mdx
juliusmarminge, line 250:
```suggestion
| renames | Rename \|\| Rename[] | Yes | Object specifying what file to rename and the new name |
```

packages/uploadthing/src/sdk/index.ts
```diff
@@ -111,3 +111,53 @@ export const listFiles = async () => {
...
+ * await renameFile([{ fileKey: "2e0fdb64-9957-4262-8e45-f372ba903ac8_image.jpg", newName: "new_image.jpg" }, { fileKey: "1649353b-04ea-48a2-9db7-31de7f562c8d_image2.jpg", newName: "new_image2.jpg" }]);
+ *
+ */
+
+type Rename = {
+  fileKey: string;
+  newName: string;
+};
+
+export const renameFile = async (updates: Rename | Rename[]) => {
```
juliusmarminge, line 138: would it make sense to call this `updateFile` instead for future proofing being able to update other attributes than just the name?

packages/uploadthing/src/sdk/index.ts
```diff
@@ -111,3 +111,53 @@ export const listFiles = async () => {
...
+    headers: {
+      "Content-Type": "application/json",
+      "x-uploadthing-api-key": getApiKeyOrThrow(),
+      "x-uploadthing-version": UPLOADTHING_VERSION,
+    },
+    body: JSON.stringify({ updates }),
+  });
+
+  const json = (await res.json()) as
+    | { files: { key: string; id: string }[] }
```
juliusmarminge, line 154: this type does not match what the infra returns

packages/uploadthing/src/sdk/index.ts
```diff
@@ -111,3 +111,53 @@ export const listFiles = async () => {
...
+    | { error: string };
+
+  if (!res.ok || "error" in json) {
+    const message = "error" in json ? json.error : "Unknown error";
+    throw new Error(message);
+  }
+
+  return res.json() as Promise<{ success: boolean }>;
```
juliusmarminge, line 162: fix the type above and this cast shouldn't be necessary

juliusmarminge (changes requested):

packages/uploadthing/src/sdk/index.ts
markflorkowski: I considered this, but figured we can always add an `updateFile()` in the future, and if all it does now is allow renaming files, it is a bit confusing to name it `updateFile()`

docs/src/pages/api-reference/server.mdx
```diff
@@ -210,3 +210,41 @@ console.log(files);
  * ]
  */
 ```
+
+### `renameFiles`
+
```
juliusmarminge, line 215: > Added in `v5.3`

PR #240, opened 2023-08-04
fix: add Airbnb's ESLint config and conform to specifications

Samathingamajig:
This project has gotten too large, in both files/lines of code and contributors, to allow inconsistent styling and low-quality code. The BEST way to enforce proper styling and high-quality code is with an ESLint configuration by our friends at [Airbnb](https://en.wikipedia.org/wiki/Airbnb). They have made an excellent ESLint configuration ([here](https://www.npmjs.com/package/eslint-config-airbnb)) and we have been fools for not adopting this sooner.

During this past week's live stream, [Theo](https://twitter.com/t3dotgg) (someone who has contributed to this repo before) discussed for almost an hour why Airbnb's configuration choices are excellent. He titled that section of the stream "AirBNB Is BAD At JavaScript", but as a frequent watcher of [Rick and Morty](https://en.wikipedia.org/wiki/Rick_and_Morty) I was able to deduce the sarcasm that Theo employed to actually give the effect of believing that Airbnb is EXCELLENT, no, GODLIKE at JavaScript. You can view the live stream section here: https://youtu.be/PDBrtngJoug?t=15207s.

There are a few slight API changes, such as `DANGEROUS__uploadFiles` --> `dangerousUploadFiles`, but that's the users' fault if they use a dangerous function, so this should be fine as a patch change in semver.

You don't even need to review this code, just go ahead and merge it in.


Samathingamajig:
just ignore the failing vercel deployment <3

mirashif:
LGTM!

markflorkowski:
Sorry, not strict enough, and too few bad opinions

PR #248, opened 2023-08-06
fix: return expected response

miguelvelasquezdev:


miguelvelasquezdev:
Oh there's another Julius PR open that solves the same thing :)

PR #249, opened 2023-08-06
fix: removes ts-expect-error

miguelvelasquezdev:


juliusmarminge (approved):
This is probably just as unsafe but i guess it looks better without an expect error

PR #251, opened 2023-08-07
fix: pass same info to onClientUploadComplete as onUploadComplete

markflorkowski:
Addresses https://discord.com/channels/966627436387266600/1102510616326967306/1138143461908951111

Previously was only providing file key and url, now includes name and size as well 
to match with the input to `onUploadComplete()`

would be nice to have the objects be identical, however that would break back 
compat. 

`onUploadComplete()` uses this shape:

```
{
   url: string,
   key: string,
   name: string, 
   size: number
}
```

while `onClientUploadComplete()` uses:

```
{
   fileName: string,
   fileSize: number,
   fileKey: string,
   fileUrl: string
}
```

### DIFF
--- a/packages/uploadthing/src/client.ts
+++ b/packages/uploadthing/src/client.ts
@@ -171,18 +171,18 @@ export const DANGEROUS__uploadFiles = async <TRouter extends FileRouter>(
 
     // TODO: remove `file` prefix in next major version
     return {
-    // @deprecated - use `name` instead
-    fileName: file.name,
-    name: file.name,
-    // @deprecated - use `size` instead
-    fileSize: file.size,
-    size: file.size,
-    // @deprecated - use `key` instead
-    fileKey: presigned.key,
-    key: presigned.key,
-    // @deprecated - use `url` instead
-    fileUrl: genUrl,
-    url: genUrl,
+      // @deprecated - use `name` instead
+      fileName: file.name,
+      name: file.name,
+      // @deprecated - use `size` instead
+      fileSize: file.size,
+      size: file.size,
+      // @deprecated - use `key` instead
+      fileKey: presigned.key,
+      key: presigned.key,
+      // @deprecated - use `url` instead
+      fileUrl: genUrl,
+      url: genUrl,
     };
   });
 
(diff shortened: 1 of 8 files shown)


juliusmarminge (approved):
Maybe add a TODO to remove the `file` prefiz in the next major

packages/uploadthing/src/client.ts
```diff
@@ -169,7 +169,10 @@ export const DANGEROUS__uploadFiles = async <TRouter extends FileRouter>(
     // Poll for file data, this way we know that the client-side onUploadComplete callback will be called after the server-side version
     await pollForFileData(presigned.key);
 
+    // TODO: remove `file` prefix in next major version
     return {
+      fileName: file.name,
+      fileSize: file.size,
       fileKey: presigned.key,
       fileUrl: genUrl,
```
juliusmarminge, line 177: Could also be a good idea to deprecate them now

```suggestion
    // @deprecated - use `name` instead
    fileName: file.name,
    name: file.name,
    // @deprecated - use `size` instead
    fileSize: file.size,
    size: file.size,
    // @deprecated - use `key` instead
    fileKey: presigned.key,
    key: presigned.key,
    // @deprecated - use `url` instead
    fileUrl: genUrl,
    url: genUrl,
```

packages/uploadthing/src/client.ts
```diff
@@ -169,9 +169,20 @@ export const DANGEROUS__uploadFiles = async <TRouter extends FileRouter>(
     // Poll for file data, this way we know that the client-side onUploadComplete callback will be called after the server-side version
     await pollForFileData(presigned.key);
 
+    // TODO: remove `file` prefix in next major version
     return {
-      fileKey: presigned.key,
-      fileUrl: genUrl,
+    // @deprecated - use `name` instead
```
juliusmarminge, line 174: Indentation broke. Suggesting on mobile is hard 😅

markflorkowski: Haha I'll fix it in a minute.

juliusmarminge:
hmm deprecated isn't showing up on hover from jsdoc :/ 
CleanShot 2023-08-08 at 08 55 56@2x

juliusmarminge:
Got it:

CleanShot 2023-08-08 at 09 02 59@2x

juliusmarminge (approved):

PR #252, opened 2023-08-08
fix(bug): declaration support broken

juliusmarminge:
Similar to #192

### DIFF
--- a/examples/appdir/tsconfig.json
+++ b/examples/appdir/tsconfig.json
@@ -2,6 +2,9 @@
   "extends": "@uploadthing/tsconfig/nextjs.json",
   "compilerOptions": {
     "plugins": [{ "name": "next" }],
+    "declaration": true,
+    "noEmit": false,
+    "outDir": "dist",
     "baseUrl": ".",
     "paths": {
       "~/*": ["./src/*"]


markflorkowski (approved):

juliusmarminge:
I do wonder why CI is passing, build is failing locally.

(I haven't attempted on fixing it yet, just threw the PR up to show how to repro the build error for @GentikSolm)

GentikSolm:
Hm, this is still working for me.
image

juliusmarminge:
> Hm, this is still working for me.
> 
> image
> 
> 

So you not see errors in your editor? utils/uploadthing file on the `useUploadthing` hook?

GentikSolm:
Ahh no i see it now. That is weird that it passes build and linting steps fine

GentikSolm:
I have this nailed down to `UploadThingError`'s type not being able to be decoupled from its actual impl, so the type cant be inferred without reference to that folder.

GentikSolm:
Tried throwing a bunch of stuff at the wall but nothing stuck. I can confirm that it is for sure `UploadThingError` though. For some reason its not able to pull the typedef out. Open to suggestions.

juliusmarminge:
I wonder if it might not be best to nuke the shared pacakge all together, we can share stuff from the `uploadthing` package anyways

GentikSolm:
that makes sense to me. I can take a stab at that refactor if thats the direction we want to go

juliusmarminge:
Wut do you think @markflorkowski ? Should we just share code from the `uploadthing` pacakge?

markflorkowski:
That seems reasonable enough to me. I did like the separation, but it isn't worth the headache.

GentikSolm:
Making a new branch for this since i dont have write perms

GentikSolm:
Resolved by #272

markflorkowski:
Superseded by #272

Issue #258, opened 2023-08-10
`uploadthing` should provide expressive error messages when the uploading file is larger than the `maxFileSize`


A7med3bdulBaset:
### What happened?

I've followed the [Getting Started with App](https://docs.uploadthing.com/nextjs/appdir). then, uploaded an image that is larger than the `maxFileSize`. I got this error:
```console
Error: Failed to upload file IMG_20220627_102753_670_4.jpg to S3
    at eval (webpack-internal:///(:3000/app-pages-browser)/./node_modules/.pnpm/uploadthing@5.3.2_zod@3.21.4/node_modules/uploadthing/dist/client.mjs:108:13)
    at async Promise.all (:3000/index 0)
    at async eval (webpack-internal:///(:3000/app-pages-browser)/./node_modules/.pnpm/@uploadthing+react@5.2.1_react-dropzone@14.2.3_react@18.2.0_uploadthing@5.3.2_zod@3.21.4/node_modules/@uploadthing/react/dist/chunk-HRSKA5IZ.mjs:122:21)
```

The state in the dashboard is `uploading` up til now
![Dashboard](https://i.ibb.co/SBGcp4m/uploadthing.png)

### What did you expect to happen?

to get `error.code` (or any equivalent) by the cause of failure. also, to update the state in the dashboard from `uploading` to `failure`

### Version

uploadthing: 5.3.2
@uploadthing/react: 5.2.1

### Reproduction Steps

set the `maxFileSize` to `1MB` then upload `>1MB` file.

### Reproduction Repository

```ts
// app/api/uploadthing/core.ts
import { createUploadthing, type FileRouter } from "uploadthing/next";

const f = createUploadthing();

const auth = (req: Request) => ({ userId: "123" });

const ourUploadthingRouter = {
	imageUploader: f({ image: { maxFileSize: "1MB" } })
		.middleware(({ req }) => {
			const user = auth(req);
			return {
				userId: user.userId,
			};
		})
		.onUploadComplete(({ file, metadata }) => {
			// eslint-disable-next-line no-console
			console.log(`Uploaded ${file.name} by userId ${metadata.userId}`);
		}),
} satisfies FileRouter;

type OurUploadthingRouter = typeof ourUploadthingRouter;

export { ourUploadthingRouter, type OurUploadthingRouter };
```

```ts
// app/api/uploadthing/route.ts
import { createNextRouteHandler } from "uploadthing/next";

import { ourUploadthingRouter } from "./core";

[...]

PR #260, opened 2023-08-10
fix: better errors on failed uploads

markflorkowski:
Addresses #258


markflorkowski:
@juliusmarminge This is something that I hacked together in ~2 min, we should probably figure out how best to do something like this though.

juliusmarminge (commented):
This won't work for the same error in server uploads though

markflorkowski:
> This won't work for the same error in server uploads though

because of DOMParser? We can swap it for some other parser, or I can write a regex to extract the strings we want...

juliusmarminge:
> > This won't work for the same error in server uploads though
> 
> 
> 
> because of DOMParser? We can swap it for some other parser, or I can write a regex to extract the strings we want... 

Yes, it's a browser thing.

Regex could work I guess :)

markflorkowski:
@juliusmarminge here's a regex version for you

Mr0Bread:
Hi @markflorkowski 

Maybe I'm missing something, but it looks like to me that you are extracting the error code despite the fact that it's not used anywhere. 

On the image below error constructor receives parsed message only
image

markflorkowski:
@Mr0Bread Yes, I was playing around with what info is returned from s3 to see what would be most useful. Probably should merge the two and return a formatted string, or pass both into the error (one as cause, one as message). This was not meant to be merged as is, more just getting an idea down so we can figure out the correct way to pass this information to the users.

juliusmarminge:
I added some stuff @markflorkowski

packages/uploadthing/src/sdk/utils.ts
```diff
@@ -83,6 +85,7 @@ export const uploadFilesInternal = async (
         formData.append(key, value);
       });
       formData.append("file", file);
+      formData.set("Policy", "");
```
markflorkowski, line 88: ? @juliusmarminge

juliusmarminge: Ohhh shoot that was my way of testing a failed upload hahhh. Sorry about thst. Revert

packages/uploadthing/src/sdk/utils.ts
```diff
@@ -94,6 +96,15 @@ export const uploadFilesInternal = async (
       });
 
       if (!s3res.ok) {
+        const text = await s3res.text();
+        const parsed = maybeParseResponseXML(text);
+        console.log({ text, parsed });
```
juliusmarminge, line 101: Rm

```suggestion
```

juliusmarminge (approved):

PR #261, opened 2023-08-12
chore: update to `@typescript/eslint@6` and split config into composable presets

juliusmarminge:
## Core changes

- Updates to typescript-eslint@6
- Splits eslint config from a single big config to smaller presets. Ref: https://github.com/pingdotgg/uploadthing/pull/225#discussion_r1269989816

## Other
- Minor monorepo fixes, sorts pkg jsons
- Rm unnecessary gitignores
- Bump pnpm
- Bump `engines` of node to 18
- Set `engines` of pnpm to 8 - prevents people from running `pnpm install` if they are not on pnpm 8


packages/config/eslint/package.json
```diff
@@ -1,24 +1,20 @@
 {
   "name": "@uploadthing/eslint-config",
-  "private": true,
```
markflorkowski, line 3: @juliusmarminge this is what did it.

PR #263, opened 2023-08-14
fix: require repro to open an issue

juliusmarminge:
hate when people don't provide a repro


markflorkowski (approved):
One issue being that if people are reporting an issue with the dashboard, there won't be a repro

juliusmarminge:
Should we add a note (or https://uplodthing.com if the issue regards the dashboard)?

Issue #109, opened 2023-05-27
Feat: Add onUploadBegin prop to UploadButton


thebishalniroula:
It would be better if we could add a new prop to UpoadButton that takes a function which runs when uploadthing begins the uploading in the background.

Use case of this feature would be when I want to show some visual feedback when the upload begins.

irtaza9:
The better should be to have all the features like onuploadprogress and onuploadbegins, etc. I think axios is offering onuploadprogress callback.

markflorkowski:
We definitely want to add an `onUploadProgress()` callback. I'll have to think a bit more about `onUploadBegin()`, but I think it makes sense as well.

irtaza9:
Cool, let me know if you guys need any help. I'll be happy to add my part.

markflorkowski:
Feel free to open a PR if you have ideas on this and want to contribute! Otherwise, i have put this in our backlog to be addressed *soon*(tm)

irtaza9:
I was thinking about it but It's tough 
![image](https://github.com/pingdotgg/uploadthing/assets/98938307/2c6ea0e6-9dab-4d30-8cb0-9b2294ab2f69)

markflorkowski:
I am thinking if we use XHR instead of fetch, it has onprogress built in, which could work. Alternatively we could do uploads as multipart, and use the completion of each part to update the progress.

irtaza9:
And that is what I want to suggest but I thought all the package is build around fetch may be you people don't go for XHR. btw what about axios? 

### XHR:

```javascript
   const xhr = new XMLHttpRequest();
   xhr.addEventListener("loadstart", handleEvent);
   xhr.addEventListener("load", handleEvent);
   xhr.addEventListener("loadend", handleEvent);
   xhr.addEventListener("progress", handleEvent);
   xhr.addEventListener("error", handleEvent);
   xhr.addEventListener("abort", handleEvent);
``` 

### helper-fun

```javascript
   function handleEvent(e) {
      console.log(e.loaded)
   }
``` 

I got this resource over [mdn](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/progress_event)

Mr0Bread:
@markflorkowski onUploadBegin prop had been released. I guess this issue can be closed now

markflorkowski:
onUploadBegin has been added as of  v5.4.0 🙌

ryanpratama14:
onUploadBegin is not working somehow...

[...]

PR #264, opened 2023-08-14
feat: onUploadBegin prop

Mr0Bread:
Addresses #109 

Adds new prop `onUploadBegin` to components. Prop accepts the callback. Callback receives single param - name of the file upload of which have just started. The callback is called once `loadstart` event of XHR fires

P.S. Will add same stuff to SolidJS package if such approach is approved


juliusmarminge (commented):
Is there any difference between this and onUploadProgress and checking if progress is 0?

markflorkowski:
I'll second what Julius is asking, but am also ok with the approach of adding an additional optional callback for this case as it is a bit more readable if you are not using progress, and only want some action on uploads starting.

Mr0Bread:
@juliusmarminge @markflorkowski 

Firstly, I noticed that unUploadProgress not always being called with 0. I guess it happens when file size is small. Secondly, there is delay between call of unUploadBegin and first invocation of onUploadProgress. The amount of delay can be seen in the attached screenshot in second log
image
Here is the code that generates such an output
```jsx
onClientUploadComplete={(res) => {
  // Do something with the response
  console.log("Files: ", res);
  alert("Upload Completed");
}}
onUploadBegin={(name) => {
  console.log("Started upload at: ", Date.now().toString());
  console.time('uploadbegin')
}}
onUploadProgress={(progress) => {
  console.timeEnd('uploadbegin')
  console.log("Progressed at: ", Date.now().toString(), ":", progress);
}}
```

markflorkowski:
> I noticed that onUploadProgress not always being called with 0.

Thats... concerning. Just for that, I think it is worthwhile to do this.  

The delays aren't a huge concern for me, since in the case of small files, by the time the event hits, the file is probably uploaded already, and for larger files, a difference of 1 second is not really a big deal IMO.

Mr0Bread:
Will be adding adjustments for SolidJS package then and docs

Mr0Bread:
@markflorkowski @juliusmarminge Added changes for SolidJS package and also appended docs

Many changes from `pnpm format` btw

examples/appdir/src/app/page.tsx
```diff
@@ -45,6 +45,9 @@ export default function Home() {
           //   },
           // }}
           className="ut-button:bg-red-400 ut-allowed-content:text-lg ut-button:ut-readying:bg-orange-300 rounded border p-3"
+          onUploadBegin={(name) => {
+            console.log("Uploading: ", name);
```
markflorkowski, line 49: I would maybe add "onUploadBegin:" to the log here, just so anything in the console is more clear where it is coming from.

packages/react/src/components/dropzone.tsx
```diff
@@ -57,17 +57,17 @@ export function UploadDropzone<TRouter extends FileRouter>(
   const $props = props as unknown as UploadDropzoneProps<TRouter> & {
     // props not exposed on public type
     // Allow to set internal state for testing
-    __internal_state?: "readying" | "ready" | "uploading",
+    __internal_state?: "readying" | "ready" | "uploading";
```
markflorkowski, line 60: how did this work before? 🤔

markflorkowski (approved):

packages/react/src/components/dropzone.tsx
juliusmarminge: __internal_XXX are properties we added in the styling PR to be able to control the component state without user interaction to show the different states in the documentation

markflorkowski: No, I mean the commas in a type def. AFAIK that should throw terrible errors, no?

Mr0Bread: Not sure about the errors you are mentioning. Worked fine in this case

examples/appdir/src/app/page.tsx
Mr0Bread: Okay

packages/react/src/components/dropzone.tsx
juliusmarminge: No, both commas and semicolons are valid in type objects

markflorkowski: Huh, TIL

examples/appdir/src/app/page.tsx
Mr0Bread: @markflorkowski done

juliusmarminge: Did your latest commit mess up formatting?

We shoudl add a ci step for that 😅

Mr0Bread: Sorry, just did run pnpm format and pushed

t3dotgg (approved):

PR #271, opened 2023-08-18
chore(docs): Add documentation for createNextRouteHandler config 

markflorkowski:


docs/src/pages/api-reference/next.mdx
```diff
@@ -0,0 +1,25 @@
...
+  config: {
+    uploadthingId: "someId",
+    uploadthingSecret: "someSecret",
+  },
+});
+```
+
+### `config` Object
+
+This configuration is entirely optional, but may be useful if you would like to
```
t3dotgg, line 19: Maybe note the env variable we use by default?

docs/src/pages/api-reference/next.mdx
```diff
@@ -0,0 +1,27 @@
+## createNextRouteHandler
+
+```tsx
+import { createNextRouteHandler } from "uploadthing/next";
+
+import { ourFileRouter } from "./uploadthing.ts";
+
+export const { GET, POST } = createNextRouteHandler({
+  router: ourFileRouter,
+  config: {
```
juliusmarminge, line 10: we have a `callbackUrl` option too: https://github.com/pingdotgg/uploadthing/blob/9b1e2b722f5a80d5bb6205491e753998e4420d6b/packages/uploadthing/src/internal/handler.ts#L130C3-L130C26

docs/src/pages/api-reference/next.mdx
juliusmarminge, line 1: should probably note down that this works for the `createNextPageApiHandler` too

juliusmarminge (changes requested):

PR #272, opened 2023-08-20
fix: declaration support

GentikSolm:
- ref: move shared to uploadthing package
- ref: migrate other packages to use uploadthing/shared
- fix: UploadButton & UploadDropzone uploadProgress background

# Pull Request

## Linked GH Issues
N/A

## Summary of Change
Fixes our favorite TS2742 bug by refactoring @uploadthing/shared into uploadthing.

Also fixes
- UploadProgress is set to 0 when upload fails
- Spinner now shows above the uploadProgress background


packages/react/package.json
```diff
@@ -37,13 +37,12 @@
     "dev": "wait-on ../uploadthing/dist/server.mjs && tsup --watch"
   },
   "dependencies": {
-    "@uploadthing/shared": "^5.2.0",
-    "tailwind-merge": "^1.13.2"
+    "tailwind-merge": "^1.13.2",
+    "uploadthing": "^5.0.0"
```
juliusmarminge, line 41: This should be a peerDep still otherwise users might have multiple versions of uploadthing installed in their project

juliusmarminge (changes requested):
On vacay so didnt review just noticed this

packages/react/package.json
GentikSolm: When I had it as a peer dep and full dep, the linter yelled about uploadThing being a peer dep but not a dev dep. Should uploadThing just be in peer and dev then?

juliusmarminge:
> Moved uploadthing back to a peerDep, however that forces it to be a devDep. Originally both the react and solid package had @uploadthing/shared as full dependencies. If this is alright, then this should be ready for merge.

Yes, peer and dev are correct

packages/uploadthing/src/shared/file-types.ts
juliusmarminge, line 1: this file has a `zod` import.

need to think if this will cause any bundler to scream if it can't find it. not entirely sure if it will behave different and potentially get pulled in (not tree shaked) when it's in `uploadthing` pacakge vs `@uploadthing/shared` 🤔

juliusmarminge:
Don't like that we need to do this, but it would save a lot of headaches. Maybe we can take some inspiration from `@tanstack/query-core` where they manage to share code wihtout issues 🤔

juliusmarminge:
probably a wont-fix. we went down a deep rabbit hole with trpc trying to move stuff into a core package but concluded it wasn't possible to make pnpm happy when it puts the transistive deps in a nested folder.

in the end i think `declaration: false` should be a viable option for all uploadthing users, I can't see why you'd need to emit declaration types for your uploadthing router...

PR #274, opened 2023-08-20
fix: incorrect app file name

joshuaepstein:
The file name for a page in the app router is incorrect as it requires the path in the website followed by "/page.tsx" whereas this is missing that:
image


markflorkowski (approved):

joshuaepstein:
I also just found in the api-reference/react that there are also these issues however some of them also use client side functions/methods without use client mentioned.

PR #278, opened 2023-08-22
fix(utapi): no caching on utapi functions

markflorkowski:


juliusmarminge (approved):
Stupid Next.js making this necessary... Why would you enable automatic caching of `POST` requests...

Issue #273, opened 2023-08-20
Dropzone for a single file


MarkLyck:
I've just setup UploadThing for my project. But I seem to be running into a silly problem of the Dropzone component being more complex than I want/need it to be, with no way to disable the stuff I don't need.

My use-case is a very simple 1 image upload in a form. I'd like my users to be able to just drag-and-drop the image unto the dropzone, and have it automatically start the upload.

I don't want my users to have to go through a second step of clicking "Upload" once they've already chosen the image. It's unnecessary.

But I don't see any setting to just auto-upload on file select? Can you add this option 🙏

MarkLyck:
Managed to get this working using `useUploadThing` but would be nice if you could use the dropzone component for a super simple scenario like this.

Mr0Bread:
@markflorkowski We could provide a way of passing additional props to `useDropzone` hook. It will provide users with more control over dropzone behaviour. In addition, we could allow passing custom `onDrop` callback which could accept several parameters like acceptedFiles and some callback to programatically start uploading. With that author could have done something like that

```jsx
<UploadDropzone
  dropzone={{
    onDrop: ({ triggerUpload }) => {
      triggerUpload()
    }
  }}
/>
```

What do you think?

P.S. Not pinging Julius since he is on vacation from what I understood

markflorkowski:
I was thinking something more like your second suggestion, a prop for choosing between uploading immediately on drop, or using the "upload" button. 

I think a boolean value is sufficient, as we can get the expected number of files from the endpoint. It does raise the question of if the endpoint accepts "up to X" and you have this option enabled, should there be a button to upload less than the maximum?

markflorkowski:
I am hesitant to expose the onDrop() directly as I feel like the benefit of our components is that they handle simple cases well enough for most people, and if you need more complexity, you can use the hook. We should support the ability to swap between these two behaviours of "immediate upload" and "confirmed upload" though.

markflorkowski:
After thinking about this some more, I think having 3 "modes" for the dropzone makes sense:

- `manual` : Current behaviour, requires user to click "upload" button after selecting files
- `auto` : immediately uploads on file selection
- `threshold`: requires user to click "upload" button, unless they select X or more files, where x is the threshold provided in the config. 

Something along the lines of:

```tsx
<UploadDropzone
    config: {
        mode: 'auto' | 'manual' | 'threshold = 'auto';
        threshold?: number;
    }
/>
```

[...]

PR #279, opened 2023-08-23
feat(react): Dropzone modes

Mr0Bread:
Three modes are available:
- manual
- auto
- threshold

`manual` - default behaviour. Upload have to be triggered manually
`auto` - will start upload right after drop
`threshold` - will start upload once if number of dropped files is equal or greater than threshold value. Threshold can be passed either explicitly through config, or maximum number of files will be used. Maximum number of files is equal to largest maximum number among specified types. Though maybe we should use different value

Addresses #273

### DIFF
--- a/docs/src/pages/api-reference/react.mdx
+++ b/docs/src/pages/api-reference/react.mdx
@@ -67,15 +67,17 @@ export const OurUploadDropzone = () => (
 
 ### Configuration
 
-| Prop                   | Type     | Required                | Notes           | Description                                                                         |
-| :--------------------- | :------- | :---------------------- | :-------------- | :---------------------------------------------------------------------------------- |
-| `<FileRouter>`         | generic  | Yes                     |                 | The type from the FileRouter you defined in your backend                            |
-| endpoint               | string   | Yes                     |                 | The name of the [route](./server#FileRouter) you want this button to upload to      |
-| input                  | string   | If set on the fileroute | Added in `v5.0` | Object matching the input schema that was set for this endpoint in your File Router |
-| onClientUploadComplete | function | No                      |                 | `{fileKey: string, fileUrl: string}[]` callback function when upload is completed   |
-| onUploadError          | function | No                      |                 | callback function when upload fails                                                 |
-| onUploadProgress       | function | No                      | Added in `v5.1` | callback function for upload progress                                               |
-| onUploadBegin          | function | No                      | Added in `v5.4` | callback function for upload begin                                                  |
+| Prop                   | Type               | Required                | Notes           | Description                                                                             |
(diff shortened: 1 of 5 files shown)


markflorkowski:
Looking over this, it seems good to me, but the default threshold is definitely a bit of a hack.

I'll need to think more on this, but we may need to skip threshold for now, unless we can specify minFileCount on the router as well, then make the logic take into account the threshold for all file types. This is much more complicated though. 

Cc @t3dotgg -- any thoughts on this?

markflorkowski:
After thinking some more on this, I think that the `threshold` mode gets too complicated, and if the user is looking for that functionality, they probably have something very specific in mind, and should use the hook to encode it themselves. 

I think for now, we should just have `manual` and `auto` modes. Keep the configuration object so we could expand on this later though.

Mr0Bread:
Okay, will adjust

Mr0Bread:
@markflorkowski Made adjustments. Added docs

docs/src/pages/api-reference/react.mdx
```diff
@@ -67,15 +67,17 @@ export const OurUploadDropzone = () => (
...
+| onClientUploadComplete | function           | No                      |                 | `{fileKey: string, fileUrl: string}[]` callback function when upload is completed       |
+| onUploadError          | function           | No                      |                 | callback function when upload fails                                                     |
+| onUploadProgress       | function           | No                      | Added in `v5.1` | callback function for upload progress                                                   |
+| onUploadBegin          | function           | No                      | Added in `v5.4` | callback function for upload begin                                                      |
+| config                 | object             | No                      | Added in `v5.5` | object to pass additional configuration for dropzone                                    |
+| config.mode            | 'auto' \| 'manual' | No                      | Added in `v5.5` | mode of dropzone. 'manual' is default one. 'auto' triggers upload right after selection |
```
markflorkowski, line 80: Since this is the only property in config right now, this is fine, we may want a separate table in the future.

examples/appdir/src/app/page.tsx
```diff
@@ -65,6 +65,9 @@ export default function Home() {
             alert(`ERROR! ${error.message}`);
           }}
           className="ut-label:text-lg ut-label:text-cyan-900 ut-allowed-content:text-base ut-button:bg-orange-500"
+          config={{
+            mode: "manual",
```
markflorkowski, line 69:
```suggestion
            mode: "manual", // change to 'auto' to upload files immediately on selection.
```

markflorkowski (approved):

docs/src/pages/api-reference/react.mdx
juliusmarminge, line 80: v5.4 isn't out yet, so we should verify what version this will land on before merging

t3dotgg (approved):

PR #280, opened 2023-08-24
feat: ExpressJS Support

Mr0Bread:
- Created `createExpressRouter`
- Added `express` entrypoint to uploadthing package
- Adjusted build options for `shared` and `mime-types` packages to fix the error
- Explicitly setting Content-Type on request to ensure that Express will parse body properly
- Scaffolded example of Express + React
- Added docs


Mr0Bread:
@markflorkowski Please take a look once you have time. I have two concerns:
- Doc structure. I made section with sub-section: Backend Adapters -> Express. I'm planning on adding adapters for other backend libs like Fastify, so thought that storing relevant topics under single sup-topic makes sense
- Requirement to register `json` middleware so that body is parsed properly. Would you prefer UT handling such thing? As far as I know `trpc` is able to parse the body on it's own. Though not sure if we want to introduce so much functionality to UT. I guess `json` middleware is common thing among Express-based apps

markflorkowski:
> * I'm planning on adding adapters for other backend libs like Fastify, so thought that storing relevant topics under single sup-topic makes sense

Agreed. You can put another subsection "coming soon" if you feel like the nesting is weird with only the one sub-subsection, but I don't think it is a big deal. 

> Requirement to register json middleware so that body is parsed properly....

`bodyParser` has caused me so much headache in the past. I think it is best to not rely on the user registering another middleware if possible. I believe that TRPC accomplishes this by checking for `req.body` -- meaning that it has already been parsed, and if it is not there parsing on their own.

packages/uploadthing/src/internal/express.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  const getBuildPerms = buildPermissionsInfoHandler<TRouter>(opts);
+
+  router.get("/", (req, res) => {
+    res.status(200);
+    res.setHeader("x-uploadthing-version", UPLOADTHING_VERSION);
+    res.send(JSON.stringify(getBuildPerms()));
+  });
+
+  return router;
```
markflorkowski, line 67: Since we require the api to be on a specific `/api/uploadthing` endpoint, could we return that here, and bypass the user needing to add it when they register the middleware?

pnpm-workspace.yaml
```diff
@@ -2,4 +2,5 @@ packages:
   - "packages/config/*"
   - "packages/*"
   - "examples/*"
+  - "examples/express/client"
```
markflorkowski, line 5: I think changing L#4 to `"examples/**` would be the generic way to accomplish this?

packages/uploadthing/src/internal/express.ts
Mr0Bread: Does not look like so. Btw, TRPC also require user to specify the endpoint
image

I checked Express docs and could not find anything useful to address that

Mr0Bread:
@markflorkowski Implemented body parsing similar to one that is being used in TRPC with little tweaks in terms of error handling and parsing string into JSON straight away

Mr0Bread:
I reverted changes to pnpm-workspace.yaml back because it caused issue during `pnpm lint` command. I didn't spend any time figuring out it's nature, decided to just go with the option that works accurately

markflorkowski:
> I reverted changes to pnpm-workspace.yaml back...

I'll look into this 👍 would like to avoid hardcoded paths whenever possible.

packages/uploadthing/src/internal/express.ts
markflorkowski: Fair point, I guess there is precedent for requiring the use to enter a specific endpoint. 

It's just a function, so you should be able to return an array, then spread it into the app.use(), but that would probably be most confusing to most users who would be using this.

docs/src/pages/backend-adapters/express.mdx
```diff
@@ -0,0 +1,166 @@
...
+
+All files uploaded to uploadthing are associated with a FileRoute. Think of a
+FileRoute similar to an endpoint, it has:
+
+- Permitted types ["image", "video", etc]
+- Max file size
+- (Optional) `middleware` to authenticate and tag requests
+- `onUploadComplete` callback for when uploads are completed
+
+```ts copy filename="src/uploadthing.ts"
```
juliusmarminge, line 62: i'm not sure what the best user experience is, but it seems like a hassle to duplicate this 40 line codeblock and make minor changes to it. maybe we should have 1 doc explaining how to make a file router for all adapters, and then in each adapters doc just explain what's special with this (different req/res in middlewares)?

adding new stuff requires a lot of doc changes if this is duplicated too much i feel like

docs/src/pages/backend-adapters/express.mdx
```diff
@@ -0,0 +1,166 @@
...
+import * as utExpress from "uploadthing/express";
+
+import { uploadRouter } from "./uploadthing";
+
+app.use(
+  "/api/uploadthing",
+  utExpress.createExpressRouter({
+    router: uploadRouter,
+  }),
+);
```
juliusmarminge, line 138: we never use `import *` anywhere else

```suggestion
import { createExpressRouter } from "uploadthing/express";

import { uploadRouter } from "./uploadthing";

app.use(
  "/api/uploadthing",
  createExpressRouter({
    router: uploadRouter,
  }),
);
```

examples/express/client/src/App.tsx
```diff
@@ -0,0 +1,104 @@
+import { generateComponents } from "@uploadthing/react";
+
+import type { OurFileRouter } from "../../uploadthing";
+
+const { UploadButton, UploadDropzone } = generateComponents<OurFileRouter>();
+
+export function App() {
```
juliusmarminge, line 7: again, i feel like this is too much duplication.

i think we should rethink our examples. every single examples doesn't have to show all use cases, just have minimal ones for setting up the file route and serving it, and consuming a single endpoint would be fine.

then we can have specific examples for different use cases (preferably using next.js???). that way it's way more easily browsed for users coming to our repo looking for an example on how to do certain things.

tldr; more, contrived examples rather than a few bigger ones provides a better use experience. i was actually thinking of doing this but then i went on vacation. might get to it later today or early next week.

examples/express/index.ts
```diff
@@ -0,0 +1,23 @@
+import "dotenv/config";
+
+import cors from "cors";
+import express from "express";
+
+import * as utExpress from "uploadthing/express";
```
juliusmarminge, line 6: again, named import over `import *` to match the rest

examples/express/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "dotenv": "^16.3.1",
+    "express": "^4.18.2",
+    "uploadthing": "*",
+    "zod": "^3.21.4"
+  },
+  "devDependencies": {
+    "@types/cors": "^2.8.13",
+    "@types/express": "^4.17.17",
+    "ts-node": "^10.9.1",
+    "ts-node-dev": "^2.0.0"
```
juliusmarminge, line 24: this gives me nightmares, can we use `tsx` instead

packages/uploadthing/src/internal/express.ts
```diff
@@ -0,0 +1,88 @@
...
+import { getStatusCodeFromError, UploadThingError } from "@uploadthing/shared";
+
+import { UPLOADTHING_VERSION } from "../constants";
+import { defaultErrorFormatter } from "./error-formatter";
+import type { RouterWithConfig } from "./handler";
+import { buildPermissionsInfoHandler, buildRequestHandler } from "./handler";
+import { getPostBody } from "./node-http/getBody";
+import type { FileRouter, inferErrorShape } from "./types";
+
+export const createExpressRouter = <TRouter extends FileRouter>(
```
juliusmarminge, line 12: the name `createExpressRouter` doesn't quite make sense to me? should it not be `createExpressHandler` or something? the next.js ones are named like that.

even if express uses the `Router` term, i think it would confuse some what the difference of a `FileRouter` and an `ExpressRouter` would be

packages/uploadthing/src/internal/express.ts
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    const { data } = bodyResult;
+
+    const response = await requestHandler({
+      req: {
+        ...req,
+        // Explicitly passing headers since
+        // without that they are being lost for some reason
+        headers: req.headers,
```
juliusmarminge, line 44: 👀

packages/uploadthing/src/internal/express.ts
juliusmarminge: > Btw, TRPC also require user to specify the endpoint

tRPC doesn't require you to serve your router on that endpoint though

> It's just a function, so you should be able to return an array, then spread it into the `app.use()`, but that would probably be most confusing to most users who would be using this.

That's what i was thinking. Not sure if it's better or worse UX

packages/uploadthing/src/internal/node-http/getBody.ts
```diff
@@ -0,0 +1,74 @@
+import type { IncomingMessage } from "http";
```
juliusmarminge, line 1: very nit (and we should probably activate a lint rule to enforce it), but [i prefer `node:http`](https://github.com/pingdotgg/uploadthing/pull/280/files#diff-d029790c4e22075502109367243c609551aa27e5cb20615a35d004299d104571R2)

```suggestion
import type { IncomingMessage } from "node:http";
```

pnpm-workspace.yaml
juliusmarminge: it's kinda hard to have nested pnpm workspaces actually, in trpc we use npm workspaces for the nested examples:

https://github.com/trpc/trpc/blob/64f28b6b612889f7785cbab4cab9d8cc1df70c99/examples/minimal-react/package.json#L5-L8

juliusmarminge (changes requested):
Bunch of these are probably not very related to this PR, just observations made when looking at what code is introduced when adding a new adapter thing.

docs/src/pages/backend-adapters/express.mdx
Mr0Bread: I literally had the varian you proposed initially, but Mark asked to refactor so that our middleware is distinguishable from other ones that potentially may be present in code.

How about I rename the function to something like `createUTExpressRouter` and adjust import accordingly?

packages/uploadthing/src/internal/node-http/getBody.ts
Mr0Bread: Okay, will adjust

docs/src/pages/backend-adapters/express.mdx
```diff
@@ -0,0 +1,166 @@
...
+    ```
+  </Tab>
+  <Tab>
+    ```bash copy 
+    pnpm install uploadthing
+    ```
+  </Tab>
+</Tabs>
+
+{/* prettier-ignore-end */}
```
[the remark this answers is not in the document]
Mr0Bread: Btw, need to double check, but I think we have other such cases, so probably should fix all of them to be consistent

docs/src/pages/backend-adapters/express.mdx
```diff
@@ -0,0 +1,166 @@
...
+
+### Creating your first FileRoute
+
+All files uploaded to uploadthing are associated with a FileRoute. Think of a
+FileRoute similar to an endpoint, it has:
+
+- Permitted types ["image", "video", etc]
+- Max file size
+- (Optional) `middleware` to authenticate and tag requests
+- `onUploadComplete` callback for when uploads are completed
```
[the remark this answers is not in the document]
Mr0Bread: I'm sure I copied it from somewhere in the docs. Need to find the spot

examples/express/client/src/App.tsx
Mr0Bread: Your point makes sense. In addition, considering your other comments, it seems we need a PR with many different improvements to docs and examples. Or separate ones for these two items

Personally, I had created example as development playground in the first place. I simply copied filerouter from other example and bootstrapped app just to be able to test new stuff

examples/express/package.json
Mr0Bread: Sure thing. Didn't know about it

packages/uploadthing/src/internal/express.ts
Mr0Bread: createUTExpressHandler?

packages/uploadthing/src/internal/express.ts
Mr0Bread: I thought about such option but decided to not suggest it as it looks too hacky to me.

Btw, are there any plans on allowing serving UT API on other endpoints? If there is such thing in mind, I would recommend current approach

Mr0Bread:
@markflorkowski @juliusmarminge made some improvements. Didn't touch docs much since it require major overhaul in overall. Renamed express function to `createUploadthingExpressHandler`

Could not make npm workspaces work properly, decided to go with pnpm workspaces, but simply limited nesting to single level

juliusmarminge:
> Didn't touch docs much since it requires major overhaul in overall from what I understood

Yea I'll try and get to that early next week

vin4raiz:
@Mr0Bread can you please share what the following steps mean - 

- Adjusted build options for shared and mime-types packages to fix the error
- Explicitly setting Content-Type on request to ensure that Express will parse body properly

I cannot find the above things in the docs. Am getting INVALID_CONTENT_TYPE error in postman. I think your steps might fix this for me

markflorkowski:
> @Mr0Bread can you please share what the following steps mean -
> 
> * Adjusted build options for shared and mime-types packages to fix the error
> * Explicitly setting Content-Type on request to ensure that Express will parse body properly
> 
> I cannot find the above things in the docs. Am getting INVALID_CONTENT_TYPE error in postman. I think your steps might fix this for me

@vin4raiz Those are referring to changes in the client code, which would have no effect on your experience with Postman. 

Please open an issue rather than replying on an old closed PR. Include a repro if possible so that we are able to better help you.

Issue #275, opened 2023-08-21
[trpc-prisma-websockets-starter] Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined


niemal:
### What happened?

I imported `utapi` inside a websocket tRPC route.

### What did you expect to happen?

No errors, normal function.

### Version

5.3.3

### Reproduction Steps

```bash
pnpm create next-app --example https://github.com/trpc/trpc --example-path examples/next-prisma-websockets-starter trpc-prisma-websockets-starter
cd trpc-prisma-websockets-starter
pnpm i
pnpm install uploadthing @uploadthing/react react-dropzone
pnpm run dev
```

Add the following lines of code:
  - `import { utapi } from 'uploadthing/server';`
    on line 11 of `src/server/routers/post.ts`
  - `await utapi.deleteFiles('test');`
    on line 63 of `src/server/routers/post.ts`

### Link to reproduction

https://github.com/niemal/trpc-prisma-websockets-starter

### Relevant log output

```bash
node:internal/errors:490
    ErrorCaptureStackTrace(err);
    ^

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in /tmp/test/trpc-prisma-websockets-starter/node_modules/.pnpm/uploadthing@5.3.3_zod@3.0.0/node_modules/@uploadthing/shared/package.json
    at __node_internal_captureLargerStackTrace (node:internal/errors:490:5)
    at new NodeError (node:internal/errors:399:5)
    at exportsNotFound (node:internal/modules/esm/resolve:361:10)
    at packageExportsResolve (node:internal/modules/esm/resolve:641:13)
    at resolveExports (node:internal/modules/cjs/loader:565:36)
    at Module._findPath (node:internal/modules/cjs/loader:634:31)
    at Module._resolveFilename (node:internal/modules/cjs/loader:1061:27)
    at u.default._resolveFilename (/tmp/test/trpc-prisma-websockets-starter/node_modules/.pnpm/@esbuild-kit+cjs-loader@2.4.2/node_modules/@esbuild-kit/cjs-loader/dist/index.js:1:1519)
    at Module._load (node:internal/modules/cjs/loader:920:27)
    at Module.require (node:internal/modules/cjs/loader:1141:19)
    at require (node:internal/modules/cjs/helpers:110:18)
[...]

PR #281, opened 2023-08-25
fix: exports of shared and mime-types

Mr0Bread:
Addresses #275 

Faced the same issue while working on ExpressJS adapter and example for it


markflorkowski (approved):
cc @t3dotgg

juliusmarminge:
Why should these have `cjs` outputs when the main package doesn't?

https://github.com/pingdotgg/uploadthing/blob/9b1e2b722f5a80d5bb6205491e753998e4420d6b/packages/config/tsup/index.js#L7

https://github.com/pingdotgg/uploadthing/blob/9b1e2b722f5a80d5bb6205491e753998e4420d6b/packages/uploadthing/tsup.config.ts#L6-L16

Mr0Bread:
Honestly, not sure why that error happens. Probably something to do with the resolution algorithm

juliusmarminge:
I added the cjs export to the root config. Doesn't hurt to have it everywhere?

Mr0Bread:
> I added the cjs export to the root config. Doesn't hurt to have it everywhere?

Seems reasonable

juliusmarminge (approved):

PR #283, opened 2023-08-28
chore: swap dev and build scripts to favor only packages/*

juliusmarminge:
Just me that feels like running dev/build should just run dev in packages/* ?

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,11 +8,11 @@
     "pnpm": ">=8.6.0"
   },
   "scripts": {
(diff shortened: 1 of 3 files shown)


Mr0Bread:
Btw, seems that Solid package does not have `dev` script

markflorkowski (approved):
This makes sense to me. Also wow how was the solid package missing a dev script....

t3dotgg (approved):

PR #286, opened 2023-08-28
fix(solid): component generation

### LINKED ISSUE #285: [Bug]: Solid QuickStart is very broken
### What happened?

Following the quickstart guide on Solid, there are quite a few issues:

## 1. Invalid components???

```ts
import { UploadButton, UploadDropzone } from "~/uploadthing/client";

export default function Home() {
  return (
    <main>
      <UploadButton endpoint="profileImage" />
      <UploadDropzone endpoint="profileImage" />
    </main>
  );
}
```

Yet, the page is displaying the Uploader component (twice):

CleanShot 2023-08-28 at 13 53 45@2x

Also note how the styling is pretty bad, not sure if this is intentional or not, but we don't seem to put any default font on?

## Wrong type in the file router:

Since Solid is using the /next entrypoint (not sure why), we get invalid type in our router metadata:

image

The correct type would be a standard `Request` type.

## Invalid file path for API route

In docs (https://docs.uploadthing.com/solid#create-an-api-route-using-the-filerouter), we say to create a file `src/api/uploadthing.ts`, this is however invalid as it should be `src/routes/api/uploadthing.ts`

## Missing dependency for dropzone

I think this might be the `solidjs-dropzone` package's fault, but it seems like they should include `attr-accept` as peerDep as it seems to not resolve that correctly:

CleanShot 2023-08-28 at 13 57 49@2x

The quickstart to produce a working example

### Version

```sh
"@uploadthing/solid": "^5.1.2",
"solid-js": "^1.7.11",
"solid-start": "^0.3.4",
"solidjs-dropzone": "^1.0.0",
"uploadthing": "^5.3.3"
```

### Reproduction Steps

1. Create a bare solid-start app
2. Follow the uploadthing solid guide: https://docs.uploadthing.com/solid
  - Install deps
  - Create router
  - Generate components
  - Create API route
3. Start dev server
4. See a pretty broken page

### Link to reproduction

https://github.com/juliusmarminge/solid-uploadthing-example

### Relevant log output

_No response_

### Code of Conduct

juliusmarminge:
Closes #285

- Fixes so that `generateComponents` returns the correct components
- Minifies example
- Fixes docs

### DIFF
--- a/examples/minimal-solidstart/src/server/uploadthing.ts
+++ b/examples/minimal-solidstart/src/server/uploadthing.ts
@@ -19,7 +19,7 @@ export const uploadRouter = {
     },
   })
     .middleware(({ req }) => {
-      // Check some condition based on the incoming requrest
+      // Check some condition based on the incoming request
       req;
       //^?
       // if (!req.headers.get("x-some-header")) {


docs/src/pages/solid.mdx
```diff
@@ -1,58 +1,40 @@
...
-</Callout>
 </Steps>
 
 ## Set Up A FileRouter
 
 <Steps>
 ### Creating your first FileRoute
 
+<Callout>
+  For more details on how to create a file router, see the [router](/router)
```
juliusmarminge, line 34: need to write this page, imo every framework guide don't have to have this. just write 1 page and list the changes to each entrypoint

docs/src/pages/solid.mdx
```diff
@@ -1,58 +1,40 @@
...
-  </Tab>
-  <Tab>
-    ```bash copy 
-    pnpm install uploadthing @uploadthing/solid solidjs-dropzone
-    ```
-  </Tab>
-</Tabs>
-
-{/* prettier-ignore-end */}
+```bash copy npm2yarn
```
markflorkowski, line 9: love this

docs/src/pages/solid.mdx
markflorkowski: Does `https://docs.uploadthing.com/api-reference/server#file-routes` have most of the information needed here?

docs/src/pages/solid.mdx
```diff
@@ -258,44 +244,89 @@ const Home: VoidComponent = () => {
...
+  return (
+    <main>
+      <input
+        type="file"
+        onChange={async (e) => {
+          const file = e.target.files?.[0];
+          if (!file) return;
+
+          // Do something with files
+          const compressed = await compress(file);
```
markflorkowski, line 321: since this is copyable code block, we should probably either have a dummy `compress` function, or do something different here.

examples/solidstart-minimal/src/server/uploadthing.ts
```diff
@@ -0,0 +1,37 @@
+import { createUploadthing } from "uploadthing/server";
+import type { FileRouter } from "uploadthing/server";
+
+const f = createUploadthing();
+
+/**
+ * This is your Uploadthing file router. For more information:
+ * @see https://docs.uploadthing.com/router
```
markflorkowski, line 8: `https://docs.uploadthing.com/api-reference/server#file-routes`

docs/src/pages/solid.mdx
juliusmarminge: yea that looks solid!

examples/solidstart-minimal/package.json
```diff
@@ -6,16 +6,15 @@
     "dev": "solid-start dev --port 9898",
     "start": "solid-start start --port 9898",
     "build": "solid-start build",
-    "lint": "eslint \"**/*.{ts,tsx}\""
+    "lint": "eslint src"
```
markflorkowski, line 9: thank you for this 😅

examples/solidstart-minimal/src/server/uploadthing.ts
```diff
@@ -0,0 +1,37 @@
...
+      maxFileSize: "4MB",
+      maxFileCount: 4,
+    },
+    video: {
+      maxFileSize: "16MB",
+    },
+  })
+    .middleware(({ req }) => {
+      // Check some condition based on the incoming requrest
+      if (!req.headers.get("x-some-header")) {
```
markflorkowski, line 23: We probably should remove this because the examples will be broken by default?

markflorkowski (approved):

examples/solidstart-minimal/src/server/uploadthing.ts
juliusmarminge: commented it

PR #287, opened 2023-08-28
chore: examples revamp

juliusmarminge:
Aims to improve explorability in our examples by making more, contrived ones instead of fewer bigger ones


pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.1'
+lockfileVersion: '6.0'
```
Mr0Bread, line 1: I noticed that I'm not a single one who faced that pnpm changes version of lockfile. Maybe we should stop changing it back to 6.1 and let it be 6.0. Is there any different between thsse two?

Mr0Bread:
How about simple example with DB usage? Like Drizzle + SQLite where file key is stored in DB after upload?

markflorkowski:
> How about simple example with DB usage? Like Drizzle + SQLite where file key is stored in DB after upload?

I actually like this, because we expect the users to be doing this, but never actually give an example of what that would look like.

pnpm-lock.yaml
markflorkowski: I don't think we have been manually changing it, I think different contributor's environments are automatically changing it whenever they change deps. What version are you on?

```➜  ~ pnpm --version
8.6.1
```

Mr0Bread: Theo changed it back to 6.1 not so long ago

Mr0Bread: Personally I had been on almost latest versions of pnpm. Updated to latest couple of days ago

examples/minimal-appdir/README.md
```diff
@@ -0,0 +1,16 @@
+# Minimal Next.js App Router example for UploadThing
+
+[![stackblitz-logo](https://res.cloudinary.com/practicaldev/image/fetch/s--jByI8PG5--/c_fill,f_auto,fl_progressive,h_320,q_auto,w_320/https://dev-to-uploads.s3.amazonaws.com/uploads/organization/profile_image/5299/3d6e2f74-98b6-437d-99d9-8222bb7ffad9.png) Open in Stackblitz](https://stackblitz.com/github/pingdotgg/uploadthing/tree/main/examples/minimal-appdir)
```
juliusmarminge, line 3: replace `tree/main` with `tree/examples-revamp` if you wanna try the link now

examples/minimal-appdir/.env.example
```diff
@@ -0,0 +1,2 @@
+# Go to https://uploadthing.com/dashboard to get your API secret
+UPLOADTHING_SECRET=secret
```
markflorkowski, line 2: nit - use `sk_live_` instead

markflorkowski: nit - use `sk_live_` instead

examples/minimal-appdir/README.md
markflorkowski: This was a good idea.

examples/minimal-appdir/next.config.js
```diff
@@ -1,7 +1,5 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
-  reactStrictMode: true,
```
markflorkowski, line 3: no longer needed?

examples/with-react-image-crop/src/app/page.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export default function Page() {
+  const [imgSrc, setImgSrc] = useState("");
+
+  return (
+    <div>
+      <div>
+        <h1>Image Cropper</h1>
+        <p>
+          This is a simple example of using the{" "}
```
markflorkowski, line 15: nit, but can these be template strings to avoid `{" "}`

markflorkowski (approved):

examples/minimal-appdir/next.config.js
juliusmarminge: It's default in appdir

PR #290, opened 2023-08-29
feat: Astro support

Mr0Bread:
contains some hacks to make code work with Astro and import.meta stuff

Points of concerns:
- requires 'output' to be set to 'hybrid' or 'server'. Without that Endpoint will be considered static and won't provide access to headers(and other dynamic stuff). Using non-standard output required using an adapter. Though I guess any non-static Astro site will be using such settings? 🤔 
- implemented hacks to support import.meta.env in addition to process.env. Maybe there are better ways to do it?


Mr0Bread:
@juliusmarminge Should I get to docs straight away or it would be better to wait for you to make doc rework?

packages/uploadthing/tsconfig.json
```diff
@@ -14,7 +14,8 @@
     "noUnusedParameters": false,
     "preserveWatchOutput": true,
     "skipLibCheck": true,
-    "strict": true
+    "strict": true,
+    "module": "ES2020"
```
Mr0Bread, line 18: I also set module to "ES2020" so that TypeScript won't complain about using import.meta. Is it fine or may cause some issues with? I mean current existing examples are building fine, but maybe there are edge cases I'm not aware about

PR #292, opened 2023-08-29
feat: astro support w/ example

juliusmarminge:
I don't think we need a specific entrypoint here, considering Astro is just Request/Response, they should be using the /server entrypoint.

Not quite sure what to do with the env situation, I wrapped the only place where we prefer process.env in a try-catch, and falling back to import.meta.env. However, for the others, where we prefer the config route, I did not do this.

We could build an object like this:

```ts
export function buildEnv() {
  try {
    return {
      UPLOADTHING_SECRET: process.env.UPLOADTHING_SECRET,
      ...
    } as const;
  } catch {
    return {
      UPLOADTHING_SECRET: import.meta.env.UPLOADTHING_SECRET,
      ...
    } as const;
  }
}
```

which should let us safely access env vars like `buildEnv().UPLOADTHING_SECRET`, without them getting treeshaked during static analysis (see https://github.com/pingdotgg/uploadthing/pull/292#discussion_r1309320797)


packages/shared/src/utils.ts
```diff
@@ -105,7 +105,14 @@ export function getTypeFromFileName(
...
 export function generateUploadThingURL(path: `/${string}`) {
-  const host = process.env.CUSTOM_INFRA_URL ?? "https://uploadthing.com";
+  let host = "https://uploadthing.com";
+  try {
+    host = process.env.CUSTOM_INFRA_URL ?? host;
+  } catch (e) {
+    // @ts-expect-error - import.meta is dumb
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
+    host = import.meta.env.CUSTOM_INFRA_URL ?? host;
+  }
```
juliusmarminge, line 115: this is the only lib change we need to make

i tried making this a reusable helper:

image

but it didn't work as the vars are statically analyzed so this function is always returning undefined,

Mr0Bread:
Btw, similar to Astro we can add docs and examples on how to add UT to any Request/Response based library/framework

juliusmarminge:
Getting this error when trying out in an app outside the monorepo:

CleanShot 2023-09-01 at 09 31 03@2x

Mr0Bread:
Well, we can simply refactor our component just like the message suggests
```
import dropzone from 'react-dropzone'
const { useDropzone } = dropzone;
```

juliusmarminge:
> Well, we can simply refactor our component just like the message suggests

The default export is their `<Dropzone />` component so I don't think that would actually work in all environments though

markflorkowski (approved):

markflorkowski:
> We could build an object like this ...

I think this is a pretty reasonable way of doing it, though there might need to be one more layer of try/catch? Since `import.meta.env` does not exist in other environments, though I am not sure you could actually land in that case... 🤔

packages/shared/src/utils.ts
```diff
@@ -105,7 +105,15 @@ export function getTypeFromFileName(
...
-  const host = process.env.CUSTOM_INFRA_URL ?? "https://uploadthing.com";
+  let host = "https://uploadthing.com";
+
+  if (typeof process !== "undefined") {
+    host = process.env.CUSTOM_INFRA_URL ?? host;
+  } else {
+    // @ts-expect-error - import.meta is dumb
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-assignment
+    host = import.meta.env.CUSTOM_INFRA_URL ?? host;
+  }
```
juliusmarminge, line 116: is this good enough for now @markflorkowski until we decide on a more general solution?

markflorkowski: Yes. Ideally I'd like to remove all direct requirements for environment variables at all, but until we do that, this is fine.

juliusmarminge: Yea totally agree

PR #294, opened 2023-08-29
chore(actions): Don't hardcode packages to release in release action

markflorkowski:

PR #295, opened 2023-08-29
fix: publish config in Vue package

Mr0Bread:


Mr0Bread:
@markflorkowski added publish config to vue package. Is there anything else required?

markflorkowski:
> @markflorkowski added publish config to vue package. Is there anything else required?

```
@examples/nuxt has a dependency on uploadthing@5.3.3 but the version of uploadthing in the repo is 5.4.0 which is not within range of the depended on version, please update the dependency version
```

The other thing I need to fix is the action that does the release ( #294 )

juliusmarminge (approved):

PR #296, opened 2023-08-29
feat: ExpressJS support

Mr0Bread:
Express Adapter


Mr0Bread:
Also would like having this approved first, so that I can make Fastify docs in the manner that we agree on here

Mr0Bread:
@markflorkowski Can you also take a look once you have time?

docs/src/pages/_meta.json
```diff
@@ -3,6 +3,7 @@
   "getting-started": "Getting Started",
   "nextjs": "Next.js Setup",
   "solid": "SolidStart Setup",
+  "backend-adapters": "Backend Adapters",
```
juliusmarminge, line 6: next and solidstart can go under backend adapters too, and then they can link to `/api-reference/react` (and solid)

juliusmarminge: we can do this restructuring later though

examples/express/client/package.json
juliusmarminge, line 1: unsure about the example

should we ship a vite app to every adapter, or should it just show the backend usage?

packages/uploadthing/src/internal/node-http/getBody.ts
```diff
@@ -0,0 +1,74 @@
...
+}): Promise<BodyResult> {
+  const { req, maxBodySize = Infinity } = opts;
+  return new Promise((resolve) => {
+    if ("body" in req) {
+      const isJsonType = req.headers["content-type"] === "application/json";
+
+      if (!isJsonType) {
+        resolve({
+          ok: false,
+          error: new Error("INVALID_CONTENT_TYPE"),
```
juliusmarminge, line 24: i think we always wanna throw `UploadThingError`'s

examples/express/client/package.json
Mr0Bread: Personally I would ship it with every adapter. In such way it serves not only as example, but also development playground

packages/uploadthing/src/internal/node-http/getBody.ts
Mr0Bread: Fixed

docs/src/pages/_meta.json
t3dotgg: Don't worry about structure too much until we have the community repo setup, I'll do an overhaul

examples/express/client/package.json
t3dotgg: I think we desperately need a separate dev playground. Leave this for now as it's super convenient, I'll prioritize something better for dev in near future

t3dotgg (approved):

t3dotgg (commented):
Wait actually this should probably be a separate package, thoughts @juliusmarminge ?

Mr0Bread:
> Wait actually this should probably be a separate package, thoughts @juliusmarminge ?

If you plan on making adapter for next a separate package, then I guess every adapter, that requires custom code to work, have to be in separate package. For instance, Fastify requires custom adapter and can be moved to separate package

juliusmarminge:
> Wait actually this should probably be a separate package, thoughts @juliusmarminge ?

Why do you think so?

packages/uploadthing/src/internal/express.ts
```diff
@@ -34,16 +32,10 @@ export const createUploadthingExpressHandler = <TRouter extends FileRouter>(
...
-      req: {
-        ...req,
-        // Explicitly passing headers since
-        // without that they are being lost for some reason
-        headers: req.headers,
-        json: () => new Promise((resolve) => resolve(data)),
-      },
+      req: Object.assign(req, {
+        json: () => Promise.resolve(bodyResult.data),
+      }),
```
juliusmarminge, line 38: this seems to fix the header situation @Mr0Bread

juliusmarminge (approved):

packages/uploadthing/src/internal/express.ts
Mr0Bread: Cool, thank you

PR #298, opened 2023-08-29
feat: Solid custom styling and content

Mr0Bread:


docs/src/pages/theming.mdx
```diff
@@ -103,51 +103,109 @@ accepts an object with keys that correspond to elements of component. The
...
-  appearance?: {
-    container?: StyleField;
-    uploadIcon?: StyleField;
-    label?: StyleField;
-    allowedContent?: StyleField;
-    button?: StyleField;
-  };
-};
-```
+<Tabs items={['ReactJS', 'SolidJS']}>
```
juliusmarminge, line 106: are these tabs necessary? the types should be the same with the exception of signal-based frameworks will use getters instead of direct attributes?

docs/src/pages/theming.mdx
```diff
@@ -427,19 +485,38 @@ If you need, you can pass classes directly to specific elements of component or
...
-      return `custom-button ${
-        ready ? "custom-button-ready" : "custom-button-not-ready"
-      } ${isUploading ? "custom-button-uploading" : ""}`;
-    },
-    container: "custom-container",
-    allowedContent: "custom-allowed-content",
-  }}
-/>
-```
+<Tabs items={["ReactJS", "SolidJS"]}>
```
juliusmarminge, line 488: same as above

docs/src/pages/theming.mdx
```diff
@@ -548,26 +625,52 @@ If you need, you can pass inline styles directly to specific elements of
...
-    container: {
-      marginTop: "1rem",
-    },
-    allowedContent: {
-      color: "#a1a1aa",
-    },
-  }}
-/>
-```
+<Tabs items={["ReactJS", "SolidJS"]}>
```
juliusmarminge, line 628: same as above

docs/src/pages/theming.mdx
```diff
@@ -646,27 +749,56 @@ component and will return a `CSSProperties` object
...
-  /* rest of props */
-  content?: {
-    uploadIcon?: ContentField;
-    label?: ContentField;
-    allowedContent?: ContentField;
-    button?: ContentField;
-  };
-};
-```
+<Tabs items={['ReactJS', 'SolidJS']}>
```
juliusmarminge, line 752: same as above

docs/src/pages/theming.mdx
```diff
@@ -677,22 +809,46 @@ type UploadDropzoneProps = {
...
-    },
-    allowedContent({ ready, fileTypes, isUploading }) {
-      if (!ready) return "Checking what you allow";
-      if (isUploading) return "Seems like stuff is uploading";
-      return `Stuff you can upload: ${fileTypes.join(", ")}`;
-    },
-  }}
-/>
-```
+<Tabs items={['ReactJS', 'SolidJS']}>
```
juliusmarminge, line 812: same as above

packages/solid/src/components/shared.tsx
```diff
@@ -0,0 +1,77 @@
...
+  20: "after:w-[20%]",
+  30: "after:w-[30%]",
+  40: "after:w-[40%]",
+  50: "after:w-[50%]",
+  60: "after:w-[60%]",
+  70: "after:w-[70%]",
+  80: "after:w-[80%]",
+  90: "after:w-[90%]",
+  100: "after:w-[100%]",
+};
```
juliusmarminge, line 61: all of this looks like a 1:1 copy of the react package? can we put it in `uploadthing/client` maybe and export with an `@internal` jsdoc note?

packages/solid/src/utils/styles.ts
juliusmarminge, line 1: this is also a 1:1 copy of the react implementation? can also be shared from `uploadthing/client`

docs/src/pages/theming.mdx
Mr0Bread: Well, difference is small, but still a difference. I wanted to make it clear for the user. Solid is signal based and also have bit different typedefs in couple of places. The main difference is that it leverages getters just like you said. How would you handle this case?

packages/solid/src/components/shared.tsx
Mr0Bread: Doesn't it make sense to move it to `@uploadthing/shared` then? I mean, to me it seems this package is exactly for shared stuff

juliusmarminge: Yea but there's declaration support issues with the shared package (see https://github.com/pingdotgg/uploadthing/pull/272). But might be better to put it in shared first and then we'll move it later if we decide we have to

Mr0Bread: Hm, if there are issues with the shared package, then I would prefer going with more robust option. In addition, I can see that we already have several utility functions inside of `uploadthing/client`. Probably, I should put style-related stuff in separate file and just re-export it from `client.ts` to not overload single file. What do you think?
image

Mr0Bread:
@juliusmarminge Moved the styling logic and similar utility functions to `uploadthing/package`. I also had to make some more generic typing since React and Solid have different types for Element and CSSProperties. I wonder what do you think about following 
```
styleFieldToCssObject<"solid", ButtonStyleFieldCallbackArgs>(
          $props.appearance?.button,
          styleFieldArg,
     )
```
Ideally, second generic argument should be inferred from second function argument type, but not sure how to achieve such behavior

juliusmarminge:
I think i fixed so the generics are inferred

packages/uploadthing/src/internal/client.ts
```diff
@@ -63,6 +63,11 @@ export const progressWidths: Record<number, string> = {
   100: "after:w-[100%]",
 };
 
+type MinCallbackArg = { ready: boolean | (() => boolean) };
+type inferRuntime<T extends MinCallbackArg> = T["ready"] extends boolean
+  ? "react"
+  : "solid";
```
Mr0Bread, line 69: Oh, that's smart

juliusmarminge: Not sure how well it scales though if we add Vue and Svelte support 😅 

There's probably a more universal way but this was the first idea i got

Mr0Bread: Yeah, you are right, but I guess we can deal with it later. Not sure about Svelte PR, but the Vue one does not have custom styling feature

Though looking at your implementation, I have an idea. Is there a way to define a property on type, that will be hidden from user?

I mean can we have something like following?

```tsx
type SolidButtonProps = {
  /**
  * @private
  */
  __runtime: "solid";
  ...
}

type ReactButtonProps = {
  /**
  * @private
  */
  __runtime: "react";
  ...
}

type Runtime<T extends { __runtime: "react" | "solid" }> = T["__runtime"] extends "react" ? "react" : "solid"
```

With such we could specify type of runtime right on type of the callback args. Hope the idea is clear

Mr0Bread:
@juliusmarminge I implemented something like I mentioned previously. User won't see "__runtime" in autocomplete suggestions and it is not part of the arguments object in runtime
image

Though now I'm using `as` for callback arguments objects. If some changes contents to invalid ones, TS won't be able to spot an error
image

I was thinking about something like following:
```ts
type DropzoneStyleFieldCallbackArgs = {
  __runtime: "solid";
  ready: () => boolean;
  isUploading: () => boolean;
  uploadProgress: () => number;
  fileTypes: () => string[];
  isDragActive: () => boolean;
};
type RuntimeCallbackArgs = Omit<DropzoneStyleFieldCallbackArgs, "__runtime">;

const runtimeArg: RuntimeCallbackArgs = {
    ready: ready,
    isUploading: uploadedThing.isUploading,
    uploadProgress: uploadProgress,
    fileTypes: () => fileInfo().fileTypes,
    isDragActive: () => isDragActive,
  }

  const styleFieldArg = {
    ...runtimeArg
  } as DropzoneStyleFieldCallbackArgs;
```

With something like above we can be sure that runtime arguments are typesafe and no unexpected changes are going to be merged

docs/src/pages/theming.mdx
juliusmarminge: we could probably just have a callout in the beginning saying that for signal-based frameworks the attributes are getters, but maybe tabs are fine. Mostly just seems like a hassle to maintain the types if you need to change in many places.

CSSProperties vs JSX.CSSProperties I'd say is just fine to refer to as CSSProperties

packages/solid/src/components/shared.tsx
juliusmarminge: >  I should put style-related stuff in separate file and just re-export it from client.ts to not overload single file. What do you think?

yea that sounds reasonable

examples/minimal-solidstart/package.json
```diff
@@ -6,7 +6,8 @@
     "dev": "solid-start dev --port 9898",
     "start": "solid-start start --port 9898",
     "build": "solid-start build",
-    "lint": "eslint . --ext .ts,.tsx"
+    "lint": "eslint . --ext .ts,.tsx",
+    "typecheck": "tsc --noEmit"
```
juliusmarminge, line 10: I added this since it was missing, and it seems like we got some issues. Not sure if it's related to this PR though:

juliusmarminge (approved):
Good to merge when the typecheck script is resolved

examples/minimal-solidstart/package.json
Mr0Bread: My guess is that it is because we have imports from both `react` and `solid` in single file `internal/component-theming.ts`
image

And since it is being imported in solid components, which are imported in solid-start example, it results in such an issues

juliusmarminge: I dont think so since we do that on the server for a bunch of frameworks?

Hmm would `import("react").Reactnode help?

Mr0Bread: Hm, import is an async function. How do you intend on using it? I mean we also refer to ReactNode only as a type, nothing to do with it during the runtime

Mr0Bread: And ESLint does not like ;D

Mr0Bread: Made it like that but error is still there
image

Mr0Bread: I enabled "skipLibCheck" for solid start and updated "solid-start" package to latest version. Locally it fixed the issue

PR #306, opened 2023-08-30
chore: bump pnpm and set engine-strict

juliusmarminge:


.npmrc
```diff
@@ -2,4 +2,6 @@
 # packages as it casues issues with manypkg
 # @link https://pnpm.io/npmrc#prefer-workspace-packages
 save-workspace-protocol=false
-prefer-workspace-packages=true
\ No newline at end of file
+prefer-workspace-packages=true
+
```
markflorkowski, line 6: nit: extra space

markflorkowski (approved):

.npmrc
juliusmarminge: It was intentional to separate it from the other 2 that relates to the comment

markflorkowski: Ah fair. All good 👍 I mentally skip over comments😅

PR #309, opened 2023-08-31
docs: minor improvements to react api reference

juliusmarminge:


docs/src/pages/api-reference/react.mdx
```diff
@@ -1,6 +1,19 @@
 ## UploadButton
 
-{/* TODO: add a picture */}
+A simple button that opens the native file picker and uploads the selected
+files. The default button is shown below. See [Theming](/theming) on how to
+customize it.
+
+import { UploadButton } from "@uploadthing/react";
+
+<UploadButton
```
markflorkowski, line 9: Maybe wrap this in som padding or a div with a background? Looks a little weird on the page standalone

juliusmarminge: Hmm what do you mean? I didn't want to apply custom styling to it as it was meant to show the default button. I don't think it looks too bad?

CleanShot 2023-09-06 at 17 35 38

markflorkowski: It just looks a little close to the text above, I was thinking to pad it down. The background thing was because it is just floating in the middle of the page, but I think adding some space above would be sufficient to make it look a little less awkward.


image

juliusmarminge: Added some margin 
CleanShot 2023-09-07 at 10 26 39@2x

markflorkowski: It's silly, but I actually fell like this makes a big difference 😅

markflorkowski (approved):

PR #310, opened 2023-08-31
feat!: allow arbitrary callback urls [UPL-62]

juliusmarminge:
This removes the config option `callbackUrl` in favor of `Request.url`, to allow any callbackUrl instead of just `/api/uploadthing`.

- [x] docs

Should probably still be released as part of a major release, but it no longer requires the user to pass a url. We default to (process.env.VERCEL_URL / import.meta.env.VERCEL_URL / window.location.origin)/api/uploadthing.

Made some changes so that internally we always pass a URL object, asserting it's a valid absolute url. All internal functions should also require a URL be passed, but this required'ness should not be exposed to users

User can pass either just an origin, just a pathname, or a full url and we'll fallback to the closest one matching the default


packages/react/src/component.tsx
```diff
@@ -29,11 +31,23 @@ export function Uploader<TRouter extends FileRouter>(
...
 }
 
-export function generateComponents<TRouter extends FileRouter>() {
+export function generateComponents<TRouter extends FileRouter>(initOpts?: {
+  /**
+   * The URL where you expose your UploadThing router.
+   * @default `/api/uploadthing`
+   */
+  url: string;
+}) {
```
juliusmarminge, line 40: having this optional makes for quite a few codepaths having weird optional chains / nullish coalescing. to me it makes sense that this is required, but it would be a breaking change that would require a major bump

packages/react/src/component.tsx
```diff
@@ -29,11 +31,23 @@ export function Uploader<TRouter extends FileRouter>(
   );
 }
 
-export function generateComponents<TRouter extends FileRouter>() {
+export function generateComponents<TRouter extends FileRouter>(initOpts?: {
+  /**
+   * The URL where you expose your UploadThing router.
+   * @default `/api/uploadthing`
```
juliusmarminge, line 37: For future proofing's sake it might be good to require this be an absolute url so we can SSR etc

docs/src/pages/getting-started.mdx
```diff
@@ -41,14 +41,6 @@ import { Tab, Tabs } from "nextra-theme-docs";
 UPLOADTHING_SECRET=... # A secret key for your app (starts with sk_live_)
 ```
 
-<Callout type="info">
```
markflorkowski, line 44: This is a really nice benefit.

docs/src/pages/nextjs/appdir.mdx
```diff
@@ -53,12 +53,6 @@ export type OurFileRouter = typeof ourFileRouter;
 
 ### Create a Next.js API route using the FileRouter
 
-<Callout type="warning">
```
markflorkowski, line 56: Maybe change this to an `info` block -- we recommend placing this at `/api/uploadthing`...

docs/src/pages/nextjs/pagedir.mdx
```diff
@@ -71,12 +71,6 @@ export type OurFileRouter = typeof ourFileRouter;
 
 ### Create a Next.js API route using the FileRouter
 
-<Callout type="warning">
```
markflorkowski, line 74: same as above

docs/src/pages/solid.mdx
```diff
@@ -123,21 +123,13 @@ export type OurFileRouter = typeof uploadRouter;
 
 ### Create an API route using the FileRouter
 
-<Callout type="warning">
```
markflorkowski, line 126: same as above

packages/react/src/component.tsx
markflorkowski: I am on with doing a major for this. I agree that it makes sense as required.

packages/uploadthing/src/internal/next/pagerouter.ts
```diff
@@ -26,8 +26,12 @@ export const createNextPageApiHandler = <TRouter extends FileRouter>(
       return;
     }
 
+    const proto = (req.headers["x-forwarded-proto"] as string) ?? "http";
```
markflorkowski, line 29: are we certain this always exists?

juliusmarminge: It does on Vercel, might be cases that slips through though 🤔 if only req.url returned the absolute url in pages/ like it does everywhere else :(

juliusmarminge: Otherwise we could expose the callbackUrl as config option on nextjs pages as a fallback🤔

packages/uploadthing/src/client.ts
```diff
@@ -102,14 +104,14 @@ export type UploadFileResponse = {
 
 export const DANGEROUS__uploadFiles = async <TRouter extends FileRouter>(
   opts: UploadFilesOptions<TRouter>,
-  config?: {
-    url?: string;
+  config: {
+    url: string;
   },
```
juliusmarminge, line 109: would it make sense to collapse this into a single parameter and put the url in the opts object?

markflorkowski: I think we had it as a config object for future expansion reasons? But you're right it does seem kind of silly

docs/src/pages/api-reference/react.mdx
```diff
@@ -29,26 +31,29 @@ export const OurUploadButton = () => (
...
+| Prop                   | Type     | Required                                                                                | Notes           | Description                                                                         |
+| :--------------------- | :------- | :-------------------------------------------------------------------------------------- | :-------------- | :---------------------------------------------------------------------------------- |
+| `<FileRouter>`         | generic  | Yes                                                                                     |                 | The type from the FileRouter you defined in your backend                            |
+| endpoint               | string   | Yes                                                                                     |                 | The name of the [route](./server#FileRouter) you want this button to upload to      |
+| url                    | string   | Yes, if imported from `@uploadthing/react`. No, if generated using `generateComponents` | Added in `v6.0` | The **absolute** url to where you server your uploadthing file router.              |
```
markflorkowski, line 38:
```suggestion
| url                    | string   | Yes, if imported from `@uploadthing/react`. No, if generated using `generateComponents` | Added in `v6.0` | The **absolute** url to where you are serving your uploadthing file router.              |
```

docs/src/pages/api-reference/react.mdx
```diff
@@ -67,17 +72,18 @@ export const OurUploadDropzone = () => (
...
+| :--------------------- | :----------------- | :-------------------------------------------------------------------------------------- | :-------------- | :-------------------------------------------------------------------------------------- |
+| `<FileRouter>`         | generic            | Yes                                                                                     |                 | The type from the FileRouter you defined in your backend                                |
+| endpoint               | string             | Yes                                                                                     |                 | The name of the [route](./server#FileRouter) you want this button to upload to          |
+| url                    | string             | Yes, if imported from `@uploadthing/react`. No, if generated using `generateComponents` | Added in `v6.0` | The **absolute** url to where you server your uploadthing file router.                  |
```
markflorkowski, line 79: `No *`  or `Yes *` with a corresponding note below would be better for the table formatting

docs/src/pages/nextjs/appdir.mdx
```diff
@@ -70,20 +69,26 @@ export const { GET, POST } = createNextRouteHandler({
...
-### Creating The UploadThing Components (optional)
+### Creating The UploadThing Components (recommended)
 
-Generating components lets you pass your generic `FileRouter` once and then have
-typesafe components everywhere, instead of having to pass the generic everytime
-you mount a component, but you can also import the components individually from
-`@uploadthing/react`.
+Generating components lets you pass your generic `FileRouter` and `url` once and
+then have typesafe components everywhere, instead of having to pass the generic
+and url everytime you mount a component, but you can also import the components
```
markflorkowski, line 76:
```
Generating components lets you pass your generic `FileRouter` and `url` once and
then have typesafe components everywhere, instead of having to pass them
everytime you mount a component, but you can also import the components
```

docs/src/pages/nextjs/pagedir.mdx
```diff
@@ -71,10 +71,9 @@ export type OurFileRouter = typeof ourFileRouter;
...
-  Note: This is the ONLY FILE WHERE THE PATH MATTERS. You need to serve this API
-  from `/api/uploadthing` as it is called via webhook to trigger
-  `onUploadComplete`.
+<Callout>
```
markflorkowski, line 74:
```suggestion
<Callout type="info">
```

docs/src/pages/nextjs/appdir.mdx
```diff
@@ -53,10 +53,9 @@ export type OurFileRouter = typeof ourFileRouter;
 
 ### Create a Next.js API route using the FileRouter
 
-<Callout type="warning">
-  Note: This is the ONLY FILE WHERE THE PATH MATTERS. You need to serve this API
-  from `/api/uploadthing` as it is called via webhook to trigger
-  `onUploadComplete`.
+<Callout>
```
markflorkowski, line 56:
```suggestion
<Callout type="info">
```

docs/src/pages/nextjs/pagedir.mdx
```diff
@@ -89,20 +88,26 @@ const handler = createNextPageApiHandler({
...
 
-### Creating The UploadThing Components (optional)
+### Creating The UploadThing Components (recommended)
 
-Generating components let's you pass your generic `FileRouter` once and then
-have typesafe components everywhere, instead of having to pass the generic
-everytime you mount a component, but you can also import the components
+Generating components lets you pass your generic `FileRouter` and `url` once and
+then have typesafe components everywhere, instead of having to pass the generic
+and url everytime you mount a component, but you can also import the components
```
markflorkowski, line 95:
```
Generating components lets you pass your generic `FileRouter` and `url` once and
then have typesafe components everywhere, instead of having to pass them
everytime you mount a component, but you can also import the components
```

docs/src/pages/solid.mdx
```diff
@@ -76,9 +76,9 @@ export type OurFileRouter = typeof uploadRouter;
...
-  **Note:** This is the ONLY FILE WHERE THE PATH MATTERS. You need to serve this
-  API from `/api/uploadthing` on your application.
+<Callout>
```
markflorkowski, line 79:
```suggestion
<Callout type ="info">
```

docs/src/pages/solid.mdx
```diff
@@ -88,29 +88,26 @@ import { uploadRouter } from "~/server/uploadthing";
...
-### Creating The UploadThing Components (optional)
+### Creating The UploadThing Components (recommended)
 
-Generating components let's you pass your generic `FileRouter` once and then
-have typesafe components everywhere, instead of having to pass the generic
-everytime you mount a component, but you can also import the components
-individually from `@uploadthing/solid`.
+Generating components lets you pass your generic `FileRouter` and `url` once and
+then have typesafe components everywhere, instead of having to pass the generic
+and url everytime you mount a component, but you can also import the components
```
markflorkowski, line 98:
```
Generating components lets you pass your generic `FileRouter` and `url` once and
then have typesafe components everywhere, instead of having to pass them
everytime you mount a component, but you can also import the components
```

examples/minimal-solidstart/src/utils/uploadthing.ts
```diff
@@ -2,7 +2,9 @@ import { generateComponents, generateSolidHelpers } from "@uploadthing/solid";
 
 import type { OurFileRouter } from "~/server/uploadthing";
 
+const url = "http://localhost:9898";
```
markflorkowski, line 5: nit: maybe add a comment here calling out that it only works in dev (even though it should be pretty obvious)

packages/react/src/component.tsx
```diff
@@ -29,11 +31,23 @@ export function Uploader<TRouter extends FileRouter>(
   );
 }
 
-export function generateComponents<TRouter extends FileRouter>() {
+export function generateComponents<TRouter extends FileRouter>(initOpts: {
+  /**
+   * Absolute URL to the UploadThing API endpoint
+   * @example http://localhost:3000/api/uploadthing
+   */
```
markflorkowski, line 38: silly thing, but also including a "real" url here as a secondary example may be nice, eg: 

``` https://www.example.com/api/uploadthing```

Just because localhost url could be confusing

File: packages/react/src/types.ts [hunk 19]
```diff
@@ -9,6 +9,11 @@ import type {
 export type UploadthingComponentProps<TRouter extends FileRouter> = {
   [TEndpoint in keyof TRouter]: {
     endpoint: TEndpoint;
+    /**
+     * Absolute URL to the UploadThing API endpoint
+     * @example http://localhost:3000/api/uploadthing
```
markflorkowski, line 14: same note about "real" example url

packages/react/src/useUploadThing.ts
```diff
@@ -37,7 +37,15 @@ const fatalClientError = (e: Error) =>
     cause: e,
   });
 
-export const INTERNAL_uploadthingHookGen = <TRouter extends FileRouter>() => {
+export const INTERNAL_uploadthingHookGen = <
+  TRouter extends FileRouter,
+>(initOpts: {
+  /**
+   * Absolute URL to the UploadThing API endpoint
+   * @example http://localhost:3000/api/uploadthing
```
markflorkowski, line 45: same note about "real" example url

packages/react/src/useUploadThing.ts
```diff
@@ -46,7 +54,10 @@ export const INTERNAL_uploadthingHookGen = <TRouter extends FileRouter>() => {
     const uploadProgress = useRef(0);
     const fileProgress = useRef<Map<string, number>>(new Map());
 
-    const permittedFileInfo = useEndpointMetadata(endpoint as string);
+    const permittedFileInfo = useEndpointMetadata(
+      initOpts.url,
+      endpoint as string,
```
markflorkowski, line 59: not really part of this, but is it possible for useEndpointMetadata to accept a `TEndpoint` directly, instead of needing to typecast?

packages/react/test/client-generator.test.ts
```diff
@@ -37,7 +37,9 @@ const router = {
     }),
 };
 
-const { useUploadThing } = generateReactHelpers<typeof router>();
+const { useUploadThing } = generateReactHelpers<typeof router>({
+  url: "0.0.0.0",
```
markflorkowski, line 41: Would this be a valid url?

packages/solid/src/component.tsx
```diff
@@ -65,12 +65,16 @@ const allowedContentTextLabelGenerator = (
 export type UploadthingComponentProps<TRouter extends FileRouter> = {
   [TEndpoint in keyof TRouter]: {
     endpoint: TEndpoint;
+    /**
+     * Absolute URL to the UploadThing API endpoint
+     * @example http://localhost:3000/api/uploadthing
```
markflorkowski, line 70: same note about "real" example url

packages/solid/src/component.tsx
```diff
@@ -334,16 +340,22 @@ function Spinner() {
   );
 }
 
-export function generateComponents<TRouter extends FileRouter>(url?: string) {
+export function generateComponents<TRouter extends FileRouter>(initOpts: {
+  /**
+   * Absolute URL to the UploadThing API endpoint
+   * @example http://localhost:3000/api/uploadthing
```
markflorkowski, line 346: same note about "real" example url

packages/solid/src/useUploadThing.ts
```diff
@@ -24,18 +24,25 @@ export type UseUploadthingProps = {
...
-  url?: string;
 };
 
-export const INTERNAL_uploadthingHookGen = <TRouter extends FileRouter>() => {
+export const INTERNAL_uploadthingHookGen = <
+  TRouter extends FileRouter,
+>(initOpts: {
+  /**
+   * Absolute URL to the UploadThing API endpoint
+   * @example http://localhost:3000/api/uploadthing
```
markflorkowski, line 34: same note about "real" example url.

packages/uploadthing/src/next-legacy.ts
```diff
@@ -38,8 +38,12 @@ export const createNextPageApiHandler = <TRouter extends FileRouter>(
       return;
     }
 
+    const proto = (req.headers["x-forwarded-proto"] as string) ?? "http";
+    const url = proto + "://" + req.headers.host + req.url;
```
markflorkowski, line 42: nit - can this be a template string? 😅

docs/src/pages/api-reference/react.mdx
t3dotgg: Not sure how I feel about naming this `url`, not clear what it's a URL for. Maybe `utEndpointURL` or something?

docs/src/pages/api-reference/react.mdx
```diff
@@ -89,11 +95,20 @@ components. It takes your File Router as a generic
...
+  generateReactHelpers<OurFileRouter>({
+    url: getBaseUrl() + "/api/uploadthing",
+  });
 ```
 
+### Configuration
+
+| Prop | Type   | Required | Notes           | Description                                               |
+| :--- | :----- | :------- | :-------------- | :-------------------------------------------------------- |
+| url  | string | Yes      | Added in `v6.0` | The **absolute** URL to the API handler you created above |
```
t3dotgg, line 110: I highly dislike making this required, will break all current users for no good reason. Also we can presume 95%+ of users will be on the local path of `/api/uploadthing`

t3dotgg (changes requested):
Not in love with requiring `url`, would like to set sane defaults since most users will be on happy path

docs/src/pages/api-reference/react.mdx
juliusmarminge: So you think we should have a default like

```ts
const defaultUrl = process.env.VERCEL_URL 
  ? `https://$${process.env.VERCEL_URL}/api/uploadthing`
  : `http://localhost:${process.env.PORT ?? 3000}/api/uploadthing`
```

Problem is then we use `process.env` on client which isn't supported in Vite based apps, so we'd need a workaround to handle that too.

@markflorkowski was fine doing a major for this, so a breaking change isn't too bad, especially since there's good migration docs. I also think it's reasonable to think people can provide this themselves, especially since we're removing the option on the backend where it can be automatically detected (think trpc links where url is required and you can deploy your trpc router anywhere, even if the majority will choose /api/[trpc])

packages/react/test/client-generator.test.ts
juliusmarminge: probably not - but the test isn't executing any requests so doesn't really matter?

packages/uploadthing/src/next-legacy.ts
juliusmarminge: if you prefer

docs/src/pages/api-reference/react.mdx
markflorkowski: > will break all current users

Major version bumps are allowed to do that IMO.

> Also we can presume 95%+ of users will be on the local path of /api/uploadthing

But we need the absolute URL, and there is not a consistent guaranteed way of constructing that in arbitrary environments. 


Perhaps we could fall back to old behaviour and throw terrible errors in the console if you try to just bump the version without adding the required config?

docs/src/pages/solid.mdx
```diff
@@ -106,11 +103,12 @@ import { generateComponents } from "@uploadthing/solid";
...
 import type { OurFileRouter } from "~/server/uploadthing";
 
-// URL is only needed for server side rendering
+// URL is only needed for server side rendering, or when your router
+// is deployed on a different path than `/api/uploadthing`
 const url = `http://localhost:${process.env.PORT ?? 3000}`;
 
 export const { UploadButton, UploadDropzone, Uploader } =
-  generateComponents<OurFileRouter>(url);
+  generateComponents<OurFileRouter>({ url });
```
juliusmarminge, line 111: should we "take a chance" here that no-one is using this option 😅 

otherwise the Solid setup would drift from the React setup.

or should we still release this in a major bump?

packages/react/src/useUploadThing.ts
```diff
@@ -40,7 +40,16 @@ const fatalClientError = (e: Error) =>
     cause: e,
   });
 
-export const INTERNAL_uploadthingHookGen = <TRouter extends FileRouter>() => {
+export const INTERNAL_uploadthingHookGen = <
```
juliusmarminge, line 43: internal function here the url is required. shouldn't touch user land

packages/uploadthing/src/client.ts
```diff
@@ -334,7 +334,9 @@ export function getFullApiUrl(maybeUrl?: string): URL {
     return "http://localhost:3000";
   })();
 
-  return maybeUrl?.startsWith("http")
-    ? new URL(maybeUrl)
-    : new URL(maybeUrl ?? "/api/uploadthing", base);
+  const url = new URL(maybeUrl ?? "/api/uploadthing", base);
```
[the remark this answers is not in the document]
juliusmarminge: Maybe `URL.canParse` and throw a good error if not?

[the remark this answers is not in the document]
juliusmarminge: Maybe not lol https://developer.mozilla.org/en-US/docs/Web/API/URL/canParse_static#browser_compatibility

juliusmarminge:
Tested on all the examples, backend adapters to make sure they all forward a valid URL

packages/uploadthing/src/client.ts
```diff
@@ -304,3 +311,37 @@ export const generateClientDropzoneAccept = (fileTypes: string[]) => {
...
+    if (import.meta.env?.VERCEL_URL) {
+      // @ts-expect-error - import meta is not defined in node
+      return `https://${import.meta.env.VERCEL_URL}`;
+    }
+
+    return "http://localhost:3000";
+  })();
+
+  try {
+    const url = new URL(maybeUrl ?? "/api/uploadthing", base);
```
markflorkowski, line 337: is it possible for `maybeUrl` to be a full url (not just path)? Not sure what happens if you supply both a full url and `base` to the URL constructor.

markflorkowski: It looks like it ignores the `base` if you provide a non-relative url.

packages/uploadthing/src/internal/handler.ts
```diff
@@ -282,6 +289,9 @@ export const buildRequestHandler = <TRouter extends FileRouter>(
           });
         }
 
+        const url = new URL(input.url ?? req.url ?? "");
```
markflorkowski, line 292: if this falls all the way through to the `""`, it will likely throw as an invalid URL

markflorkowski (commented):
Sorry it has taken me so long to review this 😅

packages/uploadthing/src/client.ts
juliusmarminge: yes if the url is absolute then it won't use the base

packages/uploadthing/src/internal/handler.ts
juliusmarminge: yea :/ this fallback was there before though so i didn't do anything about it.

we could wrap in try-catch and throw a better error maybe?

markflorkowski: Is it even possible for both `input.url` and `req.url` to be empty?  I am trying to figure out what case falls though in the first place. 

Regardless, try catch and a better error is probably best.

juliusmarminge: not really... the input.url is provided by us for adapters that doesn't provide an absolute url the req.url - so if the "" is reached we messed up

markflorkowski: Does TS get mad if the `""` is not there?

juliusmarminge: yes

markflorkowski: I guess just 

```
new URL((input.url ?? req.url) as string)
```

and wrap in a try catch just in case

juliusmarminge: done https://github.com/pingdotgg/uploadthing/pull/310/commits/eb2119fea654f89cb55d975faeb5e6c691037e7f

packages/uploadthing/src/internal/handler.ts
```diff
@@ -62,6 +61,9 @@ const fileCountLimitHit = (
 export type RouterWithConfig<TRouter extends FileRouter> = {
   router: TRouter;
   config?: {
+    /**
+     * @deprecated this option is deprecated and will be removed in a future version
```
markflorkowski, line 65: since this will not work, should we put an early escape to throw if using a deprecated input, or should we allow it to work and warn in the console?

juliusmarminge: why wont it work?

markflorkowski: 🤦I just can't read. You're right.

markflorkowski (approved):

t3dotgg (approved):

PR #313, opened 2023-08-31
fix: dev mode hook

Mr0Bread:
Basically, reused existing functionality with a bit of tweaks

Now, `pollForFileData` will actually resolve with the received file data. It is needed for the hook simulation

Flow is the following:
1. Uploading files
2. Polling for the file data
3. Simulating hook
4. Once all file promises resolved, the function `DANGEROUS__uploadFiles` is done working
5. Calling onUploadComplete on client


packages/shared/src/utils.ts
```diff
@@ -156,6 +156,8 @@ export async function pollForFileData(
     if (json.status !== "done") return null;
 
     await callback?.(json);
+
+    return json;
```
markflorkowski, line 160: 😅

packages/uploadthing/src/client.ts
```diff
@@ -259,12 +260,68 @@ export const DANGEROUS__uploadFiles = async <TRouter extends FileRouter>(
       fileUrl: genUrl,
       url: genUrl,
     };
+
+    console.log("fileData", fileData);
```
markflorkowski, line 264: remove extra log

packages/uploadthing/src/client.ts
```diff
@@ -259,12 +260,68 @@ export const DANGEROUS__uploadFiles = async <TRouter extends FileRouter>(
...
+
+    if (!fileData) {
+      console.log("[UT] Could not receive file data from UT");
```
markflorkowski, line 267: nit `did not` instead of `could not`

packages/uploadthing/src/client.ts
```diff
@@ -259,12 +260,68 @@ export const DANGEROUS__uploadFiles = async <TRouter extends FileRouter>(
...
+    }),
+    headers: {
+      "uploadthing-hook": "callback",
+    },
+  });
+
+  if (isValidResponse(response)) {
+    console.log("[UT] Successfully simulated callback for file", fileKey);
+  } else {
+    console.error(
```
markflorkowski, line 308: Should this be throwing instead?

packages/shared/src/utils.ts
Mr0Bread: I wanted this function to simply return final result so that I do not need to pass callback to do so

packages/uploadthing/src/client.ts
Mr0Bread: Hm, throwed exception may affect user code execution. Since this hook simulation happens in dev mode only, I don't think user has to make any guards for it on client side. I mean this code will become useless in production as it will never be called

packages/uploadthing/src/client.ts
```diff
@@ -258,12 +259,66 @@ export const DANGEROUS__uploadFiles = async <TRouter extends FileRouter>(
...
+    }
+
     return ret;
   });
 
   return Promise.all(fileUploadPromises);
 };
 
+const conditionalDevServer = async (file: FileData, fileKey: string) => {
+  if (process.env.NODE_ENV !== "development") return;
```
juliusmarminge, line 278: this won't work on Vite apps. We should be careful using the `process` object on client side things

juliusmarminge (changes requested):

packages/uploadthing/src/client.ts
Mr0Bread: For real. Well, I guess I can reuse the approach you have for Astro, like the one with try-catch

Mr0Bread:
@juliusmarminge Added support for import.meta

juliusmarminge:
I'm not sure if this actually solves the issue. afaict following both the infra and oss code flow, the pollForFileData will resolve as true when the file is in our db - which is **before** the server side callback may have finished.

juliusmarminge:
Yup, just confirmed.

The server callback started at 20:06:37, client callback ran 20:06:40 but the server hadn't finished until 20:06:52:


CleanShot 2023-09-13 at 22 07 28@2x


CleanShot 2023-09-13 at 22 07 38@2x

Mr0Bread:
Okay, fair enough

Mr0Bread:
Closed in favor of #351

PR #314, opened 2023-08-31
chore: lint packages only

Mr0Bread:
I guess it makes sense to lint packages only by default. The same way as it is done now for `dev` and `build` commands


juliusmarminge:
This doesn't make as much sense to me as 
- linting is often something i just do before commiting, not continuosly during dev
- the linting is fully parallelised so it beraly takes more time to lint everything than just the packages

PR #315, opened 2023-09-01
feat: fork `react-dropzone`

juliusmarminge:
This is a straight-up fork of react-dropzone (incl. rewrite in TypeScript and dropped usage of PropTypes & defaultProps) with better tooling and ESM support. The original package is unmaintained for almost a year now, and doesn't seem to be interested in merging existing PRs

There's some `@ts-expect-error`'s in there as I got tired of fixing everything 😅

- [ ] Port tests?
- [ ] `<Dropzone />`?
- [ ] Improve package code?


Mr0Bread:
That's how packages become 'zero dependencies' type ;DDD

t3dotgg (approved):
lgtm

markflorkowski (approved):
Seems reasonable to me 👍

In the longer term, should we potentially consider integrating the dropzone code directly into our component rather than having a separate package? 

cc @t3dotgg

juliusmarminge:
> In the longer term, should we potentially consider integrating the dropzone code directly into our component rather than having a separate package?

I didn't do this initially cause I thought everyone didn't need that code, but then i realized if you're generating components, you will get the dropzone code imported whether you like it or not... So yea, this could just as well be in the main package if we don't care about external consumers might wanna use it without uploadthing?

markflorkowski:
> So yea, this could just as well be in the main package if we don't care about external consumers might wanna use it without uploadthing?

Yeah, I think it would probably better this way, as we could cut out anything we don't actually need and not worry about maintaining *another* package for external consumption.

docs/src/pages/faq.mdx
```diff
@@ -27,8 +27,3 @@ their `middleware.ts` is configured incorrectly.
 Make sure that `/api/uploadthing` is NOT blocked through your middleware. It
 serves as both a validation endpoint from client AND a webhook for the
 `onUploadComplete` calls.
-
-### I'm getting the error `Named export 'useDropzone' not found`
```
markflorkowski, line 31: This is a *really* nice benefit.

markflorkowski (approved):
Excited to no longer have this as a dep. We should probably add a comment at the top of the `use-dropzone/index.tsx` making note that it is a fork and where it came from.

markflorkowski (approved):

PR #322, opened 2023-09-05
fix(sdk): prevent `uploadFilesInternal` from consuming response body twice on bad response

p6l-richard:
on a bad upload response (within `uploadFilesInternal`), the function calls `UploadThingError.fromResponse()` after having consumed the response body -- leading to a `TypeError` for attempting to consume the response body twice.

### DIFF
--- a/packages/uploadthing/src/sdk/utils.ts
+++ b/packages/uploadthing/src/sdk/utils.ts
@@ -54,7 +54,7 @@ export const uploadFilesInternal = async (
   if (!res.ok) {
     throw UploadThingError.fromResponse(res);
   }
-  
+
   const clonedRes = res.clone(); // so that `UploadThingError.fromResponse()` can consume the body again
   const json = (await res.json()) as
     | {
@@ -66,7 +66,7 @@ export const uploadFilesInternal = async (
         }[];
       }
     | { error: string };
-  
+
   if ("error" in json) {
     throw UploadThingError.fromResponse(clonedRes);
   }
--- a/packages/uploadthing/src/sdk/utils.ts
+++ b/packages/uploadthing/src/sdk/utils.ts
@@ -51,6 +51,10 @@ export const uploadFilesInternal = async (
     }),
   });
 
+  if (!res.ok) {
+    throw UploadThingError.fromResponse(res);
+  }
+  
   const json = (await res.json()) as
     | {
         data: {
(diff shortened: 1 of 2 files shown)


packages/uploadthing/src/sdk/utils.ts
```diff
@@ -61,9 +65,9 @@ export const uploadFilesInternal = async (
         }[];
       }
     | { error: string };
-
-  if (!res.ok || "error" in json) {
-    throw UploadThingError.fromResponse(res);
+  
+  if ("error" in json) {
+    throw UploadThingError.toObject(json);
```
markflorkowski, line 70: I don't think you want `toObject()` here. We'd like to throw an `UploadthingError`

markflorkowski (changes requested):

packages/uploadthing/src/sdk/utils.ts
```diff
@@ -61,9 +66,9 @@ export const uploadFilesInternal = async (
         }[];
       }
     | { error: string };
-
-  if (!res.ok || "error" in json) {
-    throw UploadThingError.fromResponse(res);
+  
+  if ("error" in json) {
+    throw UploadThingError.fromResponse(clonedRes);
```
p6l-richard, line 71: @markflorkowski combined with LOC 58 this should work now, although it might not be the prettiest way to handle this.

markflorkowski: Cloning the response is fine with me. Especially with the comment for why it is done.

markflorkowski (approved):

packages/uploadthing/src/sdk/utils.ts
p6l-richard: alternatively, could call .json() only `if (instanceof responseOrBody === Response)` in from Response, or add a similar method that skips the first line.

```ts
// error.ts
  public static async fromJsonBody(json: Json) {
    let message: string | undefined = undefined;
    if (json !== null && typeof json === "object" && !Array.isArray(json)) {
      if (typeof json.message === "string") {
        message = json.message;
      } else if (typeof json.error === "string") {
        message = json.error;
      }
    }
    return new UploadThingError({
      message,
      code: getErrorTypeFromStatusCode(response.status),
      cause: response,
      data: json,
    });
  }
```

markflorkowski: I don't think such a method provides much value, I think it's fine to just use the normal constructor if you've already parsed the body.

markflorkowski (approved):

t3dotgg (commented):
This bit me earlier today, ty for the fix!

PR #330, opened 2023-09-07
chore: more consistency in logging

markflorkowski:
This is not working right now, but is the start of what I am thinking. This 
will give us more consistent logs, as well as a way to perform ~*other things*\~
should we want to do that 👀

### DIFF
--- a/packages/shared/src/logger.ts
+++ b/packages/shared/src/logger.ts
@@ -1,0 +1,99 @@
+import { DANGEROUS__checkEnvironmentVariable } from "./utils";
+
+const logLevels = {
+  debug: 0,
+  info: 1,
+  silent: 999,
+};
+
+type LogLevel = keyof typeof logLevels;
+
+const getLogLevel = () => {
+  const level = DANGEROUS__checkEnvironmentVariable("UT_LOG_LEVEL") as LogLevel;
+
+  // TODO: remove this log
+  console.log(`[UT] Log level: ${level}`);
+
+  return logLevels[level];
+};
+
+export const log = (message: string, tag?: string, buffer?: boolean) => {
+  if (getLogLevel() <= logLevels.info) {
+    const logTemplate = tag ? `(${tag}): ${message}` : message;
+    const logStyle = "";
+
+    if (buffer) {
+      logBuffer.push({ message, tag, level: "info" });
+      return;
+    }
+
+    console.log(`[UT] ${logTemplate}`, logStyle);
+  }
+};
+
+export const debug = (message: string, tag?: string, buffer?: boolean) => {
+  if (getLogLevel() <= logLevels.debug) {
+    const logTemplate = tag ? `(${tag}): ${message}` : message;
+    const logStyle = "color: #2196f3";
+
+    if (buffer) {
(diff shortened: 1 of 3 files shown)


packages/shared/src/utils.ts
```diff
@@ -203,3 +203,63 @@ export const fileSizeToBytes = (input: string) => {
   const bytes = sizeValue * Math.pow(1024, FILESIZE_UNITS.indexOf(sizeUnit));
   return Math.floor(bytes);
 };
+
+// This is absolutely horrendous, but we need some way to resole environment
+// variables regardless of build tooling. If you have a better idea, please
+// fix this.
+export const DANGEROUS__checkEnvironmentVariable = (variable: string) => {
```
juliusmarminge, line 210: this doesn't work as the variable names has to be statically analyzable

packages/shared/src/logger.ts
```diff
@@ -0,0 +1,99 @@
...
+const logLevels = {
+  debug: 0,
+  info: 1,
+  silent: 999,
+};
+
+type LogLevel = keyof typeof logLevels;
+
+const getLogLevel = () => {
+  const level = DANGEROUS__checkEnvironmentVariable("UT_LOG_LEVEL") as LogLevel;
```
juliusmarminge, line 12: instead of relying on env var here, might be better to accept this as an option when creating an uploadthing client / server

(in conjunction with the new url option in #310)

```ts
const { useUploadThing } = generateReactHelpers({
  url: "...",
  logLevel: "debug" // here **users** could also choose to use env vars if they want as they know what runtime they're using
});

const { UploadButton } = generateComponents({
  url: "...",
  logLevel: "debug" // here **users** could also choose to use env vars if they want as they know what runtime they're using
});
```

(same on the server)

```ts
export const { GET, POST } = createNextRouteHandler({
  router: uploadRouter,
  config: {
    logLevel: "debug",
  },
});
```

packages/shared/src/utils.ts
Mr0Bread: image
It seems that we already have an idea of a solution for that

Mr0Bread: Screenshot from https://github.com/pingdotgg/uploadthing/pull/292

markflorkowski: > this doesn't work as the variable names has to be statically analyzable

Yup 😅 planning to switch to the alternative buildEnv() method you proposed in #292, though it will need to be a bit more complicated to handle the various potential prefixes.

packages/shared/src/logger.ts
markflorkowski: Ideally I'd like the user to only have to set it in one place. But you're probably right, the environment variable story is really not good.

juliusmarminge: Yup, envs works on the server since process is defined there but client is rough 😅

markflorkowski: I am realizing now though that the log wrapper in the shared package wouldn't have access to that. I'll need to make this as a class that gets instantiated and you pass the config into :/

PR #335, opened 2023-09-08
feat: next ssr plugin

juliusmarminge:


packages/react/package.json
```diff
@@ -43,10 +48,16 @@
...
+    "next": "*",
     "react": "^17.0.2 || ^18.0.0",
     "react-dropzone": "^14.2.3",
     "uploadthing": "^5.0.0"
   },
+  "peerDependenciesMeta": {
+    "next": {
+      "optional": true
+    }
+  },
```
juliusmarminge, line 60: needed so that tsup treats next as an external, but we don't want to require next.js

t3dotgg: Will this make all React users require Next though? 😅

t3dotgg (approved):

docs/src/pages/nextjs/appdir.mdx
```diff
@@ -119,4 +119,54 @@ export default function Home() {
 }
 ```
 
+### Configure automatic hydration during SSR (recommended)
+
+> Added in `v5.6`
+
+By default, `useUploadThing` (and thus also the premade components) will fetch
+your server for the route endpoint permissions which will show a loading state.
+By utilizing server components, we can perform automatic hydration of this data
```
markflorkowski, line 128:
```suggestion
By utilizing server components, we can pre-hydrate this data on the server
```

docs/src/pages/nextjs/appdir.mdx
```diff
@@ -119,4 +119,54 @@ export default function Home() {
...
+instead to avoid the loading state and show the route permissions immediately.
+
+To add automatic SSR hydration, simply render the `<NextSSRPlugin />` hydration
```
markflorkowski, line 131:
```suggestion
To add SSR hydration for all uploadthing components, simply render the `<NextSSRPlugin />` hydration
```

markflorkowski: i feel like it is a good thing to call out that this is an "all or nothing" feature that affects every uploadthing component (and doesn't need to be configured except at this global level)

docs/src/pages/nextjs/appdir.mdx
```diff
@@ -119,4 +119,54 @@ export default function Home() {
...
+    <html lang="en">
+      <body>
+        <NextSSRPlugin
+          /**
+           * The `extractRouterConfig` will extract **only** the route configs
+           * from the router to prevent additional information from being
+           * leaked to the client. The data passed to the client is the same
+           * as if you were to fetch `/api/uploadthing` directly.
+           */
+          routerConfig={extractRouterConfig(uploadRouter)}
```
markflorkowski, line 163: I wish we could just pass the whole router and the component could do this extraction for the user.

markflorkowski (approved):

docs/src/pages/nextjs/appdir.mdx
juliusmarminge: The component is a client component though so i think that will leak stuff we dont want to the client

markflorkowski: Yeah, hence why it remains a wish unfortunately :(

juliusmarminge: I guess we could make a wrapper component but i think it's hard getting that to bundle split correctly:

```tsx
// a.ts
function SSRPlugin(props: { router: FileRouter }) {
  return <SSRPluginInner routerConfig={extractRouterConfig} />
}

// b.ts
"use client"
function SSRPluginInner(props: { routerConfig: ... }) {
  useServerInsertedHTML(...)
  
  return null
}
```

markflorkowski: Yeah it is not worth the trouble IMO, too likely to cause problems for not enough gain

PR #339, opened 2023-09-11
chore: use bun package manager and test runner

juliusmarminge:
zoom zoom fast

If we don't yet wanna use the paclage manager, we can revert to https://github.com/pingdotgg/uploadthing/pull/339/commits/6848b70e21f086f46f406934a81eb8a4e55a6630 and just use the test runner

**NOTE** for this to be working you need to prune all the node_modules in the repo and reinstall from scratch

try something like to prune everything

```bash
find . \( -type d -name 'node_modules' -o -name '.turbo' -o -name 'dist' -o -name '.next' \) -exec rm -rf {} +
```


juliusmarminge:
I think the drizzle-kit not finding the sqlite driver will be resolved in their next update when they'll bundle the drivers directly.. not sure why it can't be resolved now

PR #343, opened 2023-09-12
chore: example with drizzle

Mr0Bread:


examples/with-drizzle/src/app/page.tsx
```diff
@@ -0,0 +1,66 @@
...
+      <UploadButton
+        /**
+         * @see https://docs.uploadthing.com/api-reference/react#uploadbutton
+         */
+        endpoint="videoAndImage"
+        onClientUploadComplete={(res) => {
+          console.log(`onClientUploadComplete`, res);
+          alert("Upload Completed");
+
+          setTimeout(() => fetchFiles(), 300);
```
juliusmarminge, line 27: this pattern looks a bit weird, might wanna polish it a bit. should probably also fetch the files in a server component?

Mr0Bread: The goal was to receive up-to-date list of files. I added small timeout to fetching in case there onClientUploadComplete is called too soon. The fix for dev hook was not merged yet, so the order of actions is unreliable

I was thinking about RSC for a bit, but decided to go with the simplest way. This example is about integration between DB and UT uploads

Mr0Bread: So I don't think we need a fancy ways of retrieving data on client

markflorkowski: just using a timeout is inconsistent at best, we should probably show an example of polling, either with our logic, or by leveraging something like SWR

Mr0Bread: In this case, SWR should suffice. It has `refreshInterval` which can be set to like 250ms. Link https://swr.vercel.app/docs/revalidation#revalidate-on-interval

examples/with-drizzle/src/app/uploader.tsx
```diff
@@ -0,0 +1,41 @@
...
+    <>
+      <UploadButton
+        /**
+         * @see https://docs.uploadthing.com/api-reference/react#uploadbutton
+         */
+        endpoint="videoAndImage"
+        onClientUploadComplete={(res) => {
+          console.log(`onClientUploadComplete`, res);
+          alert("Upload Completed");
+          setTimeout(router.refresh, 500);
```
juliusmarminge, line 19: this still isn't the best with a timeout :/

examples/with-drizzle/src/app/page.tsx
markflorkowski: Yup!

examples/with-drizzle-appdir/src/app/page.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+export default async function Home() {
+  const files = await unstable_cache(
+    () => {
+      return db.select().from(schema.files);
+    },
+    [],
+    // We revalidate this tag when an upload completes
+    { tags: ["files"] },
+  )();
```
juliusmarminge, line 15: this is probably how i'd do it - but not sure what the thought is on using unstable things in examples

examples/with-drizzle/src/app/page.tsx
juliusmarminge: i think for the appdir example we should do the fetching in server components (as it is now) - you can use swr in the pages/ example if the plan is to add that too

examples/with-drizzle-appdir/src/app/page.tsx
markflorkowski: Would prefer not to suggest unstable things in our examples. Though I guess drizzle might fall into that category on it's own 😂

examples/with-drizzle/src/app/page.tsx
markflorkowski: Now that I thought a bit more about it, I agree we should try to encourage server components in any appdir examples, since they will provide the best end user experience.

examples/with-drizzle-appdir/src/app/page.tsx
juliusmarminge: Ok swapped it out for a "normal react cache" - although that's probably just as experimental as the next.js one lol

markflorkowski: 😂

juliusmarminge (approved):

examples/with-drizzle/src/app/uploader.tsx
juliusmarminge: added simple polling

Mr0Bread:
@juliusmarminge Polling is good, but it could be better if we have finished with #313 so that we ensure that onClientUploadComplete is always called last in whole chain of actions

juliusmarminge:
> @juliusmarminge Polling is good, but it could be better if we have finished with #313 so that we ensure that onClientUploadComplete is always called last in whole chain of actions

We an always adjust it later. For now, polling is the option we have

PR #345, opened 2023-09-13
fix(sdk): await error before throwing

juliusmarminge:


packages/uploadthing/package.json
```diff
@@ -85,7 +86,8 @@
   "eslintConfig": {
     "root": true,
     "extends": [
-      "@uploadthing/eslint-config/base"
+      "@uploadthing/eslint-config/base",
+      "plugin:@uploadthing/recommended"
```
juliusmarminge, line 90: it didn't find the plugin if i put it in the @uploadthing/eslint-config sadly :/

markflorkowski: Huh, weird. 

Maybe: 
https://github.com/microsoft/rushstack/tree/main/eslint/eslint-patch

markflorkowski (approved):

packages/uploadthing/package.json
juliusmarminge: seems to do the trick!

PR #346, opened 2023-09-13
fix: ignore glob pattern for examples

juliusmarminge:
Before: 
CleanShot 2023-09-13 at 10 42 13@2x

After:
CleanShot 2023-09-13 at 10 42 36@2x

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -12,6 +12,7 @@
     "onlyUpdatePeerDependentsWhenOutOfRange": true
   },
   "ignore": [
-    "docs"
+    "docs",
+    "@example/*"
   ]
 }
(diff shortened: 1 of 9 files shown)


markflorkowski (approved):
Thank you for this 😅

OOC, is there a way to update the deps in the examples when new versions are released? Ideally at some point we should probably do some testing to make sure that the examples all work as expected as new package versions are released

juliusmarminge:
> OOC, is there a way to update the deps in the examples when new versions are released?

Pretty sure that should already happen? If not I'll look into it

markflorkowski:
> Pretty sure that should already happen? If not I'll look into it

I didn't actually check tbh, if it does, that's great :)

PR #351, opened 2023-09-14
feat!: support returning data from server callback

juliusmarminge:
- ensures serverside callback finishes before clientside one is invoked. this is done by removing the s3 polling from the client, and instead fetch the user's server. this fetch request will be resolved when the server callback has finished executing by using an EventEmitter serverside
- adds ability to return json data from the serverside callback which will be passed to the clientside one

I've had to make some breaking type changes for the clientside function `DANGEROUS_uploadFiles` which has propagated to some other public facing types => this requires a major bump.

I think it might be possible to remove the breaking changes if we do some internal casts instead of relying on full inference, not sure what's worth it (depending on if we plan on doing a major soon anyways - depending on what we decide for #310)


packages/uploadthing/src/next-legacy.ts
```diff
@@ -1,3 +1,6 @@
+// This node import should be fine since it's available in both node and edge runtimes
+// https://vercel.com/docs/functions/edge-functions/edge-runtime#compatible-node.js-modules
+import EventEmitter from "events";
```
juliusmarminge, line 3: should(?) also work with Bun: https://nodejs.org/api/events.html

Mr0Bread:
So now instead of many repetitive requests, only one is done that just hangs while waiting for server callback completion? Any chance of request timing out?

juliusmarminge:
> So now instead of many repetitive requests, only one is done that just hangs while waiting for server callback completion? Any chance of request timing out?

I did consider that and use [an EventSource](https://developer.mozilla.org/en-US/docs/Web/API/EventSource) instead (streaming shouldn't ever timeout as long as you start streaming quickly) but I started out like this since it was the easiest.

packages/uploadthing/src/client.ts
```diff
@@ -244,11 +254,12 @@ export const DANGEROUS__uploadFiles = async <TRouter extends FileRouter>(
     // Generate a URL for the uploaded image since AWS won't give me one
     const genUrl = "https://utfs.io/f/" + encodeURIComponent(fields.key);
 
-    // Poll for file data, this way we know that the client-side onUploadComplete callback will be called after the server-side version
-    await pollForFileData(presigned.key);
+    const serverdata = await fetch("/api/uploadthing", {
```
markflorkowski, line 257: nit
```suggestion
    const serverData = await fetch("/api/uploadthing", {
```

packages/uploadthing/src/client.ts
```diff
@@ -98,10 +103,17 @@ export type UploadFileResponse = {
    */
   fileUrl: string;
   url: string;
+
+  // Matches what's returned from the serverside `onUploadComplete` callback
+  serverdata: TServerOutput;
```
markflorkowski, line 108: nit

```suggestion
  serverData: TServerOutput;
```

packages/uploadthing/src/internal/handler.ts
```diff
@@ -224,10 +226,12 @@ export const buildRequestHandler = <
         input?: Json;
       };
 
-      await uploadable.resolver({
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+      const res = await uploadable.resolver({
         file: reqBody.file,
         metadata: reqBody.metadata,
       });
+      ee?.emit("callbackDone", res);
```
markflorkowski, line 234: Should this be prefixed to reduce chances of conflict with user code? 

Seems like you instantiate the EventEmitter though, so I am guessing this is not an issue? I haven't used EventEmitter before, so I am not 100% on the details of how it works.

markflorkowski (approved):

packages/react/test/client-generator.test.ts
```diff
@@ -1,17 +1,14 @@
...
 import { generateReactHelpers } from "../src/hooks";
 
-function ignoreErrors(fn: () => void) {
-  try {
-    fn();
-  } catch {
-    // no-op
-  }
-}
+const doNotExecute = (_fn: (...args: any[]) => any) => {
```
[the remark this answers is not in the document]
juliusmarminge: I just got tired of the rules of hook warnings since we called the hook outside a component, and since we don't test any runtime code, it made sense to turn the ignoreErrors to a doNotExecute. Ofc we can go the extra mile and use something like `renderHook` from `@testing-library/react` but i dont think we need that at this point.

packages/uploadthing/src/internal/handler.ts
juliusmarminge: Me neither actually, but the eventlistener is just on this EventEmitter instance, so I'd assume it shouldnt conflict??

markflorkowski: Yeah that's what I was thinking.

packages/uploadthing/src/client.ts
```diff
@@ -257,17 +268,20 @@ export const DANGEROUS__uploadFiles = async <TRouter extends FileRouter>(
       key: presigned.key,
       fileUrl: genUrl,
       url: genUrl,
+
+      serverdata,
```
juliusmarminge, line 272:
```suggestion
      serverData,
```

packages/uploadthing/src/client.ts
```diff
@@ -253,11 +261,12 @@ export const DANGEROUS__uploadFiles = async <TRouter extends FileRouter>(
     // Generate a URL for the uploaded image since AWS won't give me one
     const genUrl = "https://utfs.io/f/" + encodeURIComponent(fields.key);
 
-    // Poll for file data, this way we know that the client-side onUploadComplete callback will be called after the server-side version
-    await pollForFileData(presigned.key);
+    const serverData = await fetch("/api/uploadthing", {
+      headers: { "x-uploadthing-polling-key": fields.key },
+    }).then((res) => res.json());
```
juliusmarminge, line 266: this should maybe be done in a different way, streaming / polling maybe to prevent request timeout?

t3dotgg: This almost feels like it should run in it's own background process if that makes sense? These behaviors are getting complex enough to need a proper state machine and some useSyncExternalStore bs

This is probably fine for now

packages/uploadthing/src/constants.ts
```diff
@@ -3,4 +3,4 @@ import type { PackageJson } from "type-fest";
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const packageJson = require("../package.json") as PackageJson;
 if (!packageJson.version) throw new Error("no version found in package.json");
-export const UPLOADTHING_VERSION = packageJson.version;
+export const UPLOADTHING_VERSION = "6"; //packageJson.version;
```
markflorkowski, line 6:
```suggestion
export const UPLOADTHING_VERSION = packageJson.version;
```

markflorkowski (approved):

markflorkowski (approved):

PR #355, opened 2023-09-15
chore: pr labeller

juliusmarminge:
should make things a bit more organised


.github/labeler.yml
juliusmarminge, line 1: not sure what labels we should have - added some arbitrary ones i felt would be nice to have. didn't add a solid one since that's getting moved away from here soon

markflorkowski: This seems like a pretty good list to start with. Maybe split docs and examples labels?

PR #358, opened 2023-09-16
feat: h3 adapter

juliusmarminge:
Generic adapter that can be used for Nitro (new SolidStart, Nuxt etc) and other servers using H3

Bored of setting up react vite frontend. Example works with any of the vite clients from the express or fastify examples. We are probably good with just a single one that connects to either of the backend adapter examples


packages/uploadthing/src/h3.ts
```diff
@@ -0,0 +1,67 @@
...
+) => {
+  const requestHandler = buildRequestHandler(opts);
+  const POST = defineEventHandler(async (event) => {
+    const response = await requestHandler({
+      req: {
+        url: getRequestURL(event).href,
+        headers: getRequestHeaders(event),
+        json: () => Promise.resolve(readBody(event)),
+      },
+      //   res: event.node.res, // <-- DO WE NEED THIS?
```
juliusmarminge, line 33: do we need to forward this for the middleware function to be able to run auth functions?

```ts
.middleware(({ req, res }) => {
  const session = await auth(req, res); // next-auth ish
})
```

packages/uploadthing/src/h3.ts
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    if (response.status !== 200) {
+      // We messed up - this should never happen
+      setResponseStatus(event, 500);
+      return "An unknown error occurred";
+    }
+
+    setResponseStatus(event, 200);
+    return JSON.stringify(response);
```
danielroe, line 53:
```suggestion
    return response;
```

packages/uploadthing/src/h3.ts
```diff
@@ -0,0 +1,67 @@
...
+    });
+
+    setHeaders(event, { "x-uploadthing-version": UPLOADTHING_VERSION });
+
+    if (response instanceof UploadThingError) {
+      setResponseStatus(event, getStatusCodeFromError(response));
+      const errorFormatter =
+        opts.router[Object.keys(opts.router)[0]]?._def.errorFormatter ??
+        defaultErrorFormatter;
+      return JSON.stringify(errorFormatter(response));
```
danielroe, line 43:
```suggestion
      return errorFormatter(response);
```

packages/uploadthing/src/h3.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    setResponseStatus(event, 200);
+    return JSON.stringify(response);
+  });
+
+  const getBuildPerms = buildPermissionsInfoHandler<TRouter>(opts);
+  const GET = defineEventHandler((event) => {
+    setResponseStatus(event, 200);
+    setHeaders(event, { "x-uploadthing-version": UPLOADTHING_VERSION });
+    return JSON.stringify(getBuildPerms());
```
danielroe, line 60:
```suggestion
    return getBuildPerms();
```

packages/uploadthing/src/h3.ts
juliusmarminge: Thanks :)

packages/uploadthing/src/h3.ts
juliusmarminge: we'll pass the entire event instead

packages/uploadthing/src/h3.ts
```diff
@@ -0,0 +1,81 @@
...
+  const GET = defineEventHandler((event) => {
+    setResponseStatus(event, 200);
+    setHeaders(event, { "x-uploadthing-version": UPLOADTHING_VERSION });
+    return getBuildPerms();
+  });
+
+  const router = createRouter()
+    .post("/api/uploadthing", POST)
+    .get("/api/uploadthing", GET);
+  return router.handler;
```
juliusmarminge, line 80: @danielroe what would you return here? we later want to support arbitrary pathnames i.e. not locking in to /api/uploadthing, does that mean I should set the path in here to just / and then let the user put in the /api/uploadthing?

```suggestion
  const router = createRouter()
    .post("/", POST)
    .get("/", GET);
  return router.handler;
```

```ts
// user code
const app = createApp();
app.use(
  "/api/uploadthing", 
  createH3EventHandler({ router: uploadRouter })
);
```

or what would you recommend here that will be adaptable to vanilla h3, nitro, nuxt, solidstart etc etc?

danielroe: Yes, that makes sense to me. We'd then have a Nuxt module, for example, that could have a user option for the path, but default to `/api/uploadthing`. But that way it's also fully configurable if using just Nitro, or another framework.

docs/src/pages/backend-adapters/h3.mdx
```diff
@@ -0,0 +1,122 @@
+import { Callout, Steps, Tabs } from "nextra-theme-docs";
+
+# Getting started with H3
+
+> Added in `v5.7`
```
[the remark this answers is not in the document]
juliusmarminge: 5.7 is the upcoming one, no?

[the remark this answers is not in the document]
markflorkowski: You're correct. I am apparently unable to read. 🤦

examples/backend-adapters/README.md
```diff
@@ -5,13 +5,14 @@
 </a>
 
 This example contains a vite-react app in the [/client](./client/) directory, as
-well as multiple backend servers in the [/server](./server/) directory. Servers
-that are included here are:
+well as multiple backend servers in the [/server](./server/) and
+[/nitro](./nitro/) directories. Servers that are included here are:
```
markflorkowski, line 9: Would be kind of nice to adjust the nesting here so that there are just `client` and `server` folders and inside the `server` folder there would be `nitro` and `others`

markflorkowski (approved):

examples/backend-adapters/README.md
juliusmarminge: Problem is nitro requires that routes/api folder and (I think) a nitro config. I could just replace nitro with a base node.js h3 app though

danielroe: The nitro config isn't required. But a base h3 app (with listhen server?) could be fine also.

juliusmarminge: > The nitro config isn't required. But a base h3 app (with listhen server?) could be fine also.

Cool, I'll do a listhen server then, we still have the nitro "setup" in the docs

juliusmarminge: Hmm @danielroe now the issue I was telling you about yesterday is back... It's matching the /api route to /api/uploadthing instead of the actual uploadthing routes...

CleanShot 2023-09-29 at 09 39 36@2x

I pushed what I got if you can spot any mistakes...

docs/src/pages/backend-adapters/h3.mdx
```diff
@@ -0,0 +1,134 @@
...
+  <Tabs.Tab>
+    ```ts filename="routes/api/uploadthing.ts"
+    import { createH3EventHandler } from "uploadthing/h3";
+
+    import { uploadRouter } from "./router";
+
+    export default useBase(
+      "/api/uploadthing",
+      createH3EventHandler({ router: uploadRouter }),
+    );
```
danielroe, line 102:
```suggestion
    export default createH3EventHandler({ router: uploadRouter });
```

docs/src/pages/backend-adapters/h3.mdx
```diff
@@ -0,0 +1,134 @@
...
+  <Tabs.Tab>
+    ```ts filename="server/api/uploadthing.ts"
+    import { createH3EventHandler } from "uploadthing/h3";
+
+    import { uploadRouter } from "./router";
+
+    export default useBase(
+      "/api/uploadthing",
+      createH3EventHandler({ router: uploadRouter }),
+    );
```
danielroe, line 115:
```suggestion
    export default createH3EventHandler({ router: uploadRouter });
```

docs/src/pages/backend-adapters/h3.mdx
```diff
@@ -0,0 +1,128 @@
...
+
+### Creating your first FileRoute
+
+All files uploaded to uploadthing are associated with a FileRoute. Following
+example is very minimalistic. To get full insight into what you can do with the
+FileRoutes, please refer to the
+[File Router API](/api-reference/server#file-routes).
+
+```ts copy filename="src/uploadthing.ts"
+import { createUploadthing, type FileRouter } from "uploadthing/fastify";
```
danielroe, line 48:
```suggestion
import { createUploadthing, type FileRouter } from "uploadthing/h3";
```

packages/uploadthing/src/h3.ts
```diff
@@ -0,0 +1,79 @@
...
+  const requestHandler = buildRequestHandler(opts);
+  const getBuildPerms = buildPermissionsInfoHandler<TRouter>(opts);
+
+  return defineEventHandler(async (event) => {
+    assertMethod(event, ["GET", "POST"]);
+    setHeaders(event, { "x-uploadthing-version": UPLOADTHING_VERSION });
+
+    // GET
+    if (event.method === "GET") {
+      setResponseStatus(event, 200);
```
danielroe, line 48: this is default so I think we can avoid....

```suggestion
```

packages/uploadthing/src/h3.ts
```diff
@@ -0,0 +1,79 @@
...
+      return errorFormatter(response) as unknown;
+    }
+
+    if (response.status !== 200) {
+      // We messed up - this should never happen
+      setResponseStatus(event, 500);
+      return "An unknown error occurred";
+    }
+
+    setResponseStatus(event, 200);
```
danielroe, line 76:
```suggestion
```

packages/uploadthing/src/h3.ts
juliusmarminge: Fair enough - the less code the better right

packages/uploadthing/src/internal/types.ts
juliusmarminge, line 1: Reminder for myself to look over these types, doesn't make sense having to add an any generic in a bunch of places like this...

PR #359, opened 2023-09-16
fix: allow `skipLibCheck: false` by removing internal type imports

juliusmarminge:
Removes the `_runtime` in favor of a `_mdwrArgs`, i.e. instead of passing a string runtime and letting the internal type decide what type `req` and `res` should have, the `createBuilder` has to provide those type itself.

This lets us remove all the lib imports completely in the internal types, meaning only the specific entrypoint will import stuff from that package.

e.g. the /next entrypoint is the only file importing stuff from next, /express is the only that imports from express.

This means that we don't block (well, at least that's the goal) `skipLibCheck: false` for people who wanna use that option since no types are getting pulled in that they don't have. 

**NOTE**: Some type imports from internal files are also removed in #360


packages/uploadthing/src/internal/types.ts
```diff
@@ -48,31 +41,26 @@ type ResolverOptions<TParams extends AnyParams> = {
 };
 
 export type AnyRuntime = "app" | "pages" | "web" | "express" | "fastify";
+
+export type MiddlewareFnArgs<TRequest, TResponse> = {
+  req: TRequest;
+  res: TResponse;
```
juliusmarminge, line 47: will extend this to also include an `event` prop [to support H3](https://github.com/pingdotgg/uploadthing/pull/358). 

That was actually the rationale for this PR in the first place :) https://github.com/pingdotgg/uploadthing/actions/runs/6206560816/job/16850938399?pr=358#step:8:512

markflorkowski (approved):

markflorkowski:
I think this is also just a generally cleaner approach.

juliusmarminge:
> I think this is also just a generally cleaner approach.

yep - totally.

PR #360, opened 2023-09-16
refactor: rm weird export chain

juliusmarminge:
Previously, the /server export re-exported from /internal/edge, which was an export from /next. This is veeeery weird.

I removed the edge one completely, and made the /server entrypoint the one actually doing the work, then /next can re-export from server but with the custom `NextRequest` type

also made a reusable helper `formatError` to reduce some duplicated logic around error formatting


packages/uploadthing/src/internal/handler.ts
```diff
@@ -149,21 +85,12 @@ type UploadThingResponse = {
...
   return async (input: {
     req: RequestLike;
-    res?: TRuntime extends "pages"
-      ? NextApiResponse
-      : TRuntime extends "express"
-      ? ExpressResponse
-      : TRuntime extends "fastify"
-      ? FastifyReply
-      : undefined;
+    res?: unknown;
```
juliusmarminge, line 93: doesn't really make a lot of sense that this is typed, it's just in the way similar to #359

packages/uploadthing/src/next.ts
```diff
@@ -21,52 +17,14 @@ export const createUploadthing = <TErrorShape extends Json>(
 export const createNextRouteHandler = <TRouter extends FileRouter>(
```
juliusmarminge, line 17: this doesn't actually need to exist, we could alias it to `createServerHandler` and potentially deprecate it in favor of

```ts
import { createServerHandler } from "uploadthing/server";
import { createUploadThing } from "uploadthing/next";
```

packages/uploadthing/src/next.ts
```diff
@@ -1,73 +1,13 @@
...
-      status: 200,
-      headers: {
-        "x-uploadthing-version": UPLOADTHING_VERSION,
-      },
-    });
-  };
-
-  return { GET, POST };
-};
+export const createNextRouteHandler = createServerHandler;
```
juliusmarminge, line 13: this doesn't actually need to exist, we could alias it to `createServerHandler` and potentially deprecate it in favor of

```ts
import { createServerHandler } from "uploadthing/server";
import { createUploadThing } from "uploadthing/next";
```

or do you think we should keep it? no reason to special treat next here really...

juliusmarminge: actually we did need this to be it's own thing since Next only accepts a very strict signature on the exports and the request being typed as `Request | { request: Request }` wasn't very welcome.

packages/uploadthing/src/internal/error-formatter.ts
```diff
@@ -1,7 +1,21 @@
...
     message: error.message,
   };
 }
+
+export function formatError<TRouter extends FileRouter>(
+  error: UploadThingError,
+  router: TRouter,
+): inferErrorShape<TRouter> {
+  const errorFormatter =
+    router[Object.keys(router)[0]]?._def.errorFormatter ??
```
markflorkowski, line 16: hmmmm this seems a bit sketchy

markflorkowski (approved):

packages/uploadthing/src/internal/error-formatter.ts
juliusmarminge: Haha right? It's the same code as before though 😅

PR #361, opened 2023-09-16
fix: move log to prevent spam in dev

juliusmarminge:
Ref: https://discord.com/channels/966627436387266600/1102510616326967306/1152679262336339968

CleanShot 2023-09-16 at 21 06 17@2x


markflorkowski (approved):
Surprised nobody called this out earlier. I hadn't even noticed it myself 😅

PR #364, opened 2023-09-17
chore: Doc page with Fetch runtime setup examples

Mr0Bread:
Created new doc page that explains process of setting up UT on server-side that are based on Request/Response

As an example I referred to: https://trpc.io/docs/server/adapters/fetch


Mr0Bread:
Should doc page contain links to examples source code/stackblitz?

docs/src/pages/backend-adapters/fetch.mdx
```diff
@@ -0,0 +1,145 @@
...
+
+app.group("/api/uploadthing", (app) =>
+  app
+    .post("/", (context) => {
+      return POST(context.request);
+    })
+    .get("/", (context) => {
+      return GET(context.request);
+    }),
+);
```
juliusmarminge, line 107: What's the benefit of a group here? Seems to me 

```ts
app.get("/api/uploadthing", ...)
app.post("/api/uploadthing", ...)
```

Does the same thing?

docs/src/pages/backend-adapters/fetch.mdx
```diff
@@ -0,0 +1,145 @@
...
+});
+
+const app = new Hono();
+const ut = new Hono();
+
+ut.get("/", (c) => {
+  return GET(c.req.raw);
+}).post("/", (c) => {
+  return POST(c.req.raw);
+});
```
juliusmarminge, line 132: Same here, just add the routes individually? 

I have never used Hono but seems odd to create a new instance?

juliusmarminge (approved):
We should think on how we want to structure ezamples. Having the same vite app like 10 times hurts CI time a looot, especially from forks that cant use the remote cache.

I wonder if a single vite app "connecting" to localhost:3000 and then all the backend adapters can be like 2 files 🤔 food for thought

Mr0Bread:
> Really it could even just be a single example with a single router, and then just a bunch of different `dev:hono`, `dev:fastify` scripts that starts a different server 🤔

Probably then something like that

- examples
  - minimal-backend-react
    - client
    - fastify
    - express
    - hono
    - elysia

docs/src/pages/backend-adapters/fetch.mdx
Mr0Bread: In case of both Hono and Elysia I just decided to make it through grouping so that the same endpoint "/api/uploadthing" is not repeated twice

In Hono you create new instance of Hono class when you want to create something similar to Express router

Mr0Bread:
I can make such refactorings to examples of backend adapters in scope of this PR

juliusmarminge:
> I can make such refactorings to examples of backend adapters in scope of this PR

I'd say Leave it for a follow up

PR #366, opened 2023-09-17
docs: restructure [WIP]

juliusmarminge:
trying out a bit of a refactor, not completely happy with this either though...

Maybe the `Setting up your router` should just be on it's separate page 🤔


Mr0Bread:
> Maybe the Setting up your router should just be on it's separate page 🤔

Then user, who wants to set up the UT would need to go to separate page just to check the example. Maybe we can have the most minimal example on each page with setup instructions + page with more details(though we have Server API reference where file router is described in more details)

Mr0Bread:
Also, it's insignificant, but I guess you do not need links to Vue and Svelte pages since those are not planned to be part of the official docs

juliusmarminge:
> Also, it's insignificant, but I guess you do not need links to Vue and Svelte pages since those are not planned to be part of the official docs

I can change them to external links to the respective docs/readme file later - i still think we should link to them even if they are community maintained

PR #368, opened 2023-09-18
feat: SSR plugin using server component

Mr0Bread:


juliusmarminge:
What's the goal here? This is the same implementation as #334?

PR #373, opened 2023-09-19
feat(sdk): add usage endpoint to sdk

juliusmarminge:


packages/uploadthing/src/sdk/index.ts
```diff
@@ -278,3 +278,36 @@ export const renameFile = async (updates: Rename | Rename[]) => {
 
   return json;
 };
+
+export const getUsageInfo = async () => {
+  guardServerOnly();
```
markflorkowski, line 283: Does this actually need to be server-only?

markflorkowski (approved):

packages/uploadthing/src/sdk/index.ts
juliusmarminge: we need the api key so yea?

markflorkowski: 🤦 You're right, I am dumb....

PR #382, opened 2023-09-22
fix(`UPL-63`): scream if node < 18

juliusmarminge:
Closes UPL-63


packages/uploadthing/src/internal/incompat-node-guard.ts
```diff
@@ -0,0 +1,27 @@
...
+        .substring(1)
+        .split(".")
+        .map((v) => parseInt(v, 10));
+    }
+  }
+
+  if (major && major < 18) {
+    throw new Error(
+      `[UT]: YOU ARE USING A LEGACY (${major}) NODE VERSION WHICH ISN'T OFFICIALLY SUPPORTED. PLEASE UPGRADE TO NODE 18+.`,
+    );
```
markflorkowski, line 25:
```suggestion
    );
    process.exit(1);  // Kill the process if it isn't going to work correctly anyway
```

markflorkowski (approved):

packages/uploadthing/src/internal/incompat-node-guard.ts
juliusmarminge: Is that cross-runtime compatible?

markflorkowski: Good point. I would assume so, but don't actually know for sure. I'll have to check.

markflorkowski: Should work in bun. Not sure of other runtimes that have `process` defined tbh

juliusmarminge: We can fix that when we get there if it's an issue I guess

PR #383, opened 2023-09-23
feat(react): paste from clipboard

Mr0Bread:


juliusmarminge (commented):
Works nice.

Some questions:

- Should it be configurable?
- Should add some words in docs about the behavior?
- The default text on dropzone is now outdated, should it include that you can paste a file?
- Pasting a file in the dropzone currently resets the selected files, I see that that matches the behavior if you click the dropzone multiple times - but maybe we should append the pasted files to the selection?

Mr0Bread:
@juliusmarminge 

1. In what way? Can you elaborate?
2. Yeah, fair enough
3. Hm, pasting requires having focus on dropzone. We either have to clarify that part or not include text about pasting at all to not make it confusing
4. Hm, probably you are right, I'll ask the rest of the team

juliusmarminge:
1. Being able to disable it

docs/src/pages/api-reference/react.mdx
```diff
@@ -52,6 +52,13 @@ export const OurUploadButton = () => (
 | onUploadError          | function | No                      |                 | callback function when upload fails                                                 |
 | onUploadProgress       | function | No                      | Added in `v5.1` | callback function for upload progress                                               |
 | onUploadBegin          | function | No                      | Added in `v5.4` | callback function for upload begin                                                  |
+| config                 | Config   | No                      | Added in `v5.7` | object to pass additional configuration for button                                  |
+
+Config object
+
+| Attribute    | Type    | Required | Notes           | Description                                    |
+| :----------- | :------ | :------- | :-------------- | :--------------------------------------------- |
+| disablePaste | boolean | No       | Added in `v5.7` | disables ability to paste files from clipboard |
```
juliusmarminge, line 61: imo options should be positive, so this should probably be something along the lines of `appendOnPaste` (naming is hard) with `true` as default (or maybe even `false` now and we can change default to true in next major, depending on if you see this as a breaking change or a new feature)

negated options are often just harder to parse

Mr0Bread: Fixed

docs/src/pages/api-reference/react.mdx
```diff
@@ -56,9 +56,9 @@ export const OurUploadButton = () => (
 
 Config object
 
-| Attribute    | Type    | Required | Notes           | Description                                    |
-| :----------- | :------ | :------- | :-------------- | :--------------------------------------------- |
-| disablePaste | boolean | No       | Added in `v5.7` | disables ability to paste files from clipboard |
+| Attribute     | Type    | Required | Notes           | Description                                   |
+| :------------ | :------ | :------- | :-------------- | :-------------------------------------------- |
+| appendOnPaste | boolean | No       | Added in `v5.7` | enables ability to paste files from clipboard |
```
markflorkowski, line 61: sidenote, if values have a default, we should include that info in this table

packages/react/src/components/button.tsx
```diff
@@ -105,8 +105,8 @@ export function UploadButton<TRouter extends FileRouter>(
 
   useEffect(() => {
     const handlePaste = (event: ClipboardEvent) => {
-      if ($props.config?.disablePaste) {
-        // User decided to disable paste-from-clipboard functionality
+      if (!$props.config?.appendOnPaste) {
```
markflorkowski, line 108: nit: Doesn't really require a comment, can be made into a 1 line early return

docs/src/pages/api-reference/react.mdx
Mr0Bread: It's not like it has default. Like if it is not set to true, it will be undefined. Such value is falsy. The paste handler will return if `appendOnPaste` is falsy. Not exactly 'false'.

Should we still clarify, that this feature is not going to be enabled by default?

juliusmarminge: I think it should stay disabled by default, and then I think docs are fine as is now.

We can enable it by default in the next major maybe?

Mr0Bread: The functionality will not work until appendOnPaste is set to true. So it's disabled by default

juliusmarminge (approved):

PR #384, opened 2023-09-23
fix: progress bar and spinner

Mr0Bread:
Fixes two issues:
1. Classes `after:w-[%]` were not generated because they were part of another package not included in tailwind.config. I decided to move progressWidths instead of changing tailwind.config
2. Spinner had been covered by progress bar


juliusmarminge (approved):
fair enough

PR #388, opened 2023-09-24
chore(sdk): accept constructor args + log response + better errors on utapi

### LINKED ISSUE #387: [bug]: Broken Cloudflare Workers support due to lack of `cache` support in `fetch`

```bash
System:
    OS: macOS 13.5.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 99.33 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 19.8.1 - ~/Library/Caches/fnm_multishells/87451_1695591340601/bin/node
    npm: 9.6.2 - ~/Library/Caches/fnm_multishells/87451_1695591340601/bin/npm
    pnpm: 8.1.0 - ~/.yarn/bin/pnpm
  Browsers:
    Chrome: 117.0.5938.92
    Safari: 16.6
```

### Describe the bug

Attempting to call any function from `utapi` in Cloudflare Workers will result in an exception.

See: https://github.com/cloudflare/workerd/issues/698

I believe this has been introduced in https://github.com/pingdotgg/uploadthing/pull/278

### Link to reproduction

https://stackblitz.com/edit/github-1nhyey?file=pages%2Ffunctions%2Fexample.ts

### To reproduce

```
cd pages
wrangler pages dev .
curl http://127.0.0.1:8788/example
```

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

- [X] 🙋‍♂️ Yes, I'd be down to file a PR fixing this bug!

### Code of Conduct

juliusmarminge:
- Creates a class that can accept some constructor args such as custom fetch implementation. 
  - Closes #387 
  - Closes #389

- Improves errors and logs. 
  - Requires https://github.com/pingdotgg/uploadthing-infra/pull/206


packages/uploadthing/src/sdk/index.ts
```diff
@@ -3,319 +3,288 @@ import { generateUploadThingURL, UploadThingError } from "@uploadthing/shared";
...
+  private async requestUploadThing<T extends Record<string, unknown>>(
+    pathname: `/${string}`,
+    body: Record<string, unknown>,
+    fallbackErrorMessage: string,
+  ) {
+    // Force API key to be set before requesting.
+    // Ideally we'd just throw in the constructor but since we need to export
+    // a `utapi` object we can't throw in the constructor because it would
+    // be a breaking change.
+    // FIXME: In next major
```
juliusmarminge, line 51: This would be a breaking change since previously this check was made when making the request, so they could still import `utapi` without a secret set and not get errors.

Not sure what we want to do to fix this, maybe just exporting the class and they initialize the `utapi` object themselves? That way you can still import from this file (or /server) without immediately throwing if you don't have a secret set in your envs

.changeset/wicked-birds-push.md
```diff
@@ -0,0 +1,5 @@
+---
```
markflorkowski, line 1: More changeset details, also should probably be a minor.

Initially I thought this would be breaking, but I see now that you exported an instance of the class as well, so people can continue to use it as before.

packages/uploadthing/src/sdk/index.ts
markflorkowski: > maybe just exporting the class and they initialize the utapi object themselves

This is how I would probably do it. It will be a breaking change, but any other way would likely still be breaking, and also almost guaranteed to be more complicated.

markflorkowski: > without immediately throwing if you don't have a secret set in your envs

Why do we care if this happens though? You can't use these functions without a secret in your env vars, so why not throw until you do? As long as the error message is clear 🤷

markflorkowski: fair enough 😂

.changeset/wicked-birds-push.md
juliusmarminge: Done https://github.com/pingdotgg/uploadthing/commit/6ece5176f56e4700b8042ed3e75caad94b1b30f3

packages/uploadthing/src/sdk/index.ts
```diff
@@ -3,319 +3,297 @@ import { generateUploadThingURL, UploadThingError } from "@uploadthing/shared";
...
+   * Provide a custom fetch function.
+   * @default globalThis.fetch
+   */
+  fetch?: FetchEsque;
+  /**
+   * Provide a custom UploadThing API key.
+   * @default process.env.UPLOADTHING_SECRET
+   */
+  apiKey?: string;
 }
```
juliusmarminge, line 25: // TODO: custom logger option?

packages/uploadthing/src/sdk/index.ts
```diff
@@ -3,319 +3,297 @@ import { generateUploadThingURL, UploadThingError } from "@uploadthing/shared";
...
-  return process.env.UPLOADTHING_SECRET;
+import type { FetchEsque, MaybeUrl } from "./types";
+import type { FileEsque, UploadFileResponse } from "./utils";
+import {
+  getApiKeyOrThrow,
+  guardServerOnly,
+  uploadFilesInternal,
+} from "./utils";
+
+export interface UTApiOptions {
```
juliusmarminge, line 14: TODO: allow custom logger?

markflorkowski (approved):

PR #390, opened 2023-09-26
fix: ensure that uploadProgress fires 0 when upload starts

GentikSolm:
Pretty straight forward, fires a 0 progress callback the same time we initially set uploading state to true

### DIFF
--- a/packages/react/src/useUploadThing.ts
+++ b/packages/react/src/useUploadThing.ts
@@ -59,6 +59,7 @@ export const INTERNAL_uploadthingHookGen = <TRouter extends FileRouter>() => {
     const startUpload = useEvent(async (...args: FuncInput) => {
       const [files, input] = args;
       setUploading(true);
+      opts?.onUploadProgress?.(0);
       try {
         const res = await DANGEROUS__uploadFiles({
           files,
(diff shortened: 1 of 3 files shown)


packages/react/src/useUploadThing.ts
```diff
@@ -59,6 +59,7 @@ export const INTERNAL_uploadthingHookGen = <TRouter extends FileRouter>() => {
     const startUpload = useEvent(async (...args: FuncInput) => {
       const [files, input] = args;
       setUploading(true);
+      opts?.onUploadProgress?.(0);
```
juliusmarminge, line 62: is there a possibility that we'd now call onUploadProgress twice with 0?

Mr0Bread: If I remember correctly, there is such possibility. So we need a mechanism to prevent it from happening.

Maybe wrap callback in closure that will have mutable variable called 'isCalledWithZero' and if statement checking incoming argument and whether it was called with 0. If argument is 0 and variable is false, we call callback and set variable to true. If argument is 0 and variable is true, do nothing

GentikSolm: Look a few lines down to line 77. When I originally built this I made sure to add some checks for that.

markflorkowski (approved):

juliusmarminge (approved):

Issue #391, opened 2023-09-26
[bug]: React Button component isn't keyboard-focusable


HugeLetters:
### Provide environment information

```bash
...
```

### Describe the bug

Since the input element inside has a [hidden](https://github.com/pingdotgg/uploadthing/blob/a421df3b26236f939ea569c80c6472da07895049/packages/react/src/components/button.tsx#L108) class keyboard navigation for it is disabled. 

My fix suggestion would be to either replace `hidden` with `absolute w-0 h-0 opacity-0` or put `tabindex=0` on the root div like you do for Dropbox component. 

### Link to reproduction

https://stackblitz.com/edit/github-jvitfq?file=src%2Fapp%2Fpage.tsx

### To reproduce

I've added a red background to focused elements in the link above to showcase the issue more clearly - dropbox and native input have the background when using TAB to navigate, button component does not.

### 👨‍👧‍👦 Contributing

- [X] 🙋‍♂️ Yes, I'd be down to file a PR fixing this bug!

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

markflorkowski:
Good callout! Will get a pr up for this soon

markflorkowski:
Will be in the next release!

PR #392, opened 2023-09-27
fix(components): Allow upload button to be focusable via keyboard

markflorkowski:
Addresses #391

To Do:
- Add focus ring

### DIFF
--- a/packages/react/src/components/button.tsx
+++ b/packages/react/src/components/button.tsx
@@ -146,7 +146,7 @@ export function UploadButton<TRouter extends FileRouter>(
     >
       <label
         className={twMerge(
-          "relative flex h-10 w-36 cursor-pointer items-center justify-center overflow-hidden rounded-md text-white after:transition-[width] after:duration-500 focus-within:ring-blue-600 focus-within:ring-offset-2",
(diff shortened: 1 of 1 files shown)


packages/react/src/components/button.tsx
```diff
@@ -146,7 +146,7 @@ export function UploadButton<TRouter extends FileRouter>(
     >
       <label
         className={twMerge(
-          "relative flex h-10 w-36 cursor-pointer items-center justify-center overflow-hidden rounded-md text-white after:transition-[width] after:duration-500",
+          "relative flex h-10 w-36 cursor-pointer items-center justify-center overflow-hidden rounded-md text-white after:transition-[width] after:duration-500 focus-within:ring-blue-600 focus-within:ring-offset-2",
```
juliusmarminge, line 149: I don't see this focus state: 

https://github.com/pingdotgg/uploadthing/assets/51714798/d52f404e-95cc-4919-a328-e1fc42b9d25d

juliusmarminge: just had to add the ring class too

juliusmarminge (approved):

PR #395, opened 2023-09-28
fix: incompat node guard for node > 18

juliusmarminge:
logic is hard...

### DIFF
--- a/packages/uploadthing/src/internal/incompat-node-guard.ts
+++ b/packages/uploadthing/src/internal/incompat-node-guard.ts
@@ -23,10 +23,11 @@ export function incompatibleNodeGuard() {
   if (!major || !minor) return;
 
   // Require ^18.13.0
-  if (!((major === 18 && minor >= 13) || major >= 19)) {
-    console.error(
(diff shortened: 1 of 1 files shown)


packages/uploadthing/src/internal/incompat-node-guard.ts
```diff
@@ -23,7 +23,7 @@ export function incompatibleNodeGuard() {
   if (!major || !minor) return;
 
   // Require ^18.13.0
-  if (!(major >= 18 && minor >= 13)) {
+  if (!((major === 18 && minor >= 13) || major >= 19)) {
```
Mr0Bread, line 26: It's just my personal preference, but wouldn't be more readable if we remove negation and invert the condition? Like

```suggestion
// In case node >= 18.13 we return and allow further execution
if ((major === 18 && minor >= 13) || major >= 19) return;

console.error(/* error goes here */)
process.exit(1)
```

juliusmarminge: good idea!

markflorkowski (approved):

PR #396, opened 2023-09-28
feat(react): Manual mode for button

Mr0Bread:


Mr0Bread:
@juliusmarminge I've raised that concern in our discord channel actually ;D

My idea was to introduce the "Clear" button. We could display it under the button instead of allowed content label

juliusmarminge:
> My idea was to introduce the "Clear" button. We could display it under the button instead of allowed content label

Sounds good to me!

markflorkowski (approved):

juliusmarminge:
Something's broken here. After clicking reset, I can't add new files?

CleanShot 2023-10-05 at 10 33 58

Mr0Bread:
@juliusmarminge Fixed the issue

juliusmarminge (approved):

PR #404, opened 2023-10-02
fix: safeParseJSON helper

juliusmarminge:
Hacky..., could maybe use https://github.com/unjs/destr


packages/shared/src/utils.ts
```diff
@@ -211,3 +217,15 @@ export const fileSizeToBytes = (input: string) => {
   const bytes = sizeValue * Math.pow(1024, FILESIZE_UNITS.indexOf(sizeUnit));
   return Math.floor(bytes);
 };
+
+export async function safeParseJSON<T>(
+  input: string | { text: () => Promise<string> },
+): Promise<T | Error> {
+  const text = typeof input === "object" ? await input.text() : input;
+  try {
+    return JSON.parse(text) as T;
```
juliusmarminge, line 226: response/request.text() then JSON.parse might not be the best here, but first awaiting .json() before falling back to awaiting .text() doesn't feel right either...

markflorkowski: I'll think on this, while it may not be the best, I feel like it's still a pretty reasonable way to handle this, even if we lose a little bit of efficiency.

markflorkowski: I guess the clone method you used in #415 is a little cleaner

packages/shared/src/types.ts
```diff
@@ -7,6 +9,67 @@ export type JsonArray = JsonValue[];
...
+ * A subset of the standard Response properties needed by UploadThing internally.
+ * @see Response from lib.dom.d.ts
+ */
+export interface ResponseEsque {
+  ok: boolean;
+  /**
+   * @remarks
+   * The built-in Response::json() method returns Promise<any>, but
+   * that's not as type-safe as unknown. We use unknown because we're
+   * more type-safe. You do want more type safety, right? 😉
```
markflorkowski, line 46: 😂

markflorkowski (approved):

PR #406, opened 2023-10-03
fix: Add optional to import.meta.env call

daalvarado:
Hi, when using Remix this file fails because import.meta.env does not exist. Proposing a change to prevent function from failing and use the host value in these cases.

Thanks

### DIFF
--- a/packages/shared/src/utils.ts
+++ b/packages/shared/src/utils.ts
@@ -112,7 +112,6 @@ export function generateUploadThingURL(path: `/${string}`) {
   } else {
     // @ts-expect-error - import.meta is dumb
     // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-assignment
-    // prevent error if import.meta.env does not exist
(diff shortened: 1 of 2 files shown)


packages/shared/src/utils.ts
```diff
@@ -112,7 +112,8 @@ export function generateUploadThingURL(path: `/${string}`) {
   } else {
     // @ts-expect-error - import.meta is dumb
     // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-assignment
-    host = import.meta.env.CUSTOM_INFRA_URL ?? host;
+    // prevent error if import.meta.env does not exist
```
markflorkowski, line 115:
```suggestion
```

markflorkowski (approved):
Adding an optional here doesn't seem terrible to me. cc @juliusmarminge @t3dotgg

juliusmarminge (approved):

shmuli9:
@daalvarado have you got an example of integrating this in a remix app?

PR #407, opened 2023-10-03
chore: BigInt support 

markflorkowski:
I don't actually think this function is used anywhere 😅


markflorkowski:
unnecessary as it turns out.

PR #416, opened 2023-10-05
feat: allow overriding content disposition

juliusmarminge:


examples/minimal-appdir/src/server/uploadthing.ts
```diff
@@ -12,6 +12,7 @@ export const uploadRouter = {
     image: {
       maxFileSize: "4MB",
       maxFileCount: 4,
+      contentDisposition: "attachment",
```
markflorkowski, line 15: comment what this is about?

markflorkowski (approved):
Should probably add this to the docs

examples/minimal-appdir/src/server/uploadthing.ts
juliusmarminge: I'll revert the example before merging - will add some words in docs instead

markflorkowski (approved):

PR #421, opened 2023-10-07
chore: react hook form example

Mr0Bread:


examples/with-react-hook-form/README.md
```diff
@@ -0,0 +1,18 @@
+# Minimal Next.js App Router example for UploadThing
```
markflorkowski, line 1: Title should probably mention react-hook-form

examples/with-react-hook-form/src/app/page.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const onSubmit: SubmitHandler<Inputs> = async (data) => {
+    if (!files) return;
+
+    const res = await startUpload(files);
+
+    if (!res) {
+      alert("Upload failed");
+      return;
+    }
+
```
markflorkowski, line 33: I'd maybe leave some comments in here explaining what information we have at different points (eg. file itself, uploaded url, other form data) and maybe a mock of how a user might want to store info in their db.

markflorkowski (commented):
Looks good overall, just left a few comments. 

We may also want an example where files upload immediately, and urls are included in the form data. (Would be a bit more complex since it should probably handle deleting files if the user decides to replace their selection)

Mr0Bread:
@markflorkowski would you like that more advanced use case within this example or in separate one?

markflorkowski:
Hmm, I think it can be in this one, since they are both rhf. Do you think that would be confusing?

Mr0Bread:
@markflorkowski Hm, I've just been thinking. Maybe it's better to simply improve current example? I mean does it make sense to upload the image each time it is dropped? What might be the reason for that? Browser preview does not require it. If form data needs an URL of the image, it can be received during form submission. Simply upload image inside of `onSubmit` and get URL from the response. Let me know, if I'm missing something here

markflorkowski:
The only reason I was thinking was it is slightly better UX in the form to have files upload in the background as you continue to fill out the form, instead of waiting for files to upload when submitting.

In the contrived example of a form that *only* contains a file upload it makes less sense, but I feel like in most cases it will be one field of many on a form.

Regardless, the url should be submitted with the form data, as presumably that data will be stored or used in some way by the application

examples/with-react-hook-form/tailwind.config.ts
```diff
@@ -0,0 +1,17 @@
+import { fontFamily } from "tailwindcss/defaultTheme";
+
+import { withUt } from "uploadthing/tw";
+
+/**
+ * @see https://docs.uploadthing.com/theming#with-tailwindcss
```
juliusmarminge, line 6: https://docs.uploadthing.com/getting-started/appdir#add-uploadthings-styles

examples/with-react-hook-form/src/app/page.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const { startUpload } = useUploadThing("videoAndImage");
+  const { handleSubmit, register, reset } = useForm<Inputs>();
+  const { getInputProps, getRootProps } = useDropzone({
+    onDrop: (files) => {
+      setFiles(files);
+    },
+  });
+
+  const onSubmit: SubmitHandler<Inputs> = async (data) => {
+    if (!files) return;
```
juliusmarminge, line 25: why aren't files set on the form?

examples/with-react-hook-form/README.md
chihabhajji: i crawled all the repo looking for a react-hook-form implementation, thank you for the suggestion

t3dotgg:
Due to the likelihood of this code being outdated, I'm gonna close and rewrite this example in near future. Sorry for delays on it 🙃

DevTGhosh:
@t3dotgg  looking forward to it.

Issue #429, opened 2023-10-17
feat: Failed uploads should not be returned via utapi.listfiles()


JEK58:
### Describe the feature you'd like to request

Failed uploads should not be returned when calling `utapi.listfiles()` or `utapi.getFileUrls(failedUploadKey)`
Right now there is no way to see if a file url actually works, or not.

### Describe the solution you'd like to see

Either do not return failed uploads, or add a status property to filter them out.

### 👨‍👧‍👦 Contributing

- [ ] 🙋‍♂️ Yes, I'd be down to file a PR implementing this feature!

markflorkowski:
Good call. Will get this out today 👍

markflorkowski:
listFiles should now include a `status` field, matching with what shows on the dashboard

JEK58:
You guys are impressively fast! Thanks!

JEK58:
I opened a PR for this types of `listFiles`: https://github.com/pingdotgg/uploadthing/pull/430

PR #430, opened 2023-10-18
feat: Upload status type added

JEK58:
`listFiles` now includes the status like shown in the Dashboard. (https://github.com/pingdotgg/uploadthing/issues/429) I added the type for this. As I don't know all possible status options it's a string for now. Maybe make it a union with all possible options?

### DIFF
--- a/packages/uploadthing/src/sdk/index.ts
+++ b/packages/uploadthing/src/sdk/index.ts
@@ -248,7 +248,11 @@ export class UTApi {
 
     // TODO: Implement filtering and pagination
     const json = await this.requestUploadThing<{
-      files: { key: string; id: string; status: "Deletion Pending" | "Failed" | "Uploaded" | "Uploading" }[];
+      files: {
+        key: string;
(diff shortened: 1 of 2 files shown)


packages/uploadthing/src/sdk/index.ts
```diff
@@ -248,7 +248,7 @@ export class UTApi {
 
     // TODO: Implement filtering and pagination
     const json = await this.requestUploadThing<{
-      files: { key: string; id: string }[];
+      files: { key: string; id: string; status: string }[];
```
juliusmarminge, line 251: This is the current union:

`"Deletion Pending" | "Failed" | "Uploaded" | "Uploading"`

JEK58: Thanks, I added it. But not sure if you like it better inline or in /types

markflorkowski (approved):
I think inline is fine for this. cc @juliusmarminge

PR #432, opened 2023-10-18
feat!: multipart uploads

juliusmarminge:
Needs https://app.graphite.dev/github/pr/pingdotgg/uploadthing-infra/219/feat-multi-part-uploads

### DIFF
--- a/packages/shared/src/types.ts
+++ b/packages/shared/src/types.ts
@@ -49,6 +49,8 @@ export interface ResponseEsque {
   json<T = unknown>(): Promise<T>;
   text(): Promise<string>;
 
+  headers: Headers;
+
   clone(): ResponseEsque;
 }
 
--- a/packages/uploadthing/src/client.ts
+++ b/packages/uploadthing/src/client.ts
@@ -117,6 +117,8 @@ export const DANGEROUS__uploadFiles = async <TRouter extends FileRouter>(
       pollingUrl,
     } = presigned;
 
+    let uploadedBytes = 0;
+
     let etags: { tag: string; partNumber: number }[];
     try {
       etags = await Promise.all(
@@ -132,9 +134,9 @@ export const DANGEROUS__uploadFiles = async <TRouter extends FileRouter>(
             fileType: file.type,
             fileName: file.name,
             maxRetries: 10,
-            onProgress: (progress) => {
-              const totalUploaded = offset + progress;
-              const percent = (totalUploaded / file.size) * 100;
+            onProgress: (delta) => {
+              uploadedBytes += delta;
+              const percent = (uploadedBytes / file.size) * 100;
               opts.onUploadProgress?.({ file: file.name, progress: percent });
             },
           });
--- a/packages/uploadthing/src/client.ts
+++ b/packages/uploadthing/src/client.ts
@@ -6,35 +6,14 @@ import {
 
 import type { UploadThingResponse } from "./internal/handler";
 import { uploadPartWithProgress } from "./internal/multi-part";
-import { maybeParseResponseXML } from "./internal/s3-error-parser";
-import type {
-  ActionType,
-  FileRouter,
-  inferEndpointInput,
-} from "./internal/types";
+import type { FileRouter, inferEndpointInput } from "./internal/types";
+import { createAPIRequestUrl, createUTReporter } from "./internal/ut-reporter";
 
 /**
  * @internal
  * Shared helpers for our premade components that's reusable by multiple frameworks
  */
 export * from "./internal/component-theming";
-
-const createAPIRequestUrl = (config: {
-  url?: string;
-  slug: string;
-  actionType: ActionType;
-}) => {
-  const url = new URL(
-    config.url ?? `${window.location.origin}/api/uploadthing`,
-  );
-
-  const queryParams = new URLSearchParams(url.search);
-  queryParams.set("actionType", config.actionType);
-  queryParams.set("slug", config.slug);
-
-  url.search = queryParams.toString();
-  return url.toString();
-};
 
 type UploadFilesOptions<TRouter extends FileRouter> = {
   [TEndpoint in keyof TRouter]: {
@@ -60,77 +39,17 @@ export type UploadFileResponse = {
   url: string;
 };
 
-async function alertUTOfError(opts: {
-  url?: string;
-  endpoint: string;
-  key: string;
-  uploadId: string;
-  fileName: string;
-  reason?: string;
-}) {
-  // tell uploadthing infra server that upload failed
-  await fetch(
-    createAPIRequestUrl({
-      url: opts.url,
-      slug: opts.endpoint,
-      actionType: "failure",
-    }),
-    {
-      method: "POST",
-      body: JSON.stringify({
-        fileKey: opts.key,
-        uploadId: opts.uploadId,
-      }),
-    },
-  );
-
-  // Attempt to parse response as XML
-  const parsed = maybeParseResponseXML(opts.reason ?? "");
-
-  // Throw an error for the client
-  if (parsed?.message) {
-    throw new UploadThingError({
-      code: parsed.code,
-      message: parsed.message,
-    });
-  } else {
-    throw new UploadThingError({
-      code: "UPLOAD_FAILED",
-      message: `Failed to upload file ${opts.fileName} to S3`,
-      cause: opts.reason,
-    });
-  }
-}
-
-async function alertUTOfUploadComplete(opts: {
-  url?: string;
-  endpoint: string;
-  uploadId: string;
-  key: string;
-  etags: { tag: string; partNumber: number }[];
-}) {
-  const { url, endpoint, ...payload } = opts;
-  const response = await fetch(
-    createAPIRequestUrl({
-      url,
-      slug: endpoint,
-      actionType: "multipart-complete",
-    }),
-    {
-      method: "POST",
-      body: JSON.stringify(payload),
-    },
-  );
-
-  return response.ok;
-}
-
 export const DANGEROUS__uploadFiles = async <TRouter extends FileRouter>(
   opts: UploadFilesOptions<TRouter>,
   config?: {
     url?: string;
   },
 ) => {
+  const reportEventToUT = createUTReporter({
+    endpoint: String(opts.endpoint),
+    url: config?.url,
+  });
+
   // Get presigned URL for S3 upload
   const s3ConnectionRes = await fetch(
     createAPIRequestUrl({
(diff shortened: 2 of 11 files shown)


juliusmarminge:
* main
  * **PR #432** <a href="https://app.graphite.dev/github/pr/pingdotgg/uploadthing/432" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/pingdotgg/uploadthing/432?utm_source=stack-comment).

packages/uploadthing/src/sdk/utils.ts
```diff
@@ -76,88 +76,135 @@ export const uploadFilesInternal = async (
...
+              console.log(
+                `Retrying part ${partNumber} after ${delay}ms, attempt ${
+                  retryCount + 1
+                }/${maxRetries}`,
+              );
+              return makeRequest();
+            }
+
+            // Max retries exceeded, tell UT server that upload failed
+            await opts.fetch(generateUploadThingURL("/api/failureCallback"), {
```
juliusmarminge, line 167: should we include a prop that this is multipart here, and then on the infra side we can handle that and call `AbortIncompleteMultipartUpload ` ?

packages/shared/src/types.ts
```diff
@@ -49,6 +49,8 @@ export interface ResponseEsque {
   json<T = unknown>(): Promise<T>;
   text(): Promise<string>;
 
+  headers: Headers;
```
juliusmarminge, line 52: HeadersEsque?

packages/uploadthing/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "uploadthing",
-  "version": "5.7.1",
+  "version": "6.7.1",
```
juliusmarminge, line 3: REVERT

packages/shared/src/types.ts
markflorkowski: Oh boy 😅 Standards are so good...

packages/uploadthing/src/client.ts
```diff
@@ -160,114 +175,94 @@ export const DANGEROUS__uploadFiles = async <TRouter extends FileRouter>(
...
+          url,
+          chunk: chunk,
+          contentDisposition,
+          fileType: file.type,
+          fileName: file.name,
+          maxRetries: 10,
+          onProgress: (progress) => {
+            const totalUploaded = offset + progress;
+            const percent = (totalUploaded / file.size) * 100;
+            opts.onUploadProgress?.({ file: file.name, progress: percent });
```
markflorkowski, line 217: Should add a check in the components to make sure we always set the progress to a higher value than it was before, otherwise the progress bar jumps around with out-of-order upload of chunks

correct way would probably be to pass the amount to add rather than the absolute percentage since I don't think there is a way for us to know the absolute amount here. 

https://github.com/pingdotgg/uploadthing/blob/dfbe1a85f777648283282b57f303e3ee57f6a46e/packages/react/src/components/button.tsx#L102C1-L102C1

packages/uploadthing/src/internal/handler.ts
```diff
@@ -237,7 +246,15 @@ export const buildRequestHandler = <TRouter extends FileRouter>(
...
+          !files.every(
+            (f) =>
+              isObject(f) &&
+              typeof f.name === "string" &&
+              typeof f.size === "number",
+          )
+        )
           return new UploadThingError({
             code: "BAD_REQUEST",
             message: "Files must be a string array",
```
markflorkowski, line 260: error message is incorrect now.

packages/uploadthing/package.json
t3dotgg: What is this and why did it happen?

packages/uploadthing/src/client.ts
```diff
@@ -84,32 +54,77 @@ type UploadFilesOptions<TRouter extends FileRouter> = {
...
   key: string;
-  /**
-   * @deprecated
-   * use `url` instead
-   */
-  fileUrl: string;
   url: string;
 };
 
+async function alertUTOfError(opts: {
```
t3dotgg, line 63: nit: `reportErrorToUT`

packages/uploadthing/src/client.ts
```diff
@@ -84,32 +54,77 @@ type UploadFilesOptions<TRouter extends FileRouter> = {
...
+      method: "POST",
+      body: JSON.stringify({
+        fileKey: opts.key,
+        uploadId: opts.uploadId,
+      }),
+    },
+  );
+
+  // Attempt to parse response as XML
+  const parsed = maybeParseResponseXML(opts.reason ?? "");
```
t3dotgg, line 88: We really need an effect equivalent...

packages/uploadthing/src/client.ts
```diff
@@ -84,32 +54,77 @@ type UploadFilesOptions<TRouter extends FileRouter> = {
...
+  } else {
+    throw new UploadThingError({
+      code: "UPLOAD_FAILED",
+      message: `Failed to upload file ${opts.fileName} to S3`,
+      cause: opts.reason,
+    });
+  }
+}
+
+async function alertUTOfUploadComplete(opts: {
```
t3dotgg, line 105: Really don't dig the `alert` prefix for these. Also crazy thought, maybe group them all into one object just for naming's sake?

packages/uploadthing/package.json
juliusmarminge: it's what i use for testing to get the correct handler on the infra side as that's decided by the sdk version :)

packages/uploadthing/src/client.ts
juliusmarminge: Went a bit further https://github.com/pingdotgg/uploadthing/compare/b817f2a0e53e0f33674c17efefa671b84d041939..b4ee21ce96ecef62b025152f0afafd252e236a5f

packages/uploadthing/src/internal/types.ts
```diff
@@ -127,5 +128,30 @@ export type inferEndpointInput<TUploader extends Uploader<any>> =
...
 
-export const VALID_ACTION_TYPES = ["upload", "failure"] as const;
+export const VALID_ACTION_TYPES = [
+  "upload",
+  "failure",
+  "multipart-complete",
+] as const;
 export type ActionType = (typeof VALID_ACTION_TYPES)[number];
+
+export type UTEvents = {
```
juliusmarminge, line 138: in the future maybe we should have a full api spec of some sort where to input to all UT api endpoints are specified and create a fetch wrapper that's typed accordingly, and automatically appends headers

markflorkowski: Yeah, that would be a good idea.

packages/uploadthing/src/internal/handler.ts
```diff
@@ -19,16 +22,35 @@ import { UPLOADTHING_VERSION } from "../constants";
...
 import { getParseFn } from "./parser";
 import { VALID_ACTION_TYPES } from "./types";
-import type { ActionType, FileRouter } from "./types";
+import type { ActionType, FileRouter, UTEvents } from "./types";
+
+/**
+ * Creates a wrapped fetch that will always forward a few headers to the server.
+ */
+const createUTFetch = (apiKey: string) => {
+  return async (endpoint: `/${string}`, payload: unknown) => {
```
markflorkowski, line 31: Potentially needs a method override if we use it everywhere. eg. polling is a `GET` iirc

packages/uploadthing/src/internal/ut-reporter.ts
```diff
@@ -0,0 +1,65 @@
...
+  ) => {
+    const url = createAPIRequestUrl({
+      url: cfg.url,
+      slug: cfg.endpoint,
+      actionType: type,
+    });
+    const response = await fetch(url, {
+      method: "POST",
+      body: JSON.stringify(payload),
+      headers: { "Content-Type": "application/json" },
```
markflorkowski, line 40: should we include the version and key here?

packages/uploadthing/src/internal/ut-reporter.ts
```diff
@@ -0,0 +1,65 @@
...
+    });
+
+    switch (type) {
+      case "failure": {
+        // why isn't this narrowed automatically?
```
markflorkowski, line 45: possibly if you add a type on the switch? eg:
```
switch (type: UTEventType) {
...

```

packages/uploadthing/src/internal/ut-reporter.ts
juliusmarminge: This just goes to the users backend so we dont have the key here

packages/uploadthing/src/internal/handler.ts
juliusmarminge: yup - I'm thinking of expanding this more later with generated types from infra side so will tackle that then, this is enough for now

packages/uploadthing/src/internal/ut-reporter.ts
juliusmarminge: nope :(

packages/uploadthing/src/internal/handler.ts
```diff
@@ -1,34 +1,57 @@
...
+  ContentDisposition,
   ExpandedRouteConfig,
   FileRouterInputKey,
   Json,
   RequestLike,
   UploadedFile,
 } from "@uploadthing/shared";
 
-import { UPLOADTHING_VERSION } from "../constants";
+// import { UPLOADTHING_VERSION } from "../constants";
```
juliusmarminge, line 21:
```suggestion
import { UPLOADTHING_VERSION } from "../constants";
```

packages/uploadthing/src/internal/handler.ts
```diff
@@ -1,34 +1,57 @@
...
+const createUTFetch = (apiKey: string) => {
+  return async (endpoint: `/${string}`, payload: unknown) => {
+    const response = await fetch(generateUploadThingURL(endpoint), {
+      method: "POST",
+      body: JSON.stringify(payload),
+      headers: {
+        "Content-Type": "application/json",
+        "x-uploadthing-api-key": apiKey,
+        // FIXME: REVERT to proper version
+        "x-uploadthing-version": "6.x.x", // UPLOADTHING_VERSION,
```
juliusmarminge, line 39:
```suggestion
        "x-uploadthing-version": UPLOADTHING_VERSION,
```

packages/uploadthing/src/sdk/index.ts
```diff
@@ -6,7 +6,7 @@ import type {
 } from "@uploadthing/shared";
 import { generateUploadThingURL, UploadThingError } from "@uploadthing/shared";
 
-import { UPLOADTHING_VERSION } from "../constants";
+// import { UPLOADTHING_VERSION } from "../constants";
```
juliusmarminge, line 9:
```suggestion
import { UPLOADTHING_VERSION } from "../constants";
```

packages/uploadthing/src/sdk/index.ts
```diff
@@ -39,7 +39,8 @@ export class UTApi {
     this.defaultHeaders = {
       "Content-Type": "application/json",
       "x-uploadthing-api-key": this.apiKey!,
-      "x-uploadthing-version": UPLOADTHING_VERSION,
+      // FIXME: REVERT to proper version
+      "x-uploadthing-version": "6.x.x", // UPLOADTHING_VERSION,
```
juliusmarminge, line 43:
```suggestion
      "x-uploadthing-version": UPLOADTHING_VERSION,
```

markflorkowski (approved):

PR #451, opened 2023-11-02
fix: improved error handling

juliusmarminge:


juliusmarminge:
* main
  * **PR #451** <a href="https://app.graphite.dev/github/pr/pingdotgg/uploadthing/451" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/pingdotgg/uploadthing/451?utm_source=stack-comment).

packages/react/src/useUploadThing.ts
```diff
@@ -89,11 +85,17 @@ export const INTERNAL_uploadthingHookGen = <TRouter extends FileRouter>() => {
...
-          error as UploadThingError<inferErrorShape<TRouter>>,
-        );
+        let error: UploadThingError<inferErrorShape<TRouter>>;
+        if (e instanceof UploadThingError) {
+          error = e as UploadThingError<inferErrorShape<TRouter>>;
+        } else {
+          error = INTERNAL_DO_NOT_USE__fatalClientError(e as Error);
+          console.error(
+            "Something went wrong. Please contact UploadThing and provide the following cause:",
+            error.cause,
```
markflorkowski, line 95: Does this potentially need to be stringified?

juliusmarminge: can you not just log an error as is?

markflorkowski: You lose the `options.cause` if you just log it.
 
image

juliusmarminge: error.cause in our case should just be a normal error without extra props, but added a conditional toString() just in case

markflorkowski (approved):

PR #453, opened 2023-11-03
feat: adds a pre-processor callback onBeforeUploadBegin

joelhooks:
This adds an `onBeforeUploadBegin` callback that is called before `startUpload` to allow for pre-processing of files (ie changing file names etc)

### DIFF
--- a/.changeset/empty-birds-beam.md
+++ b/.changeset/empty-birds-beam.md
@@ -1,5 +1,5 @@
 ---
-"uploadthing": patch
+"uploadthing": minor
 "@uploadthing/react": patch
 "@uploadthing/solid": patch
 ---
--- a/.changeset/empty-birds-beam.md
+++ b/.changeset/empty-birds-beam.md
@@ -1,7 +1,7 @@
 ---
 "uploadthing": minor
-"@uploadthing/react": patch
-"@uploadthing/solid": patch
+"@uploadthing/react": minor
+"@uploadthing/solid": minor
 ---
 
 Adds an onBeforeUploadBegin callback that is called before startUpload to allow
--- a/.changeset/empty-birds-beam.md
+++ b/.changeset/empty-birds-beam.md
@@ -1,0 +1,8 @@
+---
+"uploadthing": patch
+"@uploadthing/react": patch
+"@uploadthing/solid": patch
+---
+
+Adds an onBeforeUploadBegin callback that is called before startUpload to allow
+for pre-processing of files (ie changing file names etc)
--- a/packages/react/src/components/button.tsx
+++ b/packages/react/src/components/button.tsx
@@ -104,6 +104,7 @@ export function UploadButton<TRouter extends FileRouter>(
       },
       onUploadError: $props.onUploadError,
       onUploadBegin: $props.onUploadBegin,
+      onBeforeUploadBegin: $props.onBeforeUploadBegin,
     },
   );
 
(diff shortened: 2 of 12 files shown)


packages/react/src/components/dropzone.tsx
```diff
@@ -106,15 +106,15 @@ export function UploadDropzone<TRouter extends FileRouter>(
 
   const onDrop = useCallback(
     (acceptedFiles: FileWithPath[]) => {
+      if ($props.onBeforeUploadBegin) {
```
juliusmarminge, line 109: do you want this to run on every drop? The naming doesn't really tell that it will

packages/react/src/components/button.tsx
```diff
@@ -123,10 +124,16 @@ export function UploadButton<TRouter extends FileRouter>(
       const pastedFiles = getFilesFromClipboardEvent(event);
       if (!pastedFiles) return;
 
-      setFiles((prev) => [...prev, ...pastedFiles]);
+      let filesToUpload = pastedFiles;
+      if ($props.onBeforeUploadBegin) {
+        filesToUpload = $props.onBeforeUploadBegin(pastedFiles);
+      }
```
juliusmarminge, line 130: nit

```suggestion
const filesToUpload = $props.onBeforeUploadBegin?.(pastedFiles) ?? pastedFiles;
```

juliusmarminge (approved):
Aside from the naming conflict if ran on every drop, looks good to me!

packages/react/src/components/dropzone.tsx
joelhooks: from what I can tell it needs to run before `files` gets set for the rename to be "complete" on all the callbacks like `onUploadComplete`, so with that in mind I added it here.

in [other libs](https://github.com/odysseyscience/react-s3-uploader/blob/master/s3upload.js#L34-L36) I've seen similar functions called `scrubFilename` which might be more accurate in terms of intent versus describing when it will upload

FilipByren:
🚀  - excited for this

joelhooks (commented):
@juliusmarminge Love the cleaner updates

juliusmarminge:
- Refactored some more (the function only has to be called in startUpload itself now)
- Added the callback to the docs

juliusmarminge (approved):

.changeset/empty-birds-beam.md
```diff
@@ -0,0 +1,8 @@
+---
+"uploadthing": patch
```
markflorkowski, line 2: I would probably call this a minor. In my mind, we should generally try to version based on:
patch --> bugfix, style changes, or other very small changes 
minor --> "typical" release (new features)
major --> breaking changes

markflorkowski (approved):

.changeset/empty-birds-beam.md
juliusmarminge: Agreed, also added docs based on this being a minor. Missed that 🤩

juliusmarminge, line 2:
```suggestion
"uploadthing": minor
```

.changeset/empty-birds-beam.md
```diff
@@ -0,0 +1,8 @@
+---
+"uploadthing": minor
+"@uploadthing/react": patch
+"@uploadthing/solid": patch
```
juliusmarminge, line 4:
```suggestion
"@uploadthing/react": minor
"@uploadthing/solid": minor
```

joelhooks:
thanks for getting this over the finish line @juliusmarminge 🎉

PR #486, opened 2023-11-21
fix: smarter URL detection

juliusmarminge:
un-deprecated `config.callbackUrl` as it could be useful for some edge cases when we can't autodetect the URL.

makes the parsing of both `process.env.UPLOADTHING_URL` and `config.callbackUrl` smarter to handle relative, host and full URL strings.

### DIFF
--- a/packages/uploadthing/src/internal/handler.ts
+++ b/packages/uploadthing/src/internal/handler.ts
@@ -19,6 +19,7 @@ import type {
 
 import { UPLOADTHING_VERSION } from "../constants";
 import { conditionalDevServer } from "./dev-hook";
+import { getFullApiUrl } from "./get-full-api-url";
 import { getParseFn } from "./parser";
 import { VALID_ACTION_TYPES } from "./types";
 import type { ActionType, FileRouter, UTEvents } from "./types";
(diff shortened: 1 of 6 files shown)


juliusmarminge:
* main
  * **PR #486** <a href="https://app.graphite.dev/github/pr/pingdotgg/uploadthing/486?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/pingdotgg/uploadthing/486?utm_source=stack-comment).

markflorkowski (approved):

packages/uploadthing/src/internal/handler.ts
```diff
@@ -339,15 +336,32 @@ export const buildRequestHandler = <TRouter extends FileRouter>(
           });
         }
 
-        const callbackUrl = url.origin + url.pathname;
+        let callbackUrl = url;
+        if (config?.callbackUrl) {
```
markflorkowski, line 340: should this override `input.url`?

juliusmarminge: input.url is a thing some of our adapters injects as some of them doesn't have an absolute URL on `req.url`. But I think the user should be able to override this as the automatic detection doesn't work everytime (as mentioned in the docs for this PR).

PR #500, opened 2023-11-30
fix: url parsing from headers if req.url is localhost

juliusmarminge:
this should handle almost everything i hope 😅

### DIFF
--- a/packages/uploadthing/src/internal/handler.ts
+++ b/packages/uploadthing/src/internal/handler.ts
@@ -347,23 +347,22 @@ export const buildRequestHandler = <TRouter extends FileRouter>(
           process.env.NODE_ENV === "production" &&
           callbackUrl.host.includes("localhost")
         ) {
-          console.warn(
-            [
-              "[UT] [WARN] You are using a localhost callback url in production which is not supported.",
-              "Read more and learn how to fix it here: https://uploadthing.com/faq#my-callback-runs-in-development-but-not-in-production",
-            ].join(" "),
-          );
-          const attemptToParseThroughHeaders =
+          const attemptToParseThroughHeaders = (
             getHeader(req, "origin") ??
             getHeader(req, "referer") ??
-            getHeader(req, "host");
-          if (attemptToParseThroughHeaders) {
-            callbackUrl = getFullApiUrl(
-              attemptToParseThroughHeaders.toString(),
-            );
-            console.log(
-              "[UT] [INFO] Falled back to parsing callback url from headers:",
-              callbackUrl.href,
+            getHeader(req, "host")
+          )?.toString();
+          if (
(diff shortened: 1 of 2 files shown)


juliusmarminge:
* main
  * **PR #500** <a href="https://app.graphite.dev/github/pr/pingdotgg/uploadthing/500?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/pingdotgg/uploadthing/500?utm_source=stack-comment).

packages/uploadthing/src/internal/handler.ts
```diff
@@ -336,24 +336,7 @@ export const buildRequestHandler = <TRouter extends FileRouter>(
...
-          callbackUrl.host.includes("localhost")
-        ) {
-          console.warn(
-            [
-              "[UT] [WARN] You are using a localhost callback url in production which is not supported.",
-              "Read more and learn how to fix it here: https://uploadthing.com/faq#my-callback-runs-in-development-but-not-in-production",
-            ].join(" "),
-          );
-        }
+        const callbackUrl = resolveCallbackUrl({ config, req, url });
```
markflorkowski, line 339: This is only used once below, can we just inline it?

markflorkowski: Also is there a reason we cant just pass callbackUrl in the fetch, and need to do

`callbackUrl: callbackUrl.origin + callbackUrl.pathname,` ?

packages/uploadthing/src/internal/handler.ts
```diff
@@ -488,6 +471,55 @@ export const buildRequestHandler = <TRouter extends FileRouter>(
...
+  let parsedFromHeaders = (
+    getHeader(opts.req, "origin") ??
+    getHeader(opts.req, "referer") ??
+    getHeader(opts.req, "host") ??
+    getHeader(opts.req, "x-forwarded-host")
+  )?.toString();
+
+  if (parsedFromHeaders && !parsedFromHeaders.includes("http")) {
+    parsedFromHeaders =
+      (getHeader(opts.req, "x-forwarded-proto") ?? "https").toString() +
```
markflorkowski, line 504: L503-507 is a bit hard to parse. Mind breaking this out above, then using a template string?

markflorkowski (approved):

packages/uploadthing/src/internal/handler.ts
juliusmarminge: didn't wanna pass any potential search params etc to UT

markflorkowski: ahh makes sense.

PR #524, opened 2023-12-17
refactor: explore minimal dropzone

juliusmarminge:
Explores a minimal useDropzone as well as pulling it out to a separate package with utilities that can be reused for solid and vue implementations

This fits both the React version as well as **adding** a Solid implementation whilst still removing code 🤯 
CleanShot 2023-12-18 at 09 11 44@2x

Ref: UPL-66


packages/dropzone/src/react.tsx
```diff
@@ -0,0 +1,347 @@
...
+ *   return (
+ *     <div {...getRootProps()}>
+ *       <input {...getInputProps()} />
+ *       <p>Drag and drop some files here, or click to select files</p>
+ *     </div>
+ *   )
+ * }
+ * ```
+ */
+export function useDropzone({
```
juliusmarminge, line 56: this should be easily portable to solid, vue etc now and can use all the helpers from `core.ts`

packages/dropzone/src/solid.tsx
juliusmarminge, line 1: this should probably live in community package but proves the point that it's easy to make one that's not react and re-use a lot of the same functionality

packages/dropzone/src/core.ts
```diff
@@ -0,0 +1,200 @@
...
+export function isEnterOrSpace(event: { key?: string; keyCode?: number }) {
+  return (
+    ("key" in event && (event.key === " " || event.key === "Enter")) ||
+    ("keyCode" in event && (event.keyCode === 32 || event.keyCode === 13))
+  );
+}
+
+const isDefined = <T>(v: T | null | undefined): v is T => v != null;
+export function isValidSize(file: File, minSize: number, maxSize: number) {
+  if (isDefined(file.size)) {
```
t3dotgg, line 25: Can we early escape if not defined maybe? Would clean up the nesting of this logic a bit

t3dotgg: Oh is this code ripped from the lib? If so might be worth commenting on top

packages/dropzone/src/solid.tsx
t3dotgg: You can keep it here for now and I'll deal with it

t3dotgg (approved):
Okay fiiiiiine

packages/dropzone/src/core.ts
juliusmarminge: Most of it is the same as before (i.e. forked from react-dropzone) and some of it I cleaned up a bit when removing stuff we didnt need

packages/dropzone/src/core.ts
```diff
@@ -0,0 +1,206 @@
...
+
+export function isEnterOrSpace(event: { key?: string; keyCode?: number }) {
+  return (
+    ("key" in event && (event.key === " " || event.key === "Enter")) ||
+    ("keyCode" in event && (event.keyCode === 32 || event.keyCode === 13))
+  );
+}
+
+const isDefined = <T>(v: T | null | undefined): v is T => v != null;
+export function isValidSize(file: File, minSize: number, maxSize: number) {
```
juliusmarminge, line 31: Fixed it @t3dotgg - not sure how much of this is even needed, files should always have a size on them and minSize and maxSize are typed as numbers, so there shouldn't be a case where those are nullish either...

can probably be a one-liner

```ts
export const isValidSize = (file: File, minSz: number, maxSz: number) => file.size >= minSz && file.size <= maxSz;
```

packages/react/src/components/dropzone.tsx
```diff
@@ -1,6 +1,8 @@
 import { useCallback, useEffect, useState } from "react";
 import { twMerge } from "tailwind-merge";
 
+import type { FileWithPath } from "@uploadthing/dropzone/react";
```
markflorkowski, line 4: It feels a bit strange that this is where that type comes from, but probably fine.

markflorkowski (approved):

packages/react/src/components/dropzone.tsx
juliusmarminge: i don't see where else we'd get it from, unless we just make it a generic File instead?

markflorkowski: We could have an uploadthing/types package, but that makes it weird elsewhere, and also makes maintenance a pain. It's fine as is.

juliusmarminge: I did standard File: https://github.com/pingdotgg/uploadthing/pull/524/commits/605fdb64819a74c51fdd07f513dbb4d96cfd7cdc

PR #529, opened 2023-12-25
feat: allow async onBeforeUploadBegin callback

ibrahimyaacob92:
- PR is created to counter the issue faced when trying to validate the uploaded files with async function

image

### DIFF
--- a/packages/react/src/useUploadThing.ts
+++ b/packages/react/src/useUploadThing.ts
@@ -42,7 +42,7 @@ export type UseUploadthingProps<
   onUploadProgress?: (p: number) => void;
   onUploadError?: (e: UploadThingError<inferErrorShape<TRouter>>) => void;
   onUploadBegin?: (fileName: string) => void;
-  onBeforeUploadBegin?:  (files: File[]) => Promise<File[]> | File;
+  onBeforeUploadBegin?: (files: File[]) => Promise<File[]> | File[];
 };
 
 export const INTERNAL_uploadthingHookGen = <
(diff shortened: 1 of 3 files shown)


packages/react/src/useUploadThing.ts
```diff
@@ -14,6 +14,7 @@ import type {
   inferEndpointOutput,
   inferErrorShape,
 } from "uploadthing/server";
+import type { MaybePromise } from "../../uploadthing/src/internal/types";
```
juliusmarminge, line 17: This import path wont work. I dont know if we export it from a public entrypoint, if not you can just inline the union type i guess

juliusmarminge (changes requested):

packages/react/src/useUploadThing.ts
ibrahimyaacob92: @juliusmarminge updated

juliusmarminge (approved):

packages/react/src/useUploadThing.ts
```diff
@@ -42,7 +42,7 @@ export type UseUploadthingProps<
   onUploadProgress?: (p: number) => void;
   onUploadError?: (e: UploadThingError<inferErrorShape<TRouter>>) => void;
   onUploadBegin?: (fileName: string) => void;
-  onBeforeUploadBegin?: (files: File[]) => File[];
+  onBeforeUploadBegin?:  (files: File[]) => Promise<File[]> | File;
```
juliusmarminge, line 45: File[]

juliusmarminge:
```suggestion
  onBeforeUploadBegin?: (files: File[]) => Promise<File[]> | File[];
```

juliusmarminge (changes requested):
Just formatting left, i dont have my laptop at hand so cant do it myself

juliusmarminge (approved):

PR #542, opened 2024-01-10
chore: setup consola

juliusmarminge:


packages/react/package.json
```diff
@@ -85,6 +85,9 @@
     "extends": [
       "@uploadthing/eslint-config/base",
       "@uploadthing/eslint-config/react"
-    ]
+    ],
+    "rules": {
+      "no-console": "off"
```
juliusmarminge, line 90: added these as i am unsure if we should ship these kb's to the browser or keep it server-only?

markflorkowski: probably server only?

PR #545, opened 2024-01-11
feat: logger for serverside stuff

juliusmarminge:
Adds consola logger, adds a config option to enable more verbose logging, e.g.:

```ts
export const { GET, POST } = createNextRouteHandler({
  router: uploadRouter,
  config: {
    logLevel: "debug",
  },
});
```

### TODO

- [ ] discuss / decide what levels we want, now i just do `.error`, `.info` and `.debug` as i feel fit
- [x] add more debug logs

### Example output:

CleanShot 2024-01-11 at 14 59 27@2x

### Example output with `logLevel: "debug"` set:

CleanShot 2024-01-11 at 17 14 15@2x

### DIFF
--- a/.changeset/gorgeous-icons-dress.md
+++ b/.changeset/gorgeous-icons-dress.md
@@ -1,0 +1,6 @@
+---
+"uploadthing": minor
+"@uploadthing/shared": minor
+---
+
+feat: add `logLevel` option to enable more verbose logging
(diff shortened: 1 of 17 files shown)


markflorkowski (approved):

markflorkowski:
>  discuss / decide what levels we want, now i just do .error, .info and .debug as i feel fit

These seem fine to me for now. I can't think of anywhere I would have been more granular at the moment. 

maybe `.fatal` any time before a `process.exit()` but I think we only do that once if ever

PR #547, opened 2024-01-13
Bunchee 2

juliusmarminge:


packages/solid/package.json
juliusmarminge, line 1: TODO: handle solid later

juliusmarminge:
this was a hassle -_-

maybe i'll try again one day. super impressed with bunchee so far

PR #564, opened 2024-01-23
chore: (analytics) Report which UT frontend package is being used to make requests

markflorkowski:


packages/react/src/useUploadThing.ts
```diff
@@ -103,6 +103,7 @@ export const INTERNAL_uploadthingHookGen = <
             opts.onUploadBegin(file);
           },
           url: initOpts.url,
+          package: "uploadthing-react",
```
juliusmarminge, line 106:
```suggestion
          package: "@uploadthing/react",
```

packages/react/src/useUploadThing.ts
```diff
@@ -164,6 +165,7 @@ export const generateReactHelpers = <TRouter extends FileRouter>(initOpts?: {
       DANGEROUS__uploadFiles<TRouter, TEndpoint>(endpoint, {
         ...opts,
         url,
+        package: "uploadthing-react",
```
juliusmarminge, line 168:
```suggestion
        package: "@uploadthing/react",
```

packages/uploadthing/src/client.ts
```diff
@@ -45,6 +45,14 @@ type UploadFilesOptions<
    * @example URL { https://www.example.com/api/uploadthing }
    */
   url: URL;
+
+  /**
+   * The uploadthing package that is making this request
+   * @example "uploadthing-react"
```
juliusmarminge, line 51:
```suggestion
   * @example "@uploadthing/react"
```

packages/uploadthing/src/client.ts
```diff
@@ -243,10 +253,20 @@ export const genUploader = <TRouter extends FileRouter>(initOpts: {
    * @default (VERCEL_URL ?? window.location.origin) + "/api/uploadthing"
    */
   url?: string | URL;
+
+  /**
+   * The uploadthing package that is making this request
+   * @example "uploadthing-react"
```
juliusmarminge, line 259:
```suggestion
   * @example "@uploadthing/react"
```

packages/uploadthing/src/internal/handler.ts
```diff
@@ -151,6 +152,7 @@ export const buildRequestHandler = <TRouter extends FileRouter>(
     const uploadthingHook = getHeader(req, "uploadthing-hook") ?? undefined;
     const slug = params.get("slug") ?? undefined;
     const actionType = (params.get("actionType") as ActionType) ?? undefined;
+    const utPackage = getHeader(req, "x-ut-package") ?? "unknown";
```
juliusmarminge, line 155: why is this `x-ut-package` when its `x-uploadthing-package` above?

markflorkowski: because I am not good at writing code.

juliusmarminge (approved):
Should we also track what backend adapter is used?

markflorkowski:
> Should we also track what backend adapter is used?

Probably not a bad idea

Issue #565, opened 2024-01-23
feat: A way to know how many files are incoming in middleware


kwerdna19:
### Describe the feature you'd like to request

Please correct me if Im wrong but I dont think there is a way to tell how many files are incoming through a router middleware.

This is the use case I had in mind- on this app a user can upload a maximum of 5 images ever (at different times) - say they upload 3 images one day. And the next day they try to upload 3 more images, in the middleware I would like to lookup the users upload count and see if this count+incoming image count is less than 5, and if not, throw an error

For this, I would need the incoming image count somehow.

### Describe the solution you'd like to see

A parameter in the middleware func that tells us how many files are being processed

### Additional information

The same use case can be improved on the frontend too, as it seems there is no way to dynamically specify how many files are allowed to be dropped into a UTDropZone for example. But I think this can be left for a separate feature request / future consideration

### 👨‍👧‍👦 Contributing

- [x] 🙋‍♂️ Yes, I'd be down to file a PR implementing this feature!

kwerdna19:
On looking at this further- it seems although we have access to the NextRequest, we cannot call `req.json()` because it has already been consumed by the UT internals. Attempting to call `req.clone()` seems to error out as well, for a reason Im not sure of.

But regardless, I think the fix is to pass the body back to the middleware after it is read the first time. That does raise the question of how to type this- since it is an arbitrarily parsed json body, its type is technically `any`. Should we use zod to ensure the type of this body ?

I can contribute a PR, just not sure exactly what the preference would be on that last question

markflorkowski:
The way I would currently do what you are describing is using the `input` option to pass the number of files along to middleware, would that work for you?

Relevant docs:
- Server: https://docs.uploadthing.com/api-reference/server#input-1
- Client: https://docs.uploadthing.com/api-reference/react#configuration

kwerdna19:
> The way I would currently do what you are describing is using the `input` option to pass the number of files along to middleware, would that work for you?
> 
> Relevant docs:
> 
> * Server: https://docs.uploadthing.com/api-reference/server#input-1
> * Client: https://docs.uploadthing.com/api-reference/react#configuration

Hello, thanks for the workaround suggestion - it seems valid, albeit maybe needlessly more code, but a couple things.

How would this work in practice. I am trying it out right now and having trouble.

[...]

PR #568, opened 2024-01-23
feat: add files to middleware args

markflorkowski:
Addresses #565


juliusmarminge (approved):
Probably a way to not having to pass this arg to every adapter type but oh well

juliusmarminge (approved):

PR #569, opened 2024-01-24
refactor: split component generators for better tree-shakability

juliusmarminge:
Currently if one only uses the button the entire dropzone code (including useDropzone) will be included in the bundle

This is mitigated by adding `generateUploadButton`, `generateUploadDropzone` and `generateUploader` in favor of (now deprecated) `generateComponents` for better treeshakability.

Also does some general cleanup and improves re-usabiity of some types


packages/uploadthing/src/h3.ts
juliusmarminge, line 1: no idea why this is needed, old code type-errored for some reason. Had the same occur on the Nuxt branch 🤔

packages/react/src/hooks.ts
juliusmarminge, line 1: idk what the rationale was for this export but might as well export from main? everything requires client components anyways and the main entrypoint was already flagged with it

packages/uploadthing/src/h3.ts
markflorkowski: weird 🤔

markflorkowski (approved):

PR #573, opened 2024-01-25
fix(h3): always return a response to support non-preemptive routers

pi0:
See https://github.com/unjs/h3/issues/628#issuecomment-1910087778 for more context.

### DIFF
--- a/packages/uploadthing/src/h3.ts
+++ b/packages/uploadthing/src/h3.ts
@@ -71,7 +71,7 @@ export const createRouteHandler = <TRouter extends FileRouter>(
       return "An unknown error occurred";
     }
 
-    return response.body;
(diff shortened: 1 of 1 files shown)


packages/uploadthing/src/h3.ts
```diff
@@ -71,7 +71,7 @@ export const createRouteHandler = <TRouter extends FileRouter>(
       return "An unknown error occurred";
     }
 
-    return response.body;
+    return response.body ?? "OK";
```
pi0, line 74: It can be also an empty string btw or whatever. only not nullish

juliusmarminge:
Thank you 🙏

juliusmarminge (approved):

Issue #570, opened 2024-01-25
[docs]: update to use non-deprecated imports


shiqocred:
### Provide environment information

```bash
System:
    OS: macOS 12.2.1
    CPU: (4) x64 Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
    Memory: 49.79 MB / 8.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 20.9.0 - /usr/local/bin/node
    Yarn: 1.22.21 - /usr/local/bin/yarn
    npm: 10.1.0 - /usr/local/bin/npm
    pnpm: 8.14.3 - /usr/local/bin/pnpm
    bun: 1.0.3 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 120.1.61.116
    Chrome: 120.0.6099.234
    Safari: 15.3
  npmPackages:
    @uploadthing/react: ^6.2.1 => 6.2.1 
    typescript: 5.3.2 => 5.3.2 
    uploadthing: ^6.3.0 => 6.3.0
```

### Describe the bug

when i create component with this code: 
```import { generateComponents } from "@uploadthing/react";

import type { OurFileRouter } from "@/app/api/uploadthing/core";

export const { UploadButton, UploadDropzone, Uploader } =
  generateComponents<OurFileRouter>();
```

i get error `'generateComponents' is deprecated`

### Link to reproduction

-

### To reproduce

-

### 👨‍👧‍👦 Contributing

- [ ] 🙋‍♂️ Yes, I'd be down to file a PR fixing this bug!

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

markflorkowski:
This is correct. Is there documentation that still recommends using `generateComponents()` ?

markflorkowski:
Docs have been updated.

PR #575, opened 2024-01-25
chore(docs): Update docs to use new split component generators

markflorkowski:
Addresses #570

### DIFF
--- a/docs/src/pages/api-reference/react.mdx
+++ b/docs/src/pages/api-reference/react.mdx
@@ -1,4 +1,14 @@
-## generateComponents
+import { Callout } from "nextra-theme-docs";
+
+## Generated Components
+
+### generateComponents
+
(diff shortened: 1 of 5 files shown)


docs/src/pages/api-reference/react.mdx
```diff
@@ -14,6 +24,30 @@ export const { UploadButton, UploadDropzone } =
...
+
+```tsx copy filename="utils/uploadthing.tsx"
+import { generateUploadButton } from "@uploadthing/react";
+
+export const UploadButton = generateUploadButton<OurFileRouter>();
+```
+
+### generateUploadDropzone
+
+The `generateUploadDropzone` function is used to generate the UploadButton
```
juliusmarminge, line 41: UploadDropzone

PR #578, opened 2024-01-26
chore(astro): improve example, add docs

juliusmarminge:


docs/src/pages/getting-started/_meta.json
```diff
@@ -4,5 +4,5 @@
   "solid": "SolidStart",
   "svelte": "SvelteKit (Coming Soon)",
   "nuxt": "Nuxt.js (Coming Soon)",
-  "astro": "Astro (Coming Soon)"
+  "astro": "Astro"
```
t3dotgg, line 7: Should be moved above the "coming soon" options

docs/src/pages/getting-started/astro.mdx
```diff
@@ -1 +1,229 @@
...
+import EnvSetup from "./_shared/env-setup.mdx";
+import TailwindSetup from "./_shared/tailwind-setup.mdx";
+
+# Astro Setup
+
+You can use Astro with any of the frontend frameworks we support. This guide
+will walk you through setting it up using React, but the steps will be near
+identical for other frameworks too.
+
+<Callout type="info">
```
t3dotgg, line 12: Not in love with this. It's fine but I'd love for us to find a better place to visibly include "examples" in future

docs/src/pages/getting-started/astro.mdx
```diff
@@ -1 +1,229 @@
...
+    }),
+} satisfies FileRouter;
+
+export type OurFileRouter = typeof ourFileRouter;
+```
+
+### Create a Astro API route using the FileRouter
+
+<Callout type="info">
+  File path here doesn't matter, you can serve this from any route. We recommend
```
t3dotgg, line 71: I would flip this to "We recommend serving the UploadThing endpoint out of /api/uploadthing, but you [can change the route if you like]" with a link to the custom endpoint doc

docs/src/pages/getting-started/astro.mdx
```diff
@@ -1 +1,229 @@
...
+import type { OurFileRouter } from "~/server/uploadthing";
+
+export const UploadButton = generateUploadButton<OurFileRouter>();
+export const UploadDropzone = generateUploadDropzone<OurFileRouter>();
+```
+
+### Mount A Button And Upload!
+
+<Callout type="info">
+  Don't forget to add the `"use client;"` directive at the top of your file,
```
t3dotgg, line 132: Is this needed in Astro? Should we be telling them to use the client directives on the Astro side?

docs/src/pages/getting-started/astro.mdx
juliusmarminge: just copied from next docs

docs/src/pages/getting-started/astro.mdx
juliusmarminge: whoops nope

docs/src/pages/getting-started/astro.mdx
juliusmarminge: agree - but let's do that in another PR where we deal with that explicitely for all doc/example combos

docs/src/pages/getting-started/astro.mdx
juliusmarminge: i wont change it in this PR, do a follow-up changing all of them instead

PR #582, opened 2024-01-26
feat: float error messages to client for UploadThingErrors

GentikSolm:
When the user throws errors via `throw new UploadThingError` in middleware, we should pass the error message to the client.
The docs are currently unclear about this. As of now, errors and their messages are completely hidden contrary to the docs
in some areas. This will let users easily push errors to the client without leaking sensitive information

### DIFF
--- a/docs/src/pages/api-reference/server.mdx
+++ b/docs/src/pages/api-reference/server.mdx
@@ -233,7 +233,7 @@ f(["image"])
 
     // Throw if user isn't signed in
     if (!user)
-      throw new Error("You must be logged in to upload a profile picture");
+      throw new UploadThingError("You must be logged in to upload a profile picture");
 
     // Return userId to be used in onUploadComplete
     return { userId: user.id };
@@ -242,6 +242,12 @@ f(["image"])
     console.log("Uploaded by user", metadata.userId);
   });
 ```
+
+<Callout type="info">
+  Note: By default, a thrown `UploadThingError`'s message will be sent to the client's `onError`.
+  All other errors are turned into generic failure messages to avoid leaking sensitive
+  information.
+</Callout>
 
 ### `onUploadComplete`
 
--- a/docs/src/pages/api-reference/server.mdx
+++ b/docs/src/pages/api-reference/server.mdx
@@ -233,7 +233,9 @@ f(["image"])
 
     // Throw if user isn't signed in
     if (!user)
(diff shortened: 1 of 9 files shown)


juliusmarminge (commented):
Why do we need to do the nested UploadThingError's in the first place?

In the catch block of the middleware for example, just rethrow the original error if it was an UploadThing error instead of wrapping it in a new one?

juliusmarminge (approved):
LGTM - just needs some docs

docs/src/pages/api-reference/server.mdx
```diff
@@ -233,7 +233,7 @@ f(["image"])
 
     // Throw if user isn't signed in
     if (!user)
-      throw new Error("You must be logged in to upload a profile picture");
+      throw new UploadThingError("You must be logged in to upload a profile picture");
```
juliusmarminge: missing import

docs/src/pages/errors.mdx
```diff
@@ -7,15 +7,51 @@
...
+
+const auth = (req: Request) => ({ id: "fakeId" });
+
+export const ourFileRouter = {
+  imageUploader: f({ image: { maxFileSize: "4MB" } })
+    .middleware(async ({ req }) => {
+      const user = await auth(req);
+
+      if (!user) throw new Error(`Cant find user from req: ${req.toString()}`); // client onError will get "Failed to run middleware"
+      if(!user.id) throw new UploadThingError('No user ID') // client onError will get "No user ID"
```
juliusmarminge: maybe mention object constructor syntax too as that can be used to alter the http status code

docs/src/pages/errors.mdx
```diff
@@ -7,18 +7,101 @@
...
+   */
+      code?: keyof typeof ERROR_CODES;
+      /**
+       * Your error message describing what happened
+       * @default `An unknown error occurred`
+       */
+      message?: string;
+      /**
+       * The original error that caused this, if any.
+       * @default `env.UPLOADTHING_KEY`
```
markflorkowski, line 90: default seems incorrect?

docs/src/pages/errors.mdx
juliusmarminge: Yea so you could do like a 403 for example. Might wanna add a couple more generic ones though to make it truly usable

juliusmarminge (approved):

PR #586, opened 2024-01-30
FIXME: make express adapter work when there is a bodyparser

juliusmarminge:
this breaks:

CleanShot 2024-01-30 at 09 32 33@2x

move the bodyparser middleware below the uploadthing route handler to fix it. the new logs helps narrow down the issue but we can probably make this scenario work


juliusmarminge:
moved to #590

PR #587, opened 2024-01-30
feat: add customIds

juliusmarminge:
the API here might not be ideal but anything else would require a breaking change i think...


markflorkowski (approved):

.changeset/lucky-carrots-sin.md
```diff
@@ -0,0 +1,28 @@
...
+
+feat: add ability to provide custom identifiers when uploading files
+
+also adds ability for UTApi methods to filter based on custom identifiers
+
+```ts
+// bind custom id when uploading
+f(["image"])
+  .middleware(({ files }) => {
+    const filesWithIds = files.map((f) => ({
```
t3dotgg, line 14: IMO this is fine for now, but we should rethink the structure of this builder pattern in the future. Likely needs to be multi-step, i.e. "authenticate", "preUpload" and "onUploadComplete"

PR #514, opened 2023-12-12
fix: support cf workers in dev mode (wrangler)

juliusmarminge:
Closes #515

cc https://discord.com/channels/966627436387266600/1187130590458290246

- uses `std-env` for safe `process.env` access, no need to worry about process not being defined or having to wrangle with try-catch blocks and `import.meta.env`
- adds `isDev` and `fetch` options to handler configs to override default behaviors
- stores the dev-hook's promise in a varable that's returned in the response that can be awaited / handled as necessary depending on runtime
- changes internal handler function to use native `Request`. Adds a new prop `originalRequest` that's forwarded to the middleare function
- enabled `no-restricted-globals` for `fetch` and `process` to prohibit us from using them in the future. `process` should be imported from `std-env` and `fetch` must be passed by props so it's overridable

## Run it

```bash
pnpm i
pnpm build

cd examples/backend-adapters
cp server/.dev.vars.example server/.dev.vars # fill in the UPLOADTHING_SECRET
pnpm dev:cf
```

PR #592, opened 2024-01-31
fix(server-handler): forward original request

juliusmarminge:
Regression from https://github.com/pingdotgg/uploadthing/pull/514

### DIFF
--- a/packages/uploadthing/src/internal/handler.ts
+++ b/packages/uploadthing/src/internal/handler.ts
@@ -136,7 +136,7 @@ export const buildRequestHandler = <TRouter extends FileRouter>(
     nativeRequest: Request;
 
     // Forward to middleware handler
-    originalRequest?: unknown;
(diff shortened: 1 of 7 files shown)


packages/uploadthing/src/internal/handler.ts
```diff
@@ -128,17 +133,20 @@ export type UploadThingResponse = {
   chunkSize: number;
 }[];
 
-export const buildRequestHandler = <TRouter extends FileRouter>(
+export const buildRequestHandler = <
+  TRouter extends FileRouter,
+  Args extends MiddlewareFnArgs<any, any, any>,
```
juliusmarminge, line 138: added the generic here for extra typesafety and prevent future regressions when adding new adapters etc. now it must match what the adapter has put in their `createUploadthing()` generic

markflorkowski (approved):

PR #593, opened 2024-01-31
fix: Fixe broken tests so release works

markflorkowski:
version numbers 🤦


.prettierignore
```diff
@@ -24,3 +24,5 @@ yarn-error.log*
 dist
 .solid
 pnpm-lock.yaml
+
+CHANGELOG.md
```
juliusmarminge, line 28: Does it need to be globbed to nested folders?

juliusmarminge (approved):
Let's make the test use a fuzzy matcher or filter the version out with a `stringMatching` maybe in a follow up

.prettierignore
markflorkowski: It passes just like this... 🤷

PR #597, opened 2024-02-03
docs: update usage of `UploadThingError`

### LINKED ISSUE #589: docs: Next.js (app/) UploadThingError is not present in installed packages
### Area of Improvement

I don't know if UploadThingError should be inside of any of the installed packages that the [documentation](url) instructs by the line
`npm install uploadthing @uploadthing/react` or not. Also there are somethings that leads to somewhat ambiguity.
 
1. Though UploadThingError is used, there is no line indicating from where you imported.
2. In the Next.js (pages/) docs it's using the Error from Node itself, so you can't really try to guess.
3. In the [Error Handling](https://docs.uploadthing.com/errors) section there is a line for importing the function  `import { UploadThingError } from "uploadthing/server";` But there is not UploadThingError inside, at least it's not showing in TS.

image

### Link to related docs

https://docs.uploadthing.com/getting-started/appdir#creating-your-first-fileroute

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

- [ ] 🙋‍♂️ Yes, I'd be down to file a PR implementing the suggested changes!

petter:
Closes #589.

### DIFF
--- a/docs/src/pages/getting-started/appdir.mdx
+++ b/docs/src/pages/getting-started/appdir.mdx
@@ -25,6 +25,7 @@ FileRoute similar to an endpoint, it has:
 
 ```ts copy filename="app/api/uploadthing/core.ts"
 import { createUploadthing, type FileRouter } from "uploadthing/next";
+import { UploadThingError } from "uploadthing/server";
 
 const f = createUploadthing();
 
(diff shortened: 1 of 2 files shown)


markflorkowski (approved):
Good catch

PR #603, opened 2024-02-07
fix: resolve source types so we don't need the built packages locally

### LINKED ISSUE #591: [misc]: use project references to make types work without building dependent packages
### I understand that this issue may be closed if it should be filed in another category

- [X] I understand

### My issue

for dev we should be able to resolve the `@uploadthing/shared` types directly from source instead of having to have a built dist-folder to grab the types from

juliusmarminge:
closes #591

Allow stuff to be instantly resolved when making changes locally, without needing to build:

https://github.com/pingdotgg/uploadthing/assets/51714798/3496ffa8-ab64-489f-bba5-0c4acd0340ce

### TODO

- [x] Fix so that `pnpm build` uses the `tsconfig.build.json` so that builds still use the package name and not the relative path.
    - [ ] maybe upstream the patch i did to allow this 
     CleanShot 2024-02-07 at 09 55 06@2x

### DIFF
--- a/packages/shared/src/utils.ts
+++ b/packages/shared/src/utils.ts
@@ -247,3 +247,5 @@ export function objectKeys<T extends Record<string, unknown>>(
 export function isObject(obj: unknown): obj is Record<string, unknown> {
   return typeof obj === "object" && obj !== null && !Array.isArray(obj);
 }
+
+export const A = "a";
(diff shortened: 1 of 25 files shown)


packages/shared/src/utils.ts
```diff
@@ -247,3 +247,5 @@ export function objectKeys<T extends Record<string, unknown>>(
 export function isObject(obj: unknown): obj is Record<string, unknown> {
   return typeof obj === "object" && obj !== null && !Array.isArray(obj);
 }
+
+export const A = "a";
```
markflorkowski, line 251: ?

juliusmarminge: Oh that wasnt meant to be committed

juliusmarminge:
merging as i want it for the effect session tmrw :)

PR #605, opened 2024-02-07
docs: Update `react-dropzone` section

ImBaedin:
The docs didn't mention anything about these missing APIs, and had me scratching my head a bit.

I think I listed them all, might be better to list them in a default-closed accordion or something.


markflorkowski:
Yeah I think this is probably best collapsed by default. I believe that there is an Accordion component in the `components` directory you might be able to use

markflorkowski (approved):

PR #619, opened 2024-02-16
feat: add version lock to ensure `uploadthing/client` and `uploadthing/server`uses same version

juliusmarminge:
technically wouldn't need to be exact match, but why not?

### TODO
- [ ] maybe add version check (non-exact) between `@uploadthing/react` and `uploadthing/client` ?


markflorkowski (approved):

markflorkowski:
Currently the versions don't track each other though?

PR #622, opened 2024-02-19
fix: `appendOnPaste` and `mode=auto` not working together

juliusmarminge:


packages/react/src/components/dropzone.tsx
```diff
@@ -149,13 +149,17 @@ export function UploadDropzone<
       if (document.activeElement !== rootRef.current) return;
 
       const pastedFiles = getFilesFromClipboardEvent(event);
-      if (!pastedFiles) return;
+      if (!pastedFiles?.length) return;
 
-      setFiles((prev) => [...prev, ...pastedFiles]);
+      let filesToUpload = pastedFiles;
+      setFiles((prev) => {
+        filesToUpload = [...prev, ...pastedFiles];
```
juliusmarminge, line 156: idk if this is cursed or good...

packages/react/src/components/dropzone.tsx
```diff
@@ -149,13 +149,17 @@ export function UploadDropzone<
       if (document.activeElement !== rootRef.current) return;
 
       const pastedFiles = getFilesFromClipboardEvent(event);
-      if (!pastedFiles) return;
+      if (!pastedFiles?.length) return;
```
markflorkowski, line 152: Should we consider an [eslint plugin](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/HEAD/docs/rules/explicit-length-check.md) to enforce this?

packages/react/src/components/dropzone.tsx
markflorkowski: How is this different from before....? 😅

markflorkowski (approved):

packages/react/src/components/dropzone.tsx
juliusmarminge: State update cycle

markflorkowski: 🥲

PR #627, opened 2024-02-20
refactor: drop usage of `util.inspect`

### LINKED ISSUE #626: refactor: make logger not use `require`
Should find a way to get rid of this require: https://github.com/pingdotgg/uploadthing/blob/14d190d9d9099b9cb5bf938808f9dc8e6a886e82/packages/uploadthing/src/internal/logger.ts#L79

Maybe using https://github.com/browserify/node-util/blob/master/README.md, or an exports condition to a "fake" entrypoint that just aliases JSON.stringify ?

juliusmarminge:
Closes #626


packages/uploadthing/src/internal/logger.ts
```diff
@@ -73,15 +73,7 @@ function formatArgs(args: any[]) {
...
-    try {
-      // prefer inspect over JSON.stringify because it handles circular references, prints classes etc
-      // eslint-disable-next-line @typescript-eslint/no-var-requires, @typescript-eslint/consistent-type-imports
-      const util = require("util") as typeof import("util");
-      return util.inspect(arg, { depth: 4 });
-    } catch {
-      // fallback to JSON.stringify if inspect fails e.g. if runtime doesn't have util module
-      return JSON.stringify(arg, null, 4);
-    }
+    return JSON.stringify(arg, null, 4);
```
juliusmarminge, line 76: downside is if we attempt to log classes `Response` it will just output an empty object, but we don't do that anywhere today and according to my tests, logs looks good with JSON.stringify:

CleanShot 2024-03-01 at 21 21 12@2x

markflorkowski (approved):

PR #630, opened 2024-02-23
chore: enforce compatible `uploadthing` version for `@uploadthing/react`

juliusmarminge:
Followup #619 

Will still require us to bump the peerDependency whenever necessary but should be enforceable as long as we do

Currently just logs it out:

CleanShot 2024-02-23 at 15 12 11@2x

But could maybe throw?


packages/react/src/useUploadThing.ts
```diff
@@ -42,6 +44,17 @@ export const INTERNAL_uploadthingHookGen = <
    */
   url: URL;
 }) => {
+  console.log(
```
markflorkowski, line 47: remove?

markflorkowski (approved):

PR #636, opened 2024-02-26
fix: report failed PSPS uploads as failed to infra

juliusmarminge:
failed uploads using PSP were not reported to infra so they would still count towards quota for the user's app even though they didn't occupy space in s3

### DIFF
--- a/packages/uploadthing/src/client.ts
+++ b/packages/uploadthing/src/client.ts
@@ -331,22 +331,20 @@ async function uploadPresignedPost(
   Object.entries(presigned.fields).forEach(([k, v]) => formData.append(k, v));
   formData.append("file", file); // File data **MUST GO LAST**
 
-  const response: XMLHttpRequest = await new Promise<XMLHttpRequest>(
-    (resolve, reject) => {
-      const xhr = new XMLHttpRequest();
-      xhr.open("POST", presigned.url);
-      xhr.setRequestHeader("Accept", "application/xml");
-      xhr.upload.onprogress = (p) => {
-        opts.onUploadProgress?.({
-          file: file.name,
-          progress: (p.loaded / p.total) * 100,
-        });
-      };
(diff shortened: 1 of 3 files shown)


packages/uploadthing/src/client.ts
```diff
@@ -327,33 +327,45 @@ async function uploadPresignedPost(
...
-      opts.onUploadProgress?.({
-        file: file.name,
-        progress: (p.loaded / p.total) * 100,
-      });
-    };
-    xhr.onload = (e) => resolve(e.target as XMLHttpRequest);
-    xhr.onerror = (e) => reject(e);
-    xhr.send(formData);
-  });
+  let response: XMLHttpRequest;
```
markflorkowski, line 338: Any reason why you need let + try-catch instead of `.catch()`?

```
const response: XMLHttpRequest = await new Promise<XMLHttpRequest>(
    (resolve, reject) => {
      const xhr = new XMLHttpRequest();
      xhr.open("POST", presigned.url);
      xhr.setRequestHeader("Accept", "application/xml");
      xhr.upload.onprogress = (p) => {
        opts.onUploadProgress?.({
          file: file.name,
          progress: (p.loaded / p.total) * 100,
        });
      };
      xhr.onload = (e) => resolve(e.target as XMLHttpRequest);
      xhr.onerror = (e) => reject(e);
      xhr.send(formData);
    },
  ).catch(async (error) => {
    await opts.reportEventToUT("failure", {
      fileKey: presigned.key,
      uploadId: null,
      fileName: file.name,
      s3Error: (error as Error).toString(),
    });
    throw "unreachable"; // failure event will throw for us
  });
```

markflorkowski (approved):

packages/uploadthing/src/client.ts
juliusmarminge: Probably not - getting resrutten in effect anyway so didnt worry ablut styling too much 😅

PR #638, opened 2024-02-26
fix: reject data urls in uploadFileFromUrl()

markflorkowski:
- **reject data urls**
- **add test**


juliusmarminge (approved):

packages/uploadthing/src/sdk/index.ts
```diff
@@ -219,7 +219,19 @@ export class UTApi {
...
+
+        if (typeof url === "string") {
+          // since dataurls will result in name being too long, tell the user
+          // to use uploadFiles instead.
+          if (url.startsWith("data:")) {
+            throw new UploadThingError({
+              code: "BAD_REQUEST",
+              message:
+                "Please use uploadFiles() for data URLs. uploadFilesFromUrl() is intended for use with remote URLs only.",
+            });
```
juliusmarminge, line 231: wait should we throw? the inner function returns either data or error and let's the other continue

markflorkowski: We throw below if ant of the files fail to download. Possibly we should not throw in either case then...

juliusmarminge: right... we should probably just make sure to add an error for that place in the final result array?

packages/uploadthing/src/sdk/index.ts
```diff
@@ -256,7 +287,9 @@ export class UTApi {
       },
     );
 
-    const uploadFileResponse = Array.isArray(urls) ? uploads : uploads[0];
+    const uploadFileResponse = Array.isArray(urls)
+      ? [...errors, ...uploads]
+      : [...errors, ...uploads][0];
```
juliusmarminge, line 292: Would be nice if order was preserved

packages/uploadthing/src/sdk/sdk.test.ts
```diff
@@ -135,6 +135,36 @@ describe("uploadFilesFromUrl", () => {
...
+        return Promise.resolve(Response.json([]));
+      },
+    });
+
+    const result = await utapi.uploadFilesFromUrl(
+      "data:text/plain;base64,SGVsbG8sIFdvcmxkIQ==",
+    );
+    expect(result).toMatchInlineSnapshot(
+      `[Error: Please use uploadFiles() for data URLs. uploadFilesFromUrl() is intended for use with remote URLs only.]`,
+    );
```
juliusmarminge, line 166: `{ data: null, error }`?

packages/uploadthing/src/sdk/index.ts
markflorkowski: Would require infra changes, and I'm not sure how the UploadThingError being passed over the wire would go

juliusmarminge: Would it? We'd just need to keep track of what indexes failed?

packages/uploadthing/src/sdk/index.ts
```diff
@@ -228,24 +243,25 @@ export class UTApi {
...
+          return undefined;
         }
         logger.debug("Finished downloading file. Reading blob...");
         const blob = await fileResponse.blob();
         logger.debug("Finished reading blob.");
-        return Object.assign(blob, { name: filename });
+        return new UTFile([blob], filename);
       }),
-    );
+    ).then((files) => files.filter((x): x is UTFile => x !== undefined));
```
markflorkowski, line 258:
```suggestion
    ).then((files) => files.filter(Boolean);
```

?

packages/uploadthing/test/sdk.test.ts
```diff
@@ -212,6 +212,61 @@ describe("uploadFilesFromUrl", () => {
       },
     );
   });
+
+  // if passed data url, array contains UploadThingError
+  it("return is error if data url is passed", async () => {
```
markflorkowski, line 217:
```suggestion
  it("returns error if data url is passed", async () => {
```

PR #642, opened 2024-02-28
feat(client): add `skipPolling` option

juliusmarminge:
Adds a `skipPolling` option to the client to allow skipping polling for server data before firing the `onClientUploadComplete`. This greatly reduces the duration of the spinner and can be useful for all those that doesn't utilize server data at all.

Almost 90% don't use this feature so this should be an improvement for a majority of users:
CleanShot 2024-02-28 at 17 17 08@2x

Perhaps a better default would be `true`, but that'd be a breaking change and require a major bump. Something to keep in mind for v7 maybe (or should the field be called `pollForServerData` instead, with `true` as default and we can change to `false` in next major 🤔)

### Demo

https://github.com/pingdotgg/uploadthing/assets/51714798/c648ebe5-cd13-4eb6-a31b-2475c1895c00

### DIFF
--- a/.changeset/tasty-icons-laugh.md
+++ b/.changeset/tasty-icons-laugh.md
@@ -1,0 +1,7 @@
+---
+"docs": patch
+"@uploadthing/react": patch
+"uploadthing": patch
+---
+
+feat(client): add `skipPolling` option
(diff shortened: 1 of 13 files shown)


markflorkowski (approved):
Should probably document this behaviour. Functionally looks good though.

markflorkowski (approved):

Issue #653, opened 2024-02-29
feat: Have the ability to pass a input prop to overwrite some of the condition


Semkoo:
### Describe the feature you'd like to request

First, this is an amazing and simple-to-use tool and I appreciate it as it helped jump-start start project that I am working on. 

But, I ran into a little problem. I only wanted to upload one file at a time, without letting people choose lots of files simultaneously by holding down the 'shift' key. So, I had to do a little trick to make it work the way I wanted.

Here's a simple example of what I did:

``` tsx
"use client";

import { useEffect } from "react";
import { UploadDropzone } from "~/utils/uploadthing";

// CustomUploadThing component with special settings
type CustomUploadThing = React.ComponentProps<typeof UploadDropzone> & {
  // Adding a custom option for one file at a time
  isMultiple?: boolean;
};

// CustomUploadThing function component
export function CustomUploadThing(props: CustomUploadThing) {
  // Using useEffect to update things when props change
  useEffect(() => {
    // Finding the input element
    const input = document.querySelector("#drop-zone input");

    // If we only want one file at a time...
    if (!props.isMultiple) {
      // Remove the 'multiple' option from the input
      input?.removeAttribute("multiple");
      // We could also change the button text here if needed
    }
  }, [props.isMultiple]);

  // Rendering our UploadDropzone component inside a div
  return (
    <div id="drop-zone">
      <UploadDropzone {...props} />
    </div>
  );
}
``` 

### Describe the solution you'd like to see

So I was hopping if there is the way we pass an inputProp into UploadDropzone so for example  

```tsx  
.... 
<UploadDropzone 
   ... 
   inputProps = {
      ....
     multiple: false 
    }
    />
....
``` 

### 👨‍👧‍👦 Contributing

- [X] 🙋‍♂️ Yes, I'd be down to file a PR implementing this feature!

markflorkowski:
Hmm. There is also a bug here. it the endpoint only allows uploading a single file, the dropzone should not have the multiple prop on the input. The upload button component does this correctly.

markflorkowski:
After thinking on this some more, I don't think that it makes sense to allow the override on the components. We want to promote configuring the accepted input (file types, size, quantity) on the file router, and the components will apply the equivalent restrictions "automatically".

I have a PR up now that will fix the behavior so that the filerouter config is respected by the components. 

[...]

PR #654, opened 2024-02-29
fix(dropzone): Proper handling of endpoints that only accept 1 file

markflorkowski:
Address #653


juliusmarminge (approved):

examples/minimal-solidstart/tailwind.config.ts
```diff
@@ -1,5 +1,5 @@
 import { withUt } from "uploadthing/tw";
 
 export default withUt({
-  content: ["src/**/*.tsx"],
+  content: ["src/**/*.tsx", "../../node_modules/@uploadthing/solid/**/*.tsx"],
```
juliusmarminge, line 4: bun doesn't put node_modules in each package so the `./node_modules/...` content path the plugin sets doesn't work here.

we should document something like

```md
While `withUT` works out of the box for many cases, you might have to add the path to the uploadthing package yourself ...
```

juliusmarminge: alternatively we could just add some additional paths in the plugin, but that could come at the cost of some performance if we're including more paths than necessary (although i'd presume that would be marginally).

I'm thinking something like

```ts
// current
const contentPaths = [
  "./node_modules/@uploadthing/react/dist/**",
  "./node_modules/@uploadthing/solid/dist/**",
];

// new
const packages = ["react", "solid"];
const paths = packages.flatMap((pkg) => [
  `./node_modules/@uploadthing/${pkg}/dist/**`,
  `../node_modules/@uploadthing/${pkg}/dist/**`,
  `../../node_modules/@uploadthing/${pkg}/dist/**`,
]);
```

although some monorepos could be deeper, I think 2 levels are the most common and should cover 99% of cases

markflorkowski: I think that the cost of extra paths would be exceedingly small, so not too concerned with that route. 

`import.meta.resolve` also seems fine, worst case it doesn't work, and we fall back on the same error?
> While `withUT` works out of the box for many cases, you might have to add the path to the uploadthing package yourself ...

PR #662, opened 2024-03-03
chore: extend types

juliusmarminge:
Another follow up on #659

### DIFF
--- a/packages/uploadthing/src/client.ts
+++ b/packages/uploadthing/src/client.ts
@@ -195,18 +195,20 @@ async function uploadMultipart(
   }
 
   // Tell the server that the upload is complete
-  const uploadOk = await opts.reportEventToUT("multipart-complete", {
-    uploadId: presigned.uploadId,
-    fileKey: presigned.key,
-    etags,
-  });
-  if (!uploadOk) {
-    console.log("Failed to alert UT of upload completion");
-    throw new UploadThingError({
-      code: "UPLOAD_FAILED",
-      message: "Failed to alert UT of upload completion",
+  await opts
+    .reportEventToUT("multipart-complete", {
+      uploadId: presigned.uploadId,
+      fileKey: presigned.key,
+      etags,
+    })
+    .catch((res) => {
+      console.log("Failed to alert UT of upload completion");
+      throw new UploadThingError({
+        code: "UPLOAD_FAILED",
+        message: "Failed to alert UT of upload completion",
+        cause: res,
+      });
     });
-  }
 }
 
 async function uploadPresignedPost(
--- a/packages/uploadthing/src/client.ts
+++ b/packages/uploadthing/src/client.ts
@@ -3,22 +3,23 @@
 
 import {
   resolveMaybeUrlArg,
-  safeParseJSON,
   UploadThingError,
   withExponentialBackoff,
 } from "@uploadthing/shared";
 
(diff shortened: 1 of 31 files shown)


packages/uploadthing/src/internal/handler.ts
```diff
@@ -309,10 +309,11 @@ export const buildRequestHandler = <
             (f) =>
               isObject(f) &&
               typeof f.name === "string" &&
-              typeof f.size === "number",
+              typeof f.size === "number" &&
+              typeof f.type === "string",
```
juliusmarminge, line 313: idk if we should send this from client or use lookup using filename, but i think it should be included in the `opts.files` you get in `.middleware()`.

packages/uploadthing/src/types.ts
```diff
@@ -87,26 +71,39 @@ export type GenerateUploaderOptions = {
...
-export type UploadThingResponse = (PSPResponse | MPUResponse)[];
+/**
+ * When the client has uploaded a file and polled for data returned by `.onUploadComplete()`
+ */
+export interface ClientUploadedFileData<T> extends UploadedFileData {
+  /**
+   * Matches what's returned from the serverside `onUploadComplete` callback
+   */
+  serverData: T;
+}
```
juliusmarminge, line 109: this is the main change i wanted to make

We start with what the client sends (FileUploadData), then once middleware has run, we may have customId (FileUploadDataWithCustomId), then when the file has uplaoded we get `key` and `url` from UT (UploadedFileData). Lastly, the client polls for server data (ClientUploadedFileData)

packages/uploadthing/src/client.ts
```diff
@@ -144,15 +116,16 @@ const uploadFilesInternal = async <
...
       name: file.name,
       size: file.size,
+      type: file.type,
       key: presigned.key,
-
-      serverData,
       url: "https://utfs.io/f/" + presigned.key,
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+      serverData: serverData as any,
+      customId: null,
```
juliusmarminge, line 124: should this be filled in?

packages/uploadthing/src/internal/dev-hook.ts
```diff
@@ -28,7 +28,16 @@ export const conditionalDevServer = async (opts: {
...
+    async (json: {
+      fileData: {
+        callbackUrl: string;
+        callbackSlug: string;
+        metadata: string | null;
+        fileName: string;
+        fileSize: number;
+        customId: string | null;
+      };
+    }) => {
```
juliusmarminge, line 40: decided to inline this - not used anywhere else and very specific to the callback simulation.

packages/uploadthing/src/internal/handler.ts
```diff
@@ -474,16 +475,13 @@ export const buildRequestHandler = <
 
         return {
           cleanup: promise,
-          body: parsedResponse.map((x) => ({
-            ...x,
-            pollingUrl: generateUploadThingURL(`/api/serverCallback`),
-          })),
+          body: parsedResponse satisfies UTEvents["upload"]["out"],
```
juliusmarminge, line 478: we handle this mapping on infra, no need for mapping
CleanShot 2024-03-03 at 19 53 08@2x

markflorkowski (approved):

PR #670, opened 2024-03-05
fix(docs): prevent escape from throwing error when no toast is open

benox3:
Pressing escape on api reference page causes an error due to some weirdness with `vueaury` + `vue-sonner`. We just patch the package to add an additional type check for `vue-sonner` for whether the contains function is available.

Resolves:
```Uncaught TypeError: r.contains is not a function``` on openapi spec page


juliusmarminge:
https://docs.uploadthing.com/api-reference/openapi-spec tried pressing escape in a bunch of ways but didn't get any error. Can you record the issue you're seeing?

markflorkowski:
> https://docs.uploadthing.com/api-reference/openapi-spec tried pressing escape in a bunch of ways but didn't get any error. Can you record the issue you're seeing?


https://github.com/pingdotgg/uploadthing/assets/11494384/740eef77-746f-498a-bf55-a5b0b0a4863d

benox3:
Thanks @markflorkowski -- that's the one. 

Seems like the preview still has the issue. I'm wondering if the vercel deploy is using cache, otherwise it could be something off with the patching which was kind of finicky to implement with bun.

markflorkowski:
> Seems like the preview still has the issue. I'm wondering if the vercel deploy is using cache, otherwise it could be something off with the patching which was kind of finicky to implement with bun.

Being on the docs site, I would not be surprised if this were the case. We have had a ton of build cache issues with Nextra :/

PR #673, opened 2024-03-06
test: add client tests

juliusmarminge:
more stuff to increase confident in Effect PR. we're going places

CleanShot 2024-03-06 at 20 18 35


patches/msw+2.2.2.patch
juliusmarminge, line 1: msw doesn't really like that we don't use global fetch but pass it as args... so this makes it so we only intercept XHR and mock the other requests by passing the `mockExternalRequests` function as the fetch option...

packages/uploadthing/test/sdk.test.ts
```diff
@@ -52,7 +52,7 @@ describe("uploadFiles", () => {
     );
     expect(fetchMock).toHaveBeenNthCalledWith(
       2,
-      "https://bucket.s3.amazonaws.com/abc-123.txt",
+      "https://bucket.s3.amazonaws.com",
```
juliusmarminge, line 55: post urls sends the key in body

packages/uploadthing/test/__test-helpers.ts
```diff
@@ -1,17 +1,30 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
markflorkowski, line 1: only needed for line 129?

juliusmarminge: fixed

packages/uploadthing/test/client.test.ts
```diff
@@ -0,0 +1,215 @@
...
+        url: "https://utfs.io/f/abc-123.txt",
+      },
+    ]);
+
+    expect(s3Mock).toHaveBeenCalledOnce();
+    expect(s3Mock).toHaveBeenCalledWith({}, expect.any(Request));
+
+    expect(middlewareMock).toHaveBeenCalledOnce();
+    expect(onErrorMock).not.toHaveBeenCalled();
+    // expect(uploadCompleteMock).toHaveBeenCalledOnce();
```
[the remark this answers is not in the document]
juliusmarminge: Is there some good way to store context from the test? So we can simulate the UTApi even better and store these when uploading, and have pollupload read from that? Have to be isolated to a single test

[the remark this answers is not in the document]
juliusmarminge: took a while, but got it working

PR #679, opened 2024-03-07
chore: show msw bug

juliusmarminge:
msw doesn't intercept this call, if it did we would've got the mocked response instead of the `Invalid API Key` error which is coming from the UploadThing API

CleanShot 2024-03-07 at 14 14 05

CleanShot 2024-03-07 at 14 08 19

---

to test:

In root: `bun i`

In root: `bun vitest sdk`


juliusmarminge:
Oh right and this is the other issue, it's not compatible with HappyDOM, that's why I had to do the patch and make it not mock fetch requests in the first place..

CleanShot 2024-03-07 at 14 11 21

PR #680, opened 2024-03-07
refactor(docs): remove unnecessary useEffect for openapi spec

benox3:
We can remove the `useEffect` and just hide the TOC by leveraging `layout: "full"`.

Reference: https://nextra.site/docs/docs-theme/page-configuration#full-layout

### DIFF
--- a/docs/src/components/scalar.tsx
+++ b/docs/src/components/scalar.tsx
@@ -1,4 +1,3 @@
-import { useEffect } from "react";
 import { ApiReference as VueComponent } from "@scalar/api-reference";
 import { useTheme } from "nextra-theme-docs";
 import { applyVueInReact } from "veaury";
@@ -10,14 +9,6 @@ const specUrl = "https://uploadthing.com/api/openapi-spec.json";
 export function ScalarApiRef() {
   const theme = useTheme();
   const isDark = theme.resolvedTheme === "dark";
-
-  useEffect(() => {
-    const toc = document.querySelector('nav[aria-label="table of contents"]');
-    toc?.classList.add("hidden");
-    return () => {
-      toc?.classList.remove("hidden");
-    };
-  }, []);
 
   return (
     // @ts-ignore
--- a/docs/src/pages/api-reference/_meta.json
+++ b/docs/src/pages/api-reference/_meta.json
@@ -1,6 +1,11 @@
 {
   "react": "React",
   "server": "Server",
-  "openapi-spec": "OpenAPI Spec",
+  "openapi-spec": {
+    "title": "OpenAPI Spec",
+    "theme": {
+      "toc": false
+    }
+  },
   "ut-api": "UTApi (Server SDK)"
 }


docs/src/globals.css
```diff
@@ -79,6 +79,7 @@
  */
 .references-layout {
   height: max-content;
+  width: 100% !important;
```
markflorkowski, line 82: Linked doc did not mention this. Is it required, or was this accidentally left in from testing?

benox3: > Linked doc did not mention this. Is it required, or was this accidentally left in from testing?

@markflorkowski good catch. So what I didn't realize that the "full" layout strips away the container, breadcrumbs, etc... which is why the width was necessary to override the 100dvw.

I've made a change to remove just the `toc` via the configuration, based on their design decisions they mark the element as hidden so it would still retain the space.

Before: 
image

After:
image

markflorkowski:
Something is not quite right now 
image

benox3:
> @benox3 Something is not quite right now image

I think that's vercel cache. Try doing a reploy without existing cache. -- Also I rebased onto the latest. Seems like the changes to the spec are silently causing scalar to not render.

markflorkowski:
> Seems like the changes to the spec are silently causing scalar to not render.

Yeah trying to debug this one now.

benox3:
> > Seems like the changes to the spec are silently causing scalar to not render.
> 
> Yeah trying to debug this one now.

Looks like malformed schema. 
If you pop it in there you can see the issue on 
https://editor-next.swagger.io/. 

Some  zod defs made it into the swagger spec on accident:  
```"schema": {
                                    "data": {
                                        "_def": {
                                            "type": {
                                                "_def": {
                                                    "options": [
                                                    ...
                                                    
```

markflorkowski:
> Looks like malformed schema. If you pop it in there you can see the issue on https://editor-next.swagger.io/.

weird thing is that the playground works, both in swagger, and in Scalar's own demo.

Should still fix the invalid parts of the schema, but still...

benox3:
Looks like some Zod schema validation stuff made it's way into the spec

> > Looks like malformed schema. If you pop it in there you can see the issue on https://editor-next.swagger.io/.
> 
> weird thing is that the playground works, both in swagger, and in Scalar's own demo.
> 
> Should still fix the invalid parts of the schema, but still...

Yea my guess is they run it through a parser and then serve it as content, and the parser is graceful. Whereas in this case it's probably just thrown in as raw content.

markflorkowski:
Weird that they wouldn't use their own component in their example though.

markflorkowski:
Schema should be fixed now, but it looks like scalar is still silently failing

markflorkowski:
Spoke with the scalar team, and they are investigating what is going on. Temporarily pointed at the old generated schema so the page would not be blank.

benox3:
> Spoke with the scalar team, and they are investigating what is going on. Temporarily pointed at the old generated schema so the page would not be blank.

Looks like it's working on live @markflorkowski

markflorkowski:
Deploying what I think should resolve the messed up refs now

markflorkowski:
Should be fixed! I also bumped the version of scalar, for good measure. 

They're working on some additional fixes to make sure errors aren't lost in the SSR pass, but at least for now, it should all be good!

markflorkowski:
And the version bump broke something... 😅

markflorkowski:
Superseded by #707

PR #700, opened 2024-03-10
fix(docs): add noindex option & canonical url

RhysSullivan:
Fixes https://x.com/t3dotgg/status/1766445250527351012?s=20

Noindex:

While there are ways to auto detect this, i.e off of the Vercel env variable. But given that noindex is more of a destructive action imo it should be manually set. 

In the dev & preview branches, set NEXT_PUBLIC_DISABLE_INDEXING to true in the env vars & validate

Canonical Url:

Uses https://docs.uploadthing.com as the default, adds on the path name. Should also help with deduplication & and forks that don't have the environment variable set.
Screenshot 2024-03-10 at 1 43 37 AM

### DIFF
--- a/docs/src/theme.config.js
+++ b/docs/src/theme.config.js
@@ -163,6 +163,7 @@ const config = {
           { url: "https://docs.uploadthing.com/og.jpg?random=aaaaaaaaaaaaa" },
         ],
       },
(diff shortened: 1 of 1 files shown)


markflorkowski:
image

markflorkowski (approved):

PR #701, opened 2024-03-10
feat: allow fetch override on client

juliusmarminge:
quickly spun up the ability to allow fetch override to set headers. Ref: https://discord.com/channels/966627436387266600/1102510616326967306/1216240077605310576

### Questions

- do we need full fetch override? or is `headers` enough? or both?
- currently only possible to set on the `genUploader`, `generateReactHelpers`, `generateUploadButton` etc etc. Maybe we should also have a way to allow this override when doing the call / on the components? Thinking if the token requires React context?

```ts
const { token } = useUser();
const { startUpload } = useUploadThing({
  headers: {
    Authorization: `Bearer ${token}`
  }
})
```
- is it even useful at all to have it on the generators? Maybe it should just be on individual calls

### DIFF
--- a/packages/react/src/components/button.tsx
+++ b/packages/react/src/components/button.tsx
@@ -101,6 +101,7 @@ export function UploadButton<
   const { mode = "auto", appendOnPaste = false } = $props.config ?? {};
 
   const useUploadThing = INTERNAL_uploadthingHookGen<TRouter>({
+    fetch: $props.fetch ?? globalThis.fetch.bind(globalThis),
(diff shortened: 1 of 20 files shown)


markflorkowski:
> do we need full fetch override? or is headers enough? or both?

I can't really think of reasons why you would actually need a full fetch override, headers only is probably fine. IF other things are needed in the future we can reassess.

> is it even useful at all to have it on the generators?

Probably not. I think having it on the hook/components is probably more flexible for the user.

juliusmarminge:
Superseded by #705

PR #705, opened 2024-03-13
feat: add headers override option

juliusmarminge:


markflorkowski (approved):
Lgtm, for components should headers be in config object rather than an additional key?

PR #711, opened 2024-03-15
test: describe network using only msw

juliusmarminge:
trying to get rid of our `mockExternalRequest(db)` in favor of only using msw to describe all the external requests, no matter if the method accepts a fetch override or not.

it's going.... well not so good 😓


patches/@mswjs+interceptors+0.25.16.patch
juliusmarminge, line 1: this here fixes some incompatabilities with using `happy-dom`. Somewhere some serialization is failing which results in a `new URL("[object Request]")` which fails. this patch seems to fix it

packages/uploadthing/test/__test-helpers.ts
```diff
@@ -36,15 +38,17 @@ export const it = itBase.extend<{ db: MockDbInterface }>({
 });
 
 export const s3Mock = vi.fn();
-export const fetchMock = vi.fn();
+export const staticAssetMock = vi.fn();
+export const utApiMock = vi.fn();
```
juliusmarminge, line 42: i also tried making these `vi.spyOn(globalThis, "fetch")` but that didn't pick up half the calls...

patches/@mswjs+interceptors+0.25.16.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/@mswjs/interceptors/lib/node/interceptors/fetch/index.mjs b/node_modules/@mswjs/interceptors/lib/node/interceptors/fetch/index.mjs
+index 717678e..6790eb0 100644
+--- a/node_modules/@mswjs/interceptors/lib/node/interceptors/fetch/index.mjs
++++ b/node_modules/@mswjs/interceptors/lib/node/interceptors/fetch/index.mjs
+@@ -122,7 +122,7 @@ var _FetchInterceptor = class extends Interceptor {
+         return mockedResponse;
+       }
+       this.logger.info("no mocked response received!");
+-      return pureFetch(request).then((response) => {
++      return pureFetch(resolvedInput, init).then((response) => {
```
kettanaito, line 10: Interesting. The `request` is created by taking `init` into account:

https://github.com/mswjs/interceptors/blob/b1bc5a5adf31726996d0935aa4b178eade1f9ac5/src/interceptors/fetch/index.ts#L50

Can you tell me more about how you reproduce this issue? Would love to make a regression test out of it.

packages/uploadthing/test/__test-helpers.ts
```diff
@@ -94,111 +98,269 @@ const mockPresigned = (file: {
...
+export const setupMsw = (opts: {
+  db: MockDbInterface;
+  failS3Call?: boolean;
+}) => {
+  const server = setupServer(
+    ...s3Mocks(opts.failS3Call ?? false),
+    ...staticAssetServer,
+    ...utapiMocks(opts.db),
+  );
+  server.listen({ onUnhandledRequest: "warn" });
```
kettanaito, line 306: I recommend you move this call outside, declare the `server` once, then use the Vitest context to expose the `server` reference in each test. 

Use the [`server.use()`](https://mswjs.io/docs/api/setup-server/use) API to prepend any request handlers based on the `db` in individual test, if any. 

So the end usage looks something like this:

```ts
// __test-helpers.ts
const server = setupServer(...anyHappyPaths)

export const useDb(db) {
  server.use(...utapiMocks(db),
  // ...any other overrides.
}
```
```ts
// some.test.ts
test('something', ({ useDb }) => {
  useDb(someDb)
})
```

You can also drop "msw" from your context function. That's really an implementation detail. Name that function for what it stands for. Here, I assume it's about using a particular state of the database in an individual test.

kettanaito, line 306: If you plan on running tests concurrently, I highly recommend using the [`server.boundary()` API](https://mswjs.io/docs/api/setup-server/boundary) to scope each `server.use()` modifications to the individual test's scope.

juliusmarminge: okay i think i got something working!

juliusmarminge: good to know, thanks!

patches/@mswjs+interceptors+0.25.16.patch
juliusmarminge: Something with happy-dom when it's creating the url object for the request. seems like something doesn't get serialized as i debugged it was calling the URL constructor with the _string_ `[object Request]` resulting in a `Invalid URL` error

juliusmarminge:
UPDATE: I think I got a quite clean solution now.

Tests consuming the `db` context will now prepend UTApiMocks to the msw listeners with that db instance. No need to provide any fetch overrides etc etc, just destruct the db from the context and all should work automatically

packages/uploadthing/test/__test-helpers.ts
```diff
@@ -94,111 +77,233 @@ const mockPresigned = (file: {
...
+    });
+    return HttpResponse.json(null, { status: 200 });
+  }),
+  http.put("https://bucket.s3.amazonaws.com/:key", async ({ request }) => {
+    s3Mock(request.url, {
+      method: request.method,
+      headers: Object.fromEntries(request.headers.entries()),
+      body: await request.blob(),
+    });
+    return HttpResponse.json(null, {
```
kettanaito, line 98: If the response body is `null`, you are not gaining any benefit of using `HttpResponse.json()` (which is automatic `Content-Type` and `Content-Length` headers assignment). 

I propose you use `new HttpResponse` here:

```suggestion
    return new HttpResponse(null, {
```

packages/uploadthing/test/__test-helpers.ts
```diff
@@ -94,111 +77,233 @@ const mockPresigned = (file: {
...
+  /**
+   * S3
+   */
+  http.post("https://bucket.s3.amazonaws.com", async ({ request }) => {
+    s3Mock(request.url, {
+      method: request.method,
+      headers: Object.fromEntries(request.headers.entries()),
+      body: await request.formData(),
+    });
+    return HttpResponse.json(null, { status: 200 });
```
kettanaito, line 90: Here's the same functionality with less code:

```suggestion
    return new HttpResponse()
```

packages/uploadthing/test/__test-helpers.ts
```diff
@@ -94,111 +77,233 @@ const mockPresigned = (file: {
...
-      })();
-
-      if (url.pathname === "/api/uploadFiles") {
-        const presigneds = (body?.files as any[]).map((file) => {
+export const msw = setupServer(
+  /**
+   * S3
+   */
+  http.post("https://bucket.s3.amazonaws.com", async ({ request }) => {
+    s3Mock(request.url, {
```
kettanaito, line 85: The `s3Mock()` function seems to be accepting something _very_ similar to `input, init` of a `Request` constructor. 

I suggest you align the call signature to allow passing the `request` clone here instead:

```suggestion
    s3Mock(request.clone())
```

This will forward the entire request so you don't have to unwrap it, serialize its headers, or read its body. You can do all that in the `s3Mock()` function, if needed at all.

packages/uploadthing/test/__test-helpers.ts
```diff
@@ -94,111 +77,233 @@ const mockPresigned = (file: {
...
+    });
+  }),
+  /**
+   * Static Assets
+   */
+  http.get("https://cdn.foo.com/:fileKey", ({ request }) => {
+    staticAssetMock(request.url);
+
+    return HttpResponse.text("Lorem ipsum doler sit amet", {
+      headers: { "content-type": "text/plain" },
```
kettanaito, line 110: The `Content-Type` response header is automatically set to `text/plain` if using the `HttpResponse.text()` helper. It also sets the `Content-Length` header to produce a semantically correct response.

You can drop the `headers` here entirely.

packages/uploadthing/test/__test-helpers.ts
```diff
@@ -94,111 +77,233 @@ const mockPresigned = (file: {
...
+/**
+ * Prepend MSW listeners to mock the UploadThing API
+ * Provide the `db` instance to store data within the test
+ */
+export const useDb = (db: MockDbInterface) =>
+  msw.use(
+    http.post(
+      "https://uploadthing.com/api/prepareUpload",
+      async ({ request }) => {
+        const body = (await request.json()) as { files: any[] } & Record<
```
kettanaito, line 139: You can annotate the request body if you are reading it as json using the generic type argument to the `http.post()` handler:

```ts
http.post<{ files: File[] }>("https://uploadthing.com/api/prepareUpload", async ({ request }) => {
  const body = await request.json() 
  body // { files: File[] }
})
```

The advantage of this to keep your resolvers tidier.

packages/uploadthing/test/__test-helpers.ts
```diff
@@ -94,111 +77,233 @@ const mockPresigned = (file: {
...
+    ),
+    http.get(
+      "https://uploadthing.com/api/pollUpload/:key",
+      async ({ request }) => {
+        utApiMock(request.url, {
+          method: request.method,
+          headers: Object.fromEntries(request.headers.entries()),
+        });
+
+        const url = new URL(request.url);
```
kettanaito, line 221: Looks like you aren't doing anything with this `url`. A left-over code?

packages/uploadthing/test/__test-helpers.ts
```diff
@@ -94,111 +77,233 @@ const mockPresigned = (file: {
...
+
+export const useBadS3 = () =>
+  msw.use(
+    http.post("https://bucket.s3.amazonaws.com", async ({ request }) => {
+      s3Mock(request.url, {
+        method: request.method,
+        headers: Object.fromEntries(request.headers.entries()),
+        body: await request.formData(),
+      });
+      return HttpResponse.json(null, { status: 403 });
```
kettanaito, line 277:
```suggestion
      return new HttpResponse(null, { status: 403 });
```

kettanaito (commented):
Looks much better now! Great job, @juliusmarminge. I've left a few comments on how you can improve MSW usage even further. Give those a read.

packages/uploadthing/test/__test-helpers.ts
juliusmarminge: I really just want a way to easily assert what requests was made and what they were made with. 

I tried spy'ing on the global fetch but that didn't seem to pick up all the calls for some reason....

packages/uploadthing/test/__test-helpers.ts
juliusmarminge: I am reading the key from the dynamic url. But I think I can use `params` for that instead right?

packages/uploadthing/test/__test-helpers.ts
kettanaito: @juliusmarminge, have you tried [Life-cycle events](https://mswjs.io/docs/api/life-cycle-events)? They give you the request resolution events that you can spy on. If you don't mean to affect the network, you should use life-cycle events as those are readonly but still let you know what requests/responses happened. 

```ts
// Create a spy function that gets the intercepted request
// and serializes it for easier assertion in tests.
const requestSpy = vi.fn(async ({ request }) => ({
  method: request.method,
  headers: Object.fromEntries(request.headers.entries()),
  body: await request.formData()
}))

// Let MSW tell you whenever requests happen.
// You can also use "request:handled" to spy only
// on the requests you have request handlers for.
server.events.on('request:start', requestSpy)

// Cleaning up is important.
afterEach(() => vi.resetAllMocks)

it('some test', async () => {
  await renderYourCode()

  // Assert on the spy just like you do now. 
  expect(requestSpy).toHaveBeenCalledWith({ ... })
})
```

> I tried spy'ing on the global fetch but that didn't seem to pick up all the calls for some reason....

Not sure why this is happening. Doesn't look to be related to MSW.

packages/uploadthing/test/__test-helpers.ts
kettanaito: Yes. MSW already parses the request path parameters for you.

```ts
http.get('/:key', ({ params }) => params.key)
```

You can also provide type annotations for parameters:

```ts
http.get<{ key: string }>('/:key', resolver)
```

packages/uploadthing/test/__test-helpers.ts
juliusmarminge: Hmm your code doesn't seem to work as we're calling the spy with `Request` and can't assert on the serialized object. Tried something like this but looks like the life-cycle event listener can't be async:

CleanShot 2024-03-24 at 11 38 46

juliusmarminge: with your way we can assert with `toHaveReturnedWith` instead but that reads a bit unintuitevly imo

juliusmarminge: worked around it a bit, not the cleanest as it requires every handler to `await callRequestSpy` but it works.

kettanaito: I'm quite sure mocking the implementation will work but you are right, you'd be asserting that the function _returned_ an object, wasn't called with it.

PR #717, opened 2024-03-19
docs: allow deployments from forks

juliusmarminge:
since adding the custom env the docs are no longer auto-deployed from forks since it requires reading environment variables from vercel.

this approach uses a system-env instead so we can remove the env from vercel and thus allowing forks to auto-deploy

### DIFF
--- a/docs/next.config.mjs
+++ b/docs/next.config.mjs
@@ -13,4 +13,7 @@ const withNextra = nextra({
 export default withNextra({
   reactStrictMode: true,
   eslint: { ignoreDuringBuilds: true },
+  env: {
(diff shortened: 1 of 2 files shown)


docs/next.config.mjs
```diff
@@ -13,4 +13,7 @@ const withNextra = nextra({
 export default withNextra({
   reactStrictMode: true,
   eslint: { ignoreDuringBuilds: true },
+  env: {
+    NO_INDEX: process.env.VERCEL_ENV !== "production" ? "true" : "false",
+  },
```
juliusmarminge, line 18: need `nextConfig.env` to inject an otherwise server-only env to the client bundle

PR #728, opened 2024-03-23
docs: Update ut-api.mdx to match type defs

FortyGazelle700:
Looking at `uploadthing/packages/uploadthing/src/sdk/types.ts` lines `83-85` I see the following lines of code that do not match the documentation.
```ts
type KeyRename = { key: string; newName: string };
type CustomIdRename = { customId: string; newName: string };
export type RenameFileUpdate = KeyRename | CustomIdRename;
```


juliusmarminge (approved):
Thanks!

markflorkowski (approved):

PR #731, opened 2024-03-25
test: implement polling and retries in test

juliusmarminge:
more tests to improve confidence of Effect rewrite

### DIFF
--- a/packages/uploadthing/src/internal/multi-part.ts
+++ b/packages/uploadthing/src/internal/multi-part.ts
@@ -113,7 +113,7 @@ export async function uploadPartWithProgress(
         // Add a delay before retrying (exponential backoff can be used)
         const delay = Math.pow(2, retryCount) * 1000;
         await new Promise((res) => setTimeout(res, delay));
-        await uploadPartWithProgress(opts, retryCount + 1); // Retry the request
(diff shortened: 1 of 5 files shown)


patches/msw+2.2.10.patch
juliusmarminge, line 1: Ref: https://github.com/mswjs/msw/pull/2108/files

packages/uploadthing/src/internal/multi-part.ts
```diff
@@ -113,7 +113,7 @@ export async function uploadPartWithProgress(
         // Add a delay before retrying (exponential backoff can be used)
         const delay = Math.pow(2, retryCount) * 1000;
         await new Promise((res) => setTimeout(res, delay));
-        await uploadPartWithProgress(opts, retryCount + 1); // Retry the request
+        resolve(await uploadPartWithProgress(opts, retryCount + 1)); // Retry the request
```
juliusmarminge, line 116: how funny that when validating the new solution i think i found a bug in the current solution xddd

the promise never resolved if it retried so it just hang forever 😅

markflorkowski: 😅

PR #732, opened 2024-03-26
chore: solid pass @typescript-eslint/prefer-optional-chain

markflorkowski:
Not sure why this wasn't caught in other branches, but should resolve lint issue for https://github.com/pingdotgg/uploadthing/pull/225#issuecomment-2019130693


markflorkowski:
Note: intentionally skipping changeset, as this is not a functional change.

PR #734, opened 2024-03-26
fix: dropzone form submit

### LINKED ISSUE #730: [bug]: FE: form doesn't submit when it contains uploadthing dropzone

```bash
System:
    OS: Linux 6.7 Garuda Linux
    CPU: (8) x64 Intel(R) Core(TM) i5-9300H CPU @ 2.40GHz
    Memory: 5.13 GB / 15.46 GB
    Container: Yes
    Shell: 5.2.26 - /bin/bash
  Binaries:
    Node: 18.19.1 - /usr/local/bin/node
    Yarn: 1.22.21 - /usr/local/bin/yarn
    npm: 10.2.4 - /usr/local/bin/npm
    pnpm: 8.15.3 - /usr/local/bin/pnpm
    bun: 1.0.26 - ~/.bun/bin/bun
  npmPackages:
    @uploadthing/react: ^6.4.0 => 6.4.0 
    typescript: ^5 => 5.4.3 
    uploadthing: ^6.6.0 => 6.6.0
```

### Describe the bug

When you have a UploadDropzone wrapped in a form, and then press enter on a input in the form, the form is expected to submit, but with a UploadDropzone inside, it doesn't submit.

Repo with reproduction: https://github.com/soulr344/uploadthing-bug-form
Live reproduction: https://uploadthing-bug-form.vercel.app/bug

### Link to reproduction

https://uploadthing-bug-form.vercel.app/bug

### To reproduce

Open the link above, you should have a form in there. Type something in the demo input, and press enter. You should get a message saying the form is submitted at the end of the form. Then, click on reset status, tick the "Enable Uploadthing Field", type something on the Demo input and press enter. It doesnt submit anymore.

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

- [x] 🙋‍♂️ Yes, I'd be down to file a PR fixing this bug!

### Code of Conduct

juliusmarminge:
Closes #730


markflorkowski (approved):
The same time as this, can you make sure the `<label />`s are using the correct `htmlFor` values? I think I remember seeing that they were invalid

juliusmarminge:
> The same time as this, can you make sure the `<label />`s are using the correct `htmlFor` values? I think I remember seeing that they were invalid

Do we even need `htmlFor` when the `<input/>` tag is a child of the `<label/>`?

markflorkowski:
>  Do we even need htmlFor when the <input/> tag is a child of the <label/>?

Technically, no, it is following guidelines to not use the `htmlFor` as long as the first element within the label is the input being labelled. I prefer to always use it though, since it doesn't hurt, and wont be forgotten if we were to refactor the component. 

That said, it definitely should not be incorrect (pointing to an id that does not exist) as it currently is:
image

juliusmarminge:
>  I prefer to always use it though, since it doesn't hurt, and wont be forgotten if we were to refactor the component.

Just that it can conflict with a different id the user has set in their form.

Could use `useId` maybe? Only available in React though. I think leaving it out is fine?

markflorkowski:
> Could use useId maybe? Only available in React though. I think leaving it out is fine?

Fair enough, just make sure to remove existing references 👍

markflorkowski (approved):

Issue #710, opened 2024-03-15
feat: minimum upload 


ccakmak60:
### Describe the feature you'd like to request

a way to set a minimum requirement for images uploaded 

### Describe the solution you'd like to see

add a minFileCount in the imageUploader 

### 👨‍👧‍👦 Contributing

- [ ] 🙋‍♂️ Yes, I'd be down to file a PR implementing this feature!

markflorkowski:
Agreed, this seems like a useful feature!

markflorkowski:
Can you explain more about the use case you have for this field? Just trying to make sure I understand how this would be used so we can design the API appropriately 👍

ccakmak60:
Sorry for the late reply, but my use case was uploading a minimum number of images for a user to train an image generation model.

PR #739, opened 2024-04-03
feat: minFileCount option on RouteConfig

markflorkowski:
Addresses #710

- **add minFileCount option to RouteConfig**
- **update docs to include minFileCount**

To do:

- [ ] typescript (validate that minFileCount <= maxFileCount)
- [ ] test -- add test for ^^
- [ ] confirm expected behaviour (should min be 0, or should it be 1?)

### DIFF
--- a/packages/shared/src/component-utils.ts
+++ b/packages/shared/src/component-utils.ts
@@ -61,10 +61,14 @@ export const INTERNAL_doFormatting = (config?: ExpandedRouteConfig): string => {
   const key = allowedTypes[0];
   const formattedKey = formattedTypes[0];
 
-  const { maxFileSize, maxFileCount } = config[key]!;
+  const { maxFileSize, maxFileCount, minFileCount } = config[key]!;
 
   if (maxFileCount && maxFileCount > 1) {
-    return `${formattedKey}s up to ${maxFileSize}, max ${maxFileCount}`;
+    if (minFileCount > 0) {
+      return `${formattedKey}s up to ${maxFileSize}, at least ${minFileCount}, max ${maxFileCount}`;
+    } else {
+      return `${formattedKey}s up to ${maxFileSize}, max ${maxFileCount}`;
+    }
   } else {
     return `${formattedKey} (${maxFileSize})`;
   }
--- a/packages/shared/src/component-utils.ts
+++ b/packages/shared/src/component-utils.ts
@@ -64,7 +64,7 @@ export const INTERNAL_doFormatting = (config?: ExpandedRouteConfig): string => {
   const { maxFileSize, maxFileCount, minFileCount } = config[key]!;
 
   if (maxFileCount && maxFileCount > 1) {
-    if (minFileCount > 0) {
+    if (minFileCount > 1) {
       return `${formattedKey}s up to ${maxFileSize}, at least ${minFileCount}, max ${maxFileCount}`;
     } else {
       return `${formattedKey}s up to ${maxFileSize}, max ${maxFileCount}`;
--- a/packages/shared/src/component-utils.ts
+++ b/packages/shared/src/component-utils.ts
@@ -65,7 +65,7 @@ export const INTERNAL_doFormatting = (config?: ExpandedRouteConfig): string => {
 
   if (maxFileCount && maxFileCount > 1) {
     if (minFileCount > 1) {
-      return `${formattedKey}s up to ${maxFileSize}, at least ${minFileCount}, max ${maxFileCount}`;
+      return `${minFileCount} - ${maxFileCount} ${formattedKey}s up to ${maxFileSize}`;
     } else {
       return `${formattedKey}s up to ${maxFileSize}, max ${maxFileCount}`;
     }
--- a/packages/shared/src/types.ts
+++ b/packages/shared/src/types.ts
@@ -103,8 +103,6 @@ type RouteConfig = {
   acl?: ACL; // default is set on UT server, not backfilled like other options
 };
 
-//
-
 export type FileRouterInputKey = AllowedFileType | MimeType;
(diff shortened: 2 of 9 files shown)


packages/shared/src/component-utils.ts
```diff
@@ -61,10 +61,14 @@ export const INTERNAL_doFormatting = (config?: ExpandedRouteConfig): string => {
   const key = allowedTypes[0];
   const formattedKey = formattedTypes[0];
 
-  const { maxFileSize, maxFileCount } = config[key]!;
+  const { maxFileSize, maxFileCount, minFileCount } = config[key]!;
 
   if (maxFileCount && maxFileCount > 1) {
-    return `${formattedKey}s up to ${maxFileSize}, max ${maxFileCount}`;
+    if (minFileCount > 0) {
```
juliusmarminge, line 67: any reason default isn't 1?

packages/uploadthing/test/upload-builder.test.ts
```diff
@@ -15,6 +15,9 @@ it("typeerrors for invalid input", () => {
   // @ts-expect-error - invalid size format
   f({ image: { maxFileSize: "1gb" } });
 
+  // TODO: add test for minFileCount <= maxFileCount
+  //f({ image: { minFileCount: 2, maxFileCount: 1 } });
```
juliusmarminge, line 19: i think this will be tough to assure on the typelevel - runtime check should be enough?

packages/uploadthing/src/internal/handler.ts
```diff
@@ -101,6 +101,44 @@ const fileCountLimitHit = (
   return { limitHit: false };
 };
 
+const fileCountMinHit = (
```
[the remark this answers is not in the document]
markflorkowski: Definitely could be "check bounds"

packages/shared/src/component-utils.ts
markflorkowski: I had this as an open question in the pr description. Not sure which makes more sense

packages/uploadthing/test/upload-builder.test.ts
markflorkowski: When would that check happen though?

packages/shared/src/component-utils.ts
juliusmarminge: no reason to send a request for 0 files - just an unnecessary roundtrip to the UT server

packages/uploadthing/test/upload-builder.test.ts
markflorkowski: Ok, so not when expanding the config, actually when you try to do an upload. Got it.

packages/shared/src/component-utils.ts
```diff
@@ -61,10 +61,14 @@ export const INTERNAL_doFormatting = (config?: ExpandedRouteConfig): string => {
   const key = allowedTypes[0];
   const formattedKey = formattedTypes[0];
 
-  const { maxFileSize, maxFileCount } = config[key]!;
+  const { maxFileSize, maxFileCount, minFileCount } = config[key]!;
 
   if (maxFileCount && maxFileCount > 1) {
-    return `${formattedKey}s up to ${maxFileSize}, max ${maxFileCount}`;
+    if (minFileCount > 1) {
+      return `${formattedKey}s up to ${maxFileSize}, at least ${minFileCount}, max ${maxFileCount}`;
```
juliusmarminge, line 68: will this text fit well? maybe

`${minFileCount} - ${maxFileCount} ${formattedKey}s up to ${maxFileSize}`

if it's tight on space. haven't run an example so no idea if there is spaceor not

markflorkowski: Will check

packages/uploadthing/src/internal/handler.ts
```diff
@@ -62,43 +62,63 @@ const createUTFetch = (
...
+    if (count > max) {
+      results.limitHit = true;
+      results.type = key;
+      results.limit = max;
+      results.count = count;
     }
   }
 
-  return { limitHit: false };
+  return results;
```
juliusmarminge, line 121: micro-optimization: we can just return early in the for loop to skip extra iterations

packages/shared/src/types.ts
```diff
@@ -94,13 +94,17 @@ export type Time =
...
 type RouteConfig = {
   maxFileSize: FileSize;
   maxFileCount: number;
+  minFileCount: number; // must be <= maxFileCount
   contentDisposition: ContentDisposition;
   acl?: ACL; // default is set on UT server, not backfilled like other options
 };
 
+//
+
```
juliusmarminge, line 107:
```suggestion
```

packages/uploadthing/src/internal/handler.ts
```diff
@@ -63,42 +63,59 @@ const createUTFetch = (
...
       logger.error(routeConfig, key);
+      const configKey =
+        typeof min !== "number" ? "minFileCount" : "maxFileCount";
       throw new UploadThingError({
         code: "BAD_REQUEST",
-        message: "Invalid config during file count",
-        cause: `Expected route config to have a maxFileCount for key ${key} but none was found.`,
+        message: `Invalid config during file count - missing ${configKey}`,
+        cause: `Expected route config to have a ${configKey} for key ${key} but none was found.`,
       });
```
juliusmarminge, line 91: i think this is actually the wrong error message?

the `minFileCount`/`maxFileCount` can never be undefined since we backfill defaults, right?

it's the `routeConfig[key]` which may be nullable. So the error message should probably be related to that?

juliusmarminge (approved):
LGTM

PR #740, opened 2024-04-03
feat: Add support for 'model/step' MIME type

kaspnilsson:
Was missing from this allowlist. Does this require additional work?


markflorkowski (approved):

t3dotgg:
Good catch, ty for adding! no additional work, we'll get this in for next release 🙏

PR #746, opened 2024-04-05
feat: `getInputProps`

juliusmarminge:


packages/react/src/components/button.tsx
```diff
@@ -106,128 +104,104 @@ export function UploadButton<
 
   const fileInputRef = useRef<HTMLInputElement>(null);
   const labelRef = useRef<HTMLLabelElement>(null);
-  const [uploadProgressState, setUploadProgress] = useState(
+  const [uploadProgress, setUploadProgress] = useState(
     $props.__internal_upload_progress ?? 0,
   );
-  const [files, setFiles] = useState<File[]>([]);
```
juliusmarminge, line 112: i think it makes sense for useUploadThing to control state? Especially if the `getInputProps` should have an onChange that sets files. Felt weird passing those in like `getInputProps({ mode, files, setFiles })`. By letting the hook own the state we have access to it in `getInputProps` for "free"

packages/react/src/components/button.tsx
```diff
@@ -106,128 +104,104 @@ export function UploadButton<
...
-  const [isManualTriggerDisplayed, setIsManualTriggerDisplayed] =
-    useState(false);
```
juliusmarminge, line 114: this state didn't make any sense to me. it's the same as `mode === "manual" && files.length > 0`. Could declare as a variable but just inlined it in the place it was used

packages/react/src/components/button.tsx
```diff
@@ -106,128 +104,104 @@ export function UploadButton<
...
-  const uploadProgress =
-    $props.__internal_upload_progress ?? uploadProgressState;
```
juliusmarminge, line 116: we already pass `__internal_upload_progress` into `setState` so no need to force it here too. We never chagne the progress in docs so doesn't have to be "dynamic"

packages/react/src/components/button.tsx
```diff
@@ -106,128 +104,104 @@ export function UploadButton<
...
-      if (!pastedFiles) return;
+    const pastedFiles = getFilesFromClipboardEvent(event);
+    if (!pastedFiles) return;
 
-      setFiles((prev) => [...prev, ...pastedFiles]);
+    let filesToUpload = pastedFiles;
+    setFiles((prev) => {
+      filesToUpload = [...prev, ...pastedFiles];
+      return filesToUpload;
+    });
```
juliusmarminge, line 161: This bug was fixed for dropzone in an earlier PR (https://github.com/pingdotgg/uploadthing/pull/622/files)

packages/react/src/components/button.tsx
juliusmarminge: Have to think how it's used in forms though

t3dotgg (approved):
The getInputProps bit looks great to me, hopefully that's not too hard to get up to date. Sorry for letting this slip 🙃

packages/react/src/hooks/use-uploadthing.ts
```diff
@@ -0,0 +1,265 @@
...
+  const useUploadThing = <
+    TEndpoint extends keyof TRouter,
+    TSkipPolling extends boolean = false,
+  >(
+    endpoint: TEndpoint,
+    opts?: UseUploadthingProps<TRouter, TEndpoint, TSkipPolling>,
+  ) => {
+    const [files, setFiles] = useControllableState({
+      prop: opts?.files,
+      onChange: opts?.onFilesChange,
```
juliusmarminge, line 81: ## note to make this compat with the changes introduced in #886

PR #747, opened 2024-04-07
chore: rename solid helper to `createUploadThing`

juliusmarminge:
Ref: https://github.com/pingdotgg/uploadthing/pull/225#issuecomment-2041461668

Follows `createSignal`, `createEffect`, `createMemo` etc etc, as well as some community libs (`createQuery` for TanStack Query for example)

### DIFF
--- a/docs/src/pages/getting-started/solid.mdx
+++ b/docs/src/pages/getting-started/solid.mdx
@@ -177,7 +177,7 @@ export default function Home() {
 UploadThing provides a few premade components that you can use to upload files
 in your app. For component customization, see [Theming](/theming), and for even
 more advanced use cases check out the
-[hook usage](#advanced-usage-useuploadthing)
+[hook usage](#advanced-usage-createuploadthing)
 
 #### UploadButton
 
@@ -293,7 +293,7 @@ export default function Home() {
 }
 ```
 
-## Advanced usage: `useUploadThing`
+## Advanced usage: `createUploadThing`
 
 For advanced use cases, the premade components might not be flexible enough to
 suit your needs. In that case, you can use the `useUploadThing` hook to build
@@ -301,9 +301,9 @@ your own components with a custom upload flow:
 
 <Steps>
 
-### Create the hook
+### Create the helper
 
-First, generate a typed hook using the `generateSolidHelpers` function from
+First, generate a typed helper using the `generateSolidHelpers` function from
 `@uploadthing/solid`:
 
 ```tsx copy filename="src/utils/uploadthing.ts"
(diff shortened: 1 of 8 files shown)


docs/src/pages/getting-started/solid.mdx
```diff
@@ -293,42 +293,43 @@ export default function Home() {
...
+## Advanced usage: `createUploadThing`
 
 For advanced use cases, the premade components might not be flexible enough to
 suit your needs. In that case, you can use the `useUploadThing` hook to build
 your own components with a custom upload flow:
 
 <Steps>
 
-### Create the hook
+### Create the helper
```
juliusmarminge, line 304: idk if there is an established name for these `createX` functions in Solid...

docs/src/pages/getting-started/solid.mdx
```diff
@@ -293,42 +293,43 @@ export default function Home() {
...
+import { createUploadThing } from "~/utils/uploadthing";
 
 async function compress(file: File) {
   // Run some compression algorithm on the file
   return file;
 }
 
 export default function Home() {
-  const { startUpload } = useUploadThing("profileImage", {
+  const { startUpload } = createUploadThing("profileImage", {
```
juliusmarminge, line 332: Destructuring also doesn't seem to very Solid-y: 
https://tanstack.com/query/latest/docs/framework/solid/overview#important-differences-between-solid-query--react-query

docs/src/pages/getting-started/solid.mdx
markflorkowski: I think it would be considered a "store" using Solid's vocabulary?
image
<https://docs.solidjs.com/advanced-concepts/fine-grained-reactivity#reactive-primitives>

markflorkowski: Not actually sure there is one as I look into this more

N0tExisting:
> Not sure if it matters, but we use `createUploadthing()` for creating the router on the backend already...

should be fine, because its only exported in the `uploadthing/server` package and only imported once in server side code, making any conflicts quite unlikely.

docs/src/pages/getting-started/solid.mdx
N0tExisting: `createX` functions are for reactive utilies/helper/primitives[^1], so I would say that calling it a helper would be correct.

[^1]: <https://primitives.solidjs.community/#make-non-reactive-vs-create-reactive>

markflorkowski:
> making any conflicts quite unlikely.

less concerned about actual code related conflicts, more about it being confusing for users.

markflorkowski (approved):

PR #293, opened 2023-08-29
feat: VueJS + NuxtJS support (#199)

markflorkowski:
Add support for NuxtJS and VueJS

- [x] Create builder function for Nuxt that will scaffold API route handler
- [x] Adjust types to comply with NuxtJS type system in case of API Routes
- [x] Expose proper Request object in middleware
- [x] Create package for VueJS support
- [x] Custom ~hooks~ composables for VueJS
- [x] Button
- [x] Dropzone
- [x] Upload progress indicator in default components

Update:

Vue package with UploadButton is added. Currently, it has external dependency on [VueUse](https://github.com/vueuse/vueuse) and it's useFetch. Ideally, it should be replaced with something more lightweight

Since I've used `defineComponent` function to create UploadButton component, I had to make single top-level prop that can be passed in component config
image

So everything else can be passed into config prop object

# Example of usage in SFC:
 
image

Typesafety and autocomplete works as expected
image
image

Correct types in middleware
image

Issues present so far that has to be fixed:
- [x] Import from `@uploadthing/vue` is not working. import from `@uploadthing/vue/index` works but first option is more preferable
- [x] Can't make TailwindCSS work in NuxtJS example for some reason. It does not generate classes
- [x] Both UploadButton and UploadDropzone require `ClientOnly` to work properly. Not sure how much of an issue it is though

# Example from FE
Monosnap screencast 2023-07-06 13-26-01

PR #753, opened 2024-04-10
fix: solid dropzone stopping too many events

juliusmarminge:


packages/dropzone/src/solid.tsx
```diff
@@ -192,7 +185,7 @@ export function createDropzone(_props: DropzoneOptions) {
       }
     });
 
-    if (isValidQuantity(acceptedFiles, props.multiple, props.maxFiles)) {
+    if (!isValidQuantity(acceptedFiles, props.multiple, props.maxFiles)) {
```
juliusmarminge, line 188: cc https://github.com/pingdotgg/uploadthing/pull/293#discussion_r1558319740

PR #755, opened 2024-04-10
chore: Automatically mark/close stale issues

markflorkowski:
Automatically marks issues as stale after 2 weeks of inactivity. Closes stale issues after a further 7 days of inactivity. 

Would consider making these shorter as well.


juliusmarminge (approved):
I’d lean a bit shorter, maybe 10+5 but its fine as is

markflorkowski:
I'll reduce it, also maybe switch to exclude certain labels rather than only checking certain ones

Issue #769, opened 2024-04-25
[bug]: Module not found: Can't resolve 'fast-check'


duc-hoang-161:
### Provide environment information

```bash
This is my local
  System:
    OS: macOS 13.1
    CPU: (8) arm64 Apple M1
    Memory: 233.52 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 20.10.0 - ~/.nvm/versions/node/v20.10.0/bin/node
    Yarn: 1.22.21 - ~/.nvm/versions/node/v20.10.0/bin/yarn
    npm: 10.2.3 - ~/.nvm/versions/node/v20.10.0/bin/npm
    pnpm: 8.9.2 - ~/Library/pnpm/pnpm
  Browsers:
    Chrome: 124.0.6367.79
    Safari: 16.2
  npmPackages:
    @uploadthing/react: 6.5.0 => 6.5.0
    typescript: ^5.4.5 => 5.4.5
    uploadthing: 6.10.0 => 6.10.0
```

### Describe the bug

I encountered a build failure in my project after upgrading uploadthing to the latest version (6.10.0). The build failed on Vercel (yarn@1.22.19), in my Bitbucket pipeline (yarn@4.1.0), and on my local. After reverting to uploadthing@6.9.0 and @uploadthing/react@6.4.4, the build was successful.

Below is the log from with-tailwindcss example. It has the same error.

```
 ❯ yarn install
yarn install v1.22.21
info No lockfile found.
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
warning "uploadthing > @effect/schema@0.66.8" has unmet peer dependency "fast-check@^3.13.2".
[4/4] 🔨  Building fresh packages...

success Saved lockfile.
✨  Done in 28.36s.
❯ cp .env.example .env
❯ yarn build
yarn run v1.22.21
$ next build
  ▲ Next.js 14.2.1
  - Environments: .env

   Creating an optimized production build ...
Failed to compile.

./node_modules/@effect/schema/dist/esm/FastCheck.js
Module not found: Can't resolve 'fast-check'

https://nextjs.org/docs/messages/module-not-found

Import trace for requested module:
./node_modules/@effect/schema/dist/esm/Arbitrary.js
./node_modules/@effect/schema/dist/esm/Schema.js
./node_modules/@uploadthing/shared/dist/index.js
[...]

PR #770, opened 2024-04-26
fix: don't validate schema for non-ok requests

juliusmarminge:
this causes some confusion errors since if the requests failed the schema wont match and it's better to show an error with the response message

updates effect, we can now drop the $-adapter thing to make yielding much nicer 👍🏼

### DIFF
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -34,8 +34,8 @@
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
-    "@effect/schema": "^0.66.0",
-    "effect": "^3.0.0",
+    "@effect/schema": "^0.66.9",
+    "effect": "^3.0.6",
     "fast-check": "^3.17.2",
     "std-env": "^3.7.0"
   },
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -35,8 +35,8 @@
   },
   "dependencies": {
     "@effect/schema": "^0.66.0",
-    "fast-check": "^3.17.2",
     "effect": "^3.0.0",
+    "fast-check": "^3.17.2",
     "std-env": "^3.7.0"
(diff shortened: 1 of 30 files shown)


packages/shared/package.json
```diff
@@ -36,6 +36,7 @@
   "dependencies": {
     "@effect/schema": "^0.66.0",
     "effect": "^3.0.0",
+    "fast-check": "^3.17.2",
```
juliusmarminge, line 39: Closes #769

packages/uploadthing/src/internal/logger.ts
```diff
@@ -108,3 +109,17 @@ export const initLogger = (level: LogLevel | undefined) => {
...
+//   Logger.make(({ logLevel, message }) => {
+//     console.log("Creaing logger", logLevel, message);
+//     logger.info(message);
+//   }),
+// );
+
+export const withLogger = (level: LogLevel = "info") =>
+  Logger.replace(
+    Logger.defaultLogger,
+    Logger.make((opts) => logger.log(opts)),
```
juliusmarminge, line 124: TODO:

markflorkowski (approved):

packages/uploadthing/src/internal/handler.ts
```diff
@@ -290,7 +309,7 @@ const runRouteMiddleware = (opts: {
     const filesWithCustomIds = files.map((file, idx) => {
       const theirs = metadata[UTFiles]?.[idx];
       if (theirs && theirs.size !== file.size) {
-        logger.warn("File size mismatch. Reverting to original size");
+        // Effect.logWarning("File size mismatch. Reverting to original size");
```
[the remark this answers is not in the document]
juliusmarminge: nope

PR #775, opened 2024-04-29
refactor: update "effect" & switch to alias imports

tim-smart:


packages/uploadthing/src/client.ts
```diff
@@ -34,11 +35,11 @@ import type {
 } from "./types";
 
 export {
-  /** @public */
-  generateMimeTypes,
   /** @public */
   generateClientDropzoneAccept,
   /** @public */
+  generateMimeTypes,
+  /** @public */
```
patroza, line 42: looks like a mistake

juliusmarminge (approved):

PR #776, opened 2024-04-29
refactor: rewrite handler more effecty

juliusmarminge:


packages/shared/src/tagged-errors.ts
```diff
@@ -53,5 +51,5 @@ export class RetryError extends TaggedError("RetryError") {}
 export class FetchError extends TaggedError("FetchError")<{
   readonly input: RequestInfo | URL;
   readonly error: unknown;
-  readonly data?: Json;
+  readonly data?: any;
```
patroza, line 54: `unknown` is better than `any`, as `any` breaks all type safety, but otherwise perhaps generics are helpful?

packages/uploadthing/src/internal/handler.ts
```diff
@@ -305,230 +262,211 @@ const runRouteMiddleware = (opts: {
...
+  if (opts.isDev) {
+    const layer = Layer.succeed(fetchContext, yield* fetchContext);
+    promise = Effect.forEach(
+      presignedUrls,
+      (file) => conditionalDevServer(file.key, opts.apiKey),
+      { concurrency: 10 },
+    ).pipe(
+      Effect.provide(ConsolaLogger),
+      Effect.provide(layer),
+      Effect.runPromise,
```
patroza, line 380: Perhaps create a runtime with `ConsolaLogger`, and use that instead.
The fetchContextLayer would also be better as `Effect.provideService(fetchContext, yield* fetchContext)` as layers originally are optimised to construct once at startup (though there are exceptions)

PR #778, opened 2024-04-30
refactor: make some suggestions

tim-smart:


packages/shared/src/effect.ts
```diff
@@ -19,17 +19,19 @@ export type FetchContextTag = {
     "x-uploadthing-be-adapter": string | undefined;
   };
 };
-export const fetchContext =
-  Context.GenericTag<FetchContextTag>("fetch-context");
+export class FetchContext extends Context.Tag("uploadthing/FetchContext")<
```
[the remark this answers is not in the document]
tim-smart: Also remove the `Tag` from the name.

packages/uploadthing/src/client.ts
```diff
@@ -102,25 +101,25 @@ export const genUploader = <TRouter extends FileRouter>(
...
+    uploadFilesInternal<TRouter, TEndpoint, TSkipPolling>(endpoint, {
       ...opts,
       url: resolveMaybeUrlArg(initOpts?.url),
       package: initOpts.package,
       // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
       input: (opts as any).input as inferEndpointInput<TRouter[TEndpoint]>,
-    }).pipe(Effect.provide(layer), Effect.runPromise);
-  };
+    }).pipe(
+      Effect.provideService(FetchContext, {
```
tim-smart, line 112: You can use Effect.provideService instead of a Layer here. Layer's are better for composing lots of services together, rather than providing simply one-off deps.

packages/uploadthing/src/internal/handler.ts
```diff
@@ -104,12 +117,12 @@ export const buildRequestHandler =
   ): RequestHandler<Args> =>
   ({ req, middlewareArgs }) =>
     pipe(
-      req instanceof Request ? Effect.succeed(req) : req,
+      Effect.isEffect(req) ? req : Effect.succeed(req),
```
tim-smart, line 120: Effect.isEffect is useful for this kind of thing - using `instanceof` can break if you ship both cjs & esm builds and the user ends up with duplicate modules.

packages/uploadthing/src/internal/validate-request-input.ts
```diff
@@ -291,7 +293,7 @@ export const parseAndValidateRequest = (opts: {
     // FIXME: This should probably provide the full context at once instead of
     // partially in the `runRequestHandlerAsync` and partially in here...
     // Ref: https://discord.com/channels/@me/1201977154577891369/1207441839972548669
-    const contextValue = yield* fetchContext;
+    const contextValue = yield* FetchContext;
```
juliusmarminge, line 296: Is there any way to fix this fixme comment here? Would prefer if we didnt have to init them as undefined higher up

tim-smart: One way to do, and how I would probably approach it, is to split it into two seperate tags. One for the fetch instance and one for the base headers.

juliusmarminge: yea seems simple enough

juliusmarminge (approved):
Thanks 🙏

PR #783, opened 2024-05-01
refactor(client): move things to pipelines

datner:
this goes deep, so 1 small change at a time, ya know?

### DIFF
--- a/packages/shared/src/effect.ts
+++ b/packages/shared/src/effect.ts
@@ -69,13 +69,13 @@ export const fetchEffUnknown = (
       Effect.tryPromise({
         try: async () => {
           const json = await res.json();
-          return { ok: res.ok, json, status: res.status };
+          return { json, ok: res.ok, status: res.status };
         },
         catch: (error) => new FetchError({ error, input }),
       }),
     ),
     Effect.filterOrFail(
-      (_) => _.ok,
+      ({ ok }) => ok,
       ({ json, status }) =>
         new FetchError({
           error: `Request to ${requestUrl} failed with status ${status}`,
@@ -113,7 +113,7 @@ export const fetchEffJson = <Schema>(
       }),
     ),
     Effect.filterOrFail(
-      (_) => _.ok,
+      ({ ok }) => ok,
       ({ json, status }) =>
         new FetchError({
           error: `Request to ${requestUrl} failed with status ${status}`,
@@ -121,7 +121,7 @@ export const fetchEffJson = <Schema>(
           input,
         }),
     ),
-    Effect.map(_ => _.json),
+    Effect.map(({ json }) => json),
     Effect.andThen(S.decode(schema)),
     Effect.withSpan("fetchJson", {
       attributes: { input: JSON.stringify(input) },
--- a/packages/shared/src/effect.ts
+++ b/packages/shared/src/effect.ts
@@ -57,7 +57,12 @@ export const fetchEffUnknown = (
   /** Schema to be used if the response returned a 2xx  */
   init?: RequestInit,
 ): Effect.Effect<unknown, FetchError, FetchContext> => {
-  const requestUrl = input instanceof Request ? input.url : input.toString();
+  const requestUrl =
+    typeof input === "string"
+      ? input
+      : input instanceof URL
+        ? input.toString()
+        : input.url;
 
   return fetchEff(input, init).pipe(
     Effect.andThen((res) =>
--- a/packages/shared/src/effect.ts
+++ b/packages/shared/src/effect.ts
@@ -50,6 +50,40 @@ export const fetchEff = (
     }),
   );
 
+// TODO: rename the other one to fetchEffJsonSchema and this to fetchEffJson
+// though generally I would avoid XandY kind utils in favor of composition
+export const fetchEffUnknown = (
+  input: RequestInfo | URL,
+  /** Schema to be used if the response returned a 2xx  */
+  init?: RequestInit,
+): Effect.Effect<unknown, FetchError, FetchContext> => {
+  const requestUrl = input instanceof Request ? input.url : input.toString();
+
+  return fetchEff(input, init).pipe(
+    Effect.andThen((res) =>
+      Effect.tryPromise({
+        try: async () => {
+          const json = await res.json();
+          return { ok: res.ok, json, status: res.status };
+        },
+        catch: (error) => new FetchError({ error, input }),
+      }),
+    ),
+    Effect.filterOrFail(
+      (_) => _.ok,
+      ({ json, status }) =>
+        new FetchError({
+          error: `Request to ${requestUrl} failed with status ${status}`,
+          data: json as Json,
+          input,
+        }),
+    ),
+    Effect.withSpan("fetchRawJson", {
+      attributes: { input: JSON.stringify(input) },
+    }),
+  );
+};
+
 export const fetchEffJson = <Schema>(
   input: RequestInfo | URL,
   /** Schema to be used if the response returned a 2xx  */
@@ -73,17 +107,16 @@ export const fetchEffJson = <Schema>(
         catch: (error) => new FetchError({ error, input }),
       }),
     ),
-    Effect.andThen(({ ok, json, status }) =>
-      ok
-        ? Effect.succeed(json)
-        : Effect.fail(
-            new FetchError({
-              error: `Request to ${requestUrl} failed with status ${status}`,
-              data: json as Json,
-              input,
-            }),
-          ),
+    Effect.filterOrFail(
+      (_) => _.ok,
+      ({ json, status }) =>
+        new FetchError({
+          error: `Request to ${requestUrl} failed with status ${status}`,
+          data: json as Json,
+          input,
+        }),
     ),
+    Effect.map(_ => _.json),
     Effect.andThen(S.decode(schema)),
     Effect.withSpan("fetchJson", {
       attributes: { input: JSON.stringify(input) },
--- a/packages/uploadthing/src/client.ts
+++ b/packages/uploadthing/src/client.ts
@@ -1,5 +1,5 @@
 import type { ParseError } from "@effect/schema/ParseResult";
-import * as Array from "effect/Array";
+import * as Arr from "effect/Array";
 import * as Console from "effect/Console";
 import * as Effect from "effect/Effect";
 import { unsafeCoerce } from "effect/Function";
@@ -155,7 +155,7 @@ const uploadFile = <
   },
   presigned: PresignedURLs[number],
 ) =>
-  Array.findFirst(opts.files, (file) => file.name === presigned.fileName).pipe(
+  Arr.findFirst(opts.files, (file) => file.name === presigned.fileName).pipe(
     Effect.tapError(() =>
       Console.error("No file found for presigned URL", presigned),
     ),
--- a/packages/uploadthing/src/client.ts
+++ b/packages/uploadthing/src/client.ts
@@ -1,14 +1,18 @@
 /* eslint-disable no-console -- Don't ship our logger to client, reduce size*/
 
 import type { ParseError } from "@effect/schema/ParseResult";
-import * as S from "@effect/schema/Schema";
+import * as Array from "effect/Array";
+import * as Console from "effect/Console";
 import * as Effect from "effect/Effect";
+import { unsafeCoerce } from "effect/Function";
+import * as Option from "effect/Option";
+import type * as Predicate from "effect/Predicate";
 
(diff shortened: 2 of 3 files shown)


packages/shared/src/effect.ts
```diff
@@ -50,6 +50,40 @@ export const fetchEff = (
...
+          error: `Request to ${requestUrl} failed with status ${status}`,
+          data: json as Json,
+          input,
+        }),
+    ),
+    Effect.withSpan("fetchRawJson", {
+      attributes: { input: JSON.stringify(input) },
+    }),
+  );
+};
```
datner, line 85: it would be best if the util could be broken into multiple components, when faced with `fetchAndParseJsonAndDecode` it's hard to later pick-and-choose which parts I want.
While it's a bit more annoying to have to do
```ts
fetch(...).pipe(
  Effect.flatMap(json),
  Effect.flatMap(decode)
)
```
it beings dividends as each piece is more useful on its own.

In this case, I wanted only the `fetch` part and the logic wrapped inside the `andParseJson` part, but without the `andDecode` part 😢

packages/shared/src/effect.ts
```diff
@@ -50,6 +50,40 @@ export const fetchEff = (
...
   );
 
+// TODO: rename the other one to fetchEffJsonSchema and this to fetchEffJson
+// though generally I would avoid XandY kind utils in favor of composition
+export const fetchEffUnknown = (
+  input: RequestInfo | URL,
+  /** Schema to be used if the response returned a 2xx  */
+  init?: RequestInit,
+): Effect.Effect<unknown, FetchError, FetchContext> => {
+  const requestUrl = input instanceof Request ? input.url : input.toString();
```
datner, line 60: I just realized that this is not an improvement, it's a bug. lol. 🐛

packages/shared/src/effect.ts
```diff
@@ -73,17 +107,16 @@ export const fetchEffJson = <Schema>(
...
-            }),
-          ),
+    Effect.filterOrFail(
+      (_) => _.ok,
+      ({ json, status }) =>
+        new FetchError({
+          error: `Request to ${requestUrl} failed with status ${status}`,
+          data: json as Json,
+          input,
+        }),
```
datner, line 117: imo `flatMap(() => cond ? succeed(val) : fail(error))` is a bit smelly. It works just find and is really simple, it's even idiomatic since it's a defacto `filterMap`, but it just reads meh. `filterOrFail` immediately exposes the _semantics_ expressed by the tertiary.

In the end, this is just stylistic

packages/uploadthing/src/client.ts
```diff
@@ -78,16 +83,19 @@ const uploadFilesInternal = <
...
-    );
-
-    return yield* Effect.forEach(
-      presigneds,
+      PresignedURLResponseSchema, // don't want to break this, you do it
+    ),
+    Effect.forEach(
       (presigned) =>
-        uploadFile(String(endpoint), { ...opts, reportEventToUT }, presigned),
+        uploadFile<TRouter, TEndpoint, TSkipPolling, TServerOutput>(
```
datner, line 90: passing down the generics helps us dodge casting

packages/uploadthing/src/client.ts
```diff
@@ -122,6 +130,28 @@ export const genUploader = <TRouter extends FileRouter>(
...
+type PollingResponse = Done | StillWaiting;
+
+const isPollingResponse = (input: unknown): input is PollingResponse =>
+  input != null && // @datner: != is not accidental
+  typeof input === "object" &&
+  "status" in input &&
+  typeof input.status === "string" &&
+  (input.status === "done"
+    ? "callbackData" in input
+    : input.status === "still waiting");
```
datner, line 149: not as good as schema, but it's close :)

packages/uploadthing/src/client.ts
```diff
@@ -136,58 +166,50 @@ const uploadFile = <
...
+          message: "No file found for presigned URL",
+          cause: `Expected file with name ${presigned.fileName} but got '${opts.files.join(",")}'`,
+        }),
+    ),
+    Effect.tap((_) => opts.onUploadBegin?.({ file: _.name })),
+    Effect.tap((_) =>
+      "urls" in presigned
+        ? uploadMultipartWithProgress(_, presigned, opts)
+        : uploadPresignedPostWithProgress(_, presigned, opts),
+    ),
```
datner, line 186: I am not with enough context to understand the flow completely, but the pipe form exposes an interesting observation: nearly all our logic is side-effects-- even the main logic (I assume), the upload is something that is done not as part as the main flow. 

Optimally, in my opinion, instead of uploading and then polling in a detached manner like this, this entire thing can be made into a stream, uploads (and downloads) fit streams very naturally. It would allow a shape more similar to
```ts
const uploadFile = (file, opts) => 
  isomorphicUpload(file, opts).pipe(
        Stream.runForEach(progress =>  => {....}), // idk, report it? callback? w/e
        Effect.map(() => ({/* the big obj with the stuff */}))
)
```

The intuition to make the upload an effect is well placed, since it'll finish only once it... finishes! But given the subsequent polling, idk?
You can also concat a polling stream to the upload stream if only polling is possible. Anyways it's just a natural way to express this flow I think

packages/uploadthing/src/client.ts
```diff
@@ -136,58 +166,50 @@ const uploadFile = <
...
           while: (res) => res instanceof RetryError,
           schedule: exponentialBackoff,
         }),
-      );
-    }
-
-    return {
+        Effect.when(() => !opts.skipPolling),
+        Effect.map(Option.getOrNull),
+        Effect.map(unsafeCoerce<unknown, TServerOutput>),
```
datner, line 203: it's just an identity function `(_) => _`, but it's like a sore thumb in the code, telling you that you're making unsafe assumptions. Just a nice util to use over straight casting

packages/shared/src/effect.ts
juliusmarminge: i think that makes sense 👍🏼

packages/uploadthing/src/client.ts
```diff
@@ -122,6 +130,28 @@ export const genUploader = <TRouter extends FileRouter>(
...
+}
+interface StillWaiting {
+  status: "still waiting";
+}
+type PollingResponse = Done | StillWaiting;
+
+const isPollingResponse = (input: unknown): input is PollingResponse =>
+  input != null && // @datner: != is not accidental
+  typeof input === "object" &&
+  "status" in input &&
```
juliusmarminge, line 145: we have an `isObject` util which should allow you to do `isObject(input) && typeof input.status === "string"`

packages/shared/src/effect.ts
```diff
@@ -50,6 +50,45 @@ export const fetchEff = (
     }),
   );
 
+// TODO: rename the other one to fetchEffJsonSchema and this to fetchEffJson
+// though generally I would avoid XandY kind utils in favor of composition
```
juliusmarminge, line 54: if you wanna do this change now feel free - else I can look into it later

packages/uploadthing/src/client.ts
```diff
@@ -136,58 +166,50 @@ const uploadFile = <
...
-      }),
-      S.Struct({ status: S.Literal("still waiting") }),
-    );
-
-    let serverData: TServerOutput | null = null;
-    if (!opts.skipPolling) {
-      serverData = yield* fetchEffJson(presigned.pollingUrl, PollingResponse, {
+  Array.findFirst(opts.files, (_) => _.name === presigned.fileName).pipe(
+    Effect.tapError(() =>
+      Console.error("No file found for presigned URL", presigned),
```
juliusmarminge, line 171: what's the difference with `Console.error` and `Effect.logError`?

packages/uploadthing/src/client.ts
```diff
@@ -122,6 +129,20 @@ export const genUploader = <TRouter extends FileRouter>(
...
+
+const isPollingResponse = (input: unknown): input is PollingResponse => {
+  if (!isObject(input)) return false;
+  if (input.status === "done") return "callbackData" in input;
+  return input.status === "still waiting";
+};
+
+const isPollDone: Predicate.Refinement<PollingResponse, Done> = (
+  input,
+): input is Done => input.status === "done";
```
juliusmarminge, line 144: why does this function need 2 explicit types? will it be "fixed" in TS5.5?

packages/uploadthing/src/client.ts
```diff
@@ -1,14 +1,17 @@
-/* eslint-disable no-console -- Don't ship our logger to client, reduce size*/
-
 import type { ParseError } from "@effect/schema/ParseResult";
-import * as S from "@effect/schema/Schema";
+import * as Array from "effect/Array";
```
juliusmarminge, line 2: did this need to be aliased to something other than `Array` to work in react native? 

https://discord.com/channels/795981131316985866/1233104222954852454/1233174648573788200

not sure if it's the case when we do the `import * as Array` syntax though, I assume it will?

packages/shared/src/effect.ts
datner: it's just a bit of shotgun surgery since it's used all over, I can two-stage it. Keep the original just as a composition of extracted 'primitives'

packages/uploadthing/src/client.ts
datner: I was sure I answered you, sorry...

the `Console` module uses the users `Console` (99/100 it's just the `console`), meaning it prints the input and thats that.

The `Effect.log*` variants use the `Logger`, which is the full logging solution for effect. It has access to much more information like time, spans, annotations, and other such things logs it wherever.
Basically:
```ts
const obj = { a: 1 }

Console.log(obj) // { a: 1 }
Effect.logInfo(obj) // time="2024-05-02T08:30:00Z" level=info message="\{ \"a\": 1 \}" fiberId=1 session=abcde
```
You can add any arbitrary number of loggers, so for example a tracer (reports the logs as events on the current span) or a file-logger producing logFmt entries or some log aggregator like LogStash for chewing through arbitrary json.

TL;DR
debugging with `Console`, reporting with `Logger` (through `Effect.log*`)

packages/uploadthing/src/client.ts
datner: no idea, refinements alway needed convincing. I guess we can remove `Predicate.Refinement<PollingResponse, Done>` and let it just infer the refinement. Either is fine

packages/uploadthing/src/client.ts
datner: Ah good catch, force of habit from my own apps-- which are not react-native :)
`Arr` is fine, yes? `A` is also common

packages/uploadthing/src/client.ts
juliusmarminge: So are we doing it wrong here? https://github.com/pingdotgg/uploadthing/blob/e976fbecbe1e760c72827e6faefac7d783ee4df0/packages/uploadthing/src/internal/handler.ts#L62-L63

datner: Depending on what you want to achieve, if you want to invoke `console.log` and friends directly but not have control over the log level then yes, if you want more robust logging with all the bells and whistles then this is the proper way. Considering this comment:
https://github.com/pingdotgg/uploadthing/blob/e976fbecbe1e760c72827e6faefac7d783ee4df0/packages/uploadthing/src/internal/logger.ts#L141

I'd reckon that your goal would align more with the `Logger` rather than the `Console`, even if _currently_ it's being implemented in the same way as `Console` would be used. Note that `Logger` does the same thing you use `Consola` for
https://github.com/pingdotgg/uploadthing/blob/e976fbecbe1e760c72827e6faefac7d783ee4df0/packages/uploadthing/src/internal/logger.ts#L86-L106

You can replace it with `Logger` just fine. `Logger` instances have a `.log` method exposed by them so you can use them outside of effect as well. Though you would want a util for logging outside of effect since the exposed `.log` is quite demanding.
```ts
export const log = (logLevel: LogLevel.LogLevel, message: unknown) =>
  MyLogger.log({
    message,
    logLevel,
    annotations: HashMap.empty(),
    date: new Date(),
    cause: Cause.empty,
    spans: List.empty(),
    context: FiberRefs.empty(),
    fiberId: FiberId.none,
  });
```


TL;DR
Nah you're gucci, but I would shuffle some things around in that logger. I just changed a `console.error` to a `Console.error` but it should be `Effect.logError` to actually use the logger for printing

juliusmarminge (approved):

datner:
Thanks @juliusmarminge !
If you're interested in the next iteration (for example to make things more composable) hit me up in discord and I'll get to it

PR #794, opened 2024-05-08
refactor(client): decouple some effects for inferrence

datner:
- Decomposes `fetchEffJson` in favor of composing `fetchEff` with a new `parseResponseJson` helper
- Remvoes `@effect/schema` from client bundle (Resulting in -15k gzip in minimal-appdir example)
- adds `backend-apaters/client-vanilla` example (good reference + for bundle analyzing without nextjs)


packages/shared/src/effect.ts
```diff
@@ -31,68 +31,67 @@ export const fetchEff = (
...
+      return { json, ok: res.ok, status: res.status };
+    },
+    catch: (error) => new FetchError({ error, input: res.request }),
+  }).pipe(
     Effect.filterOrFail(
       ({ ok }) => ok,
       ({ json, status }) =>
         new BadRequestError({
-          input,
+          input: res.request,
```
juliusmarminge, line 73: Have to double check what value we're actually getting from attaching this property, i dont think we use anything other than the url.

We already track it on the span in the base fetchEff function, so for debuggability it feels like we should improve our usage of those before starting drilling these props onto a bunch of errors🤔

packages/shared/src/effect.ts
```diff
@@ -31,68 +31,67 @@ export const fetchEff = (
...
+      headers: {
+        ...filterObjectValues(baseHeaders, (v): v is string => v != null),
+        ...init?.headers,
+      },
+    });
+    return Effect.tryPromise({
+      try: () => fetch(req.clone()),
+      catch: (error) => new FetchError({ error, input: req.clone() }),
+    }).pipe(
+      Effect.map((res) => Object.assign(res.clone(), { request: req.clone() })),
```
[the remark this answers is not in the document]
datner: I'm just not sure when the `Request` is actually consumed and couldn't be bothered to check the docs hahahaha
I was crossing my fingers you'll know when clone is required and when its not.

Assigning the request to the response is because of a business requirement uploadthing establishes that responses need to know about their requests. Which is totally valid but I wasn't sure you want this requirement to stretch your scope.. With this you can access the request wherever you need to as long as you have access to the response.

I'll fix the clone, and I'm open to ideas about the assignment

packages/shared/src/effect.ts
datner: It depends what kind of debug story you're trying to achieve. I agree that the errors could be better described, I don't think that just recording the input brings us much value..

What do you think I should do? remove it altogether?

datner:
Sorry the PR was kinda sloppy this time around, I was on my way out of the office, probably should have made this a draft

packages/shared/src/effect.ts
```diff
@@ -24,72 +24,72 @@ export class FetchContext extends Context.Tag("uploadthing/FetchContext")<
...
   init?: RequestInit,
-): Effect.Effect<ResponseWithBoundRequest, FetchError, FetchContext> =>
+): Effect.Effect<ResponseWithURL, FetchError, FetchContext> =>
   Effect.flatMap(FetchContext, ({ fetch, baseHeaders }) => {
-    const req = new Request(input, {
-      ...init,
+    const reqInfo = {
+      url: input.toString(),
+      method: init?.method,
+      body: init?.body,
```
datner, line 42: Yeah thats a solution too, I fell in love with `Request` it seems haha

packages/shared/src/effect.ts
```diff
@@ -24,72 +24,72 @@ export class FetchContext extends Context.Tag("uploadthing/FetchContext")<
...
+      Effect.withSpan("fetch", { attributes: { reqInfo } }),
     );
   });
 
-export const json = (
-  res: ResponseWithBoundRequest,
-): Effect.Effect<unknown, FetchError | BadRequestError<Json>, never> =>
+export const parseJson = (
+  res: ResponseWithURL,
+): Effect.Effect<unknown, InvalidJsonError | BadRequestError, never> =>
```
datner: the json is being parsed, no doubt about it. But it needs to be parsed again later to turn it from an `unknown` to some more significant shape, which is why I chose `json`.  Since it just turns a bit array into a json entity-- or die trying

packages/shared/src/effect.ts
juliusmarminge: HappyDOM, MSW and the Request constructor doesn't play very nice together... and also the logs should be a bit cleaner if we only store what we really care about and not all the other fuzz

packages/shared/src/effect.ts
juliusmarminge: valid point

juliusmarminge (approved):
Great improvements. Gonna merge this as is but if you had ideas on making it more DRY (perhaps uniforming parseRequestJson and parseResponseJson, coupled with some validation thingy) as discussed on Discord PRs are always welcome :).

But this is a solid improvement and happy to merge baby steps!

PR #796, opened 2024-05-10
feat: allow specifying image size on server config

juliusmarminge:
- [x] add `imageSize` to server config
  ```ts
  const uploadRouter = {
    profilePicture: f({
      image: { 
        maxFileCount: 1,
        maxFileSize: "", // <-- should this be required? we wont validate size on server or is there some calculation we can do to calculate this for them?
        imageProperties: { // naming tbd
          // require a 400x400 image
          height: 400, width: 400,
          // same as above, but height is auto-calculated
          width: 400, aspect: 1,
          // require any image in 16:9 aspect
          aspect: 16 / 9
        }
      }
    }).onUploadComplete(() => { /** update user session */ })
  } satisfies FileRouter
  ```
- [x] add example
- https://github.com/pingdotgg/uploadthing/assets/51714798/5d250f60-a21d-4170-a46d-b703a40395f6

  - [x] fix svgs
  - handle upscaling if image is smaller than specified??
- [ ] Package it up to a nice re-usable component


examples/profile-picture/src/uploadthing/server.ts
```diff
@@ -0,0 +1,56 @@
...
+  profilePicture: f({
+    image: {
+      maxFileSize: "4MB",
+      maxFileCount: 1,
+      additionalProperties: {
+        width: 200,
+        aspectRatio: 1,
+      },
+    },
+    video: {},
```
markflorkowski, line 28:
```suggestion
```

packages/shared/src/types.ts
```diff
@@ -98,27 +103,73 @@ export type ContentDisposition = (typeof ValidContentDispositions)[number];
...
+   * @remarks These properties are not validated on the server on upload
+   */
+  additionalProperties?: AdditionalProperties<TAdditionalProperties>;
 };
 
 export type FileRouterInputKey = AllowedFileType | MimeType;
 
 export type ExpandedRouteConfig = {
-  [key in FileRouterInputKey]?: RouteConfig;
+  [key in FileRouterInputKey]?: key extends `image${string}`
```
markflorkowski, line 163: This will get messy as we start to add properties to other types 😬

markflorkowski (approved):

examples/profile-picture/src/app/_components/profile-pic-uploader.tsx
```diff
@@ -0,0 +1,315 @@
...
+              {isUploading && <Loader2 className="mr-2 h-4 w-4 animate-spin" />}
+              Save
+            </Button>
+          )}
+        </div>
+      </CardFooter>
+    </Card>
+  );
+}
+
```
juliusmarminge, line 190: @markflorkowski From here on below are generic helper functions. Are they worthwhile exporting from uploadthing somewhere? Or should we keep it all one big example for now?

packages/shared/src/types.ts
```diff
@@ -98,27 +103,73 @@ export type ContentDisposition = (typeof ValidContentDispositions)[number];
...
+  /**
+   * Specify the aspect ratio of the image.
+   * @remarks If both width and height are specified, this will be ignored.
+   */
+  aspectRatio?: number;
+};
+
+type AdditionalProperties<T> = Record<string, unknown> & T;
+
+type RouteConfig<TAdditionalProperties extends Record<string, unknown>> = {
```
juliusmarminge, line 120: Sidequest I remembered when working here - It'd be cool to have more jsdoc driven API reference documentation. Jsdoc should be the driver for the api reference documentation and we shouldnt have to copy-paste stuff in multiple places

examples/profile-picture/src/app/_components/profile-pic-uploader.tsx
markflorkowski: I think leaving them as part of the example is good for now. Makes sense to export alongside the components once we ship them. 

`expandImageProperties` might make sense to export from `@uploadthing/shared`  now though. (feels similar to eg. `generatePermittedFileTypes`)

t3dotgg (approved):
Simpler than expected by line count, lgtm!

PR #801, opened 2024-05-12
refactor:  UploadButton

juliusmarminge:
More changes brought over from #746 but in a non-breaking, less chaotic manner.

Goal to make #802 easier


docs/src/pages/theming.mdx
```diff
@@ -242,6 +242,7 @@ component match your design in the exact way you want.
 ```
 
 <UploadButton
+  endpoint="mockRoute"
```
juliusmarminge, line 245: having these actually load instead of fetch failing helped reduce some impossible states we had weird paths for just to display in here

packages/react/src/components/button.tsx
```diff
@@ -140,95 +135,96 @@ export function UploadButton<
...
+  const renderButton = () => {
+    const customContent = contentFieldToContent(
+      $props.content?.button,
+      styleFieldArg,
+    );
+    if (customContent) return customContent;
 
-    return "uploading";
-  })();
+    if (state === "readying") {
```
markflorkowski, line 206: nit: inline return

packages/react/src/components/button.tsx
```diff
@@ -140,95 +135,96 @@ export function UploadButton<
...
+    }
+
+    if (state !== "uploading") {
+      if (mode === "manual" && files.length > 0) {
+        return `Upload ${files.length} file${files.length === 1 ? "" : "s"}`;
+      }
+      return `Choose File${inputProps.multiple ? `(s)` : ``}`;
+    }
+
+    if (uploadProgress === 100) {
```
markflorkowski, line 217: nit: inline return

packages/react/src/components/button.tsx
```diff
@@ -261,7 +257,7 @@ export function UploadButton<
       data-ut-element="allowed-content"
     >
       {contentFieldToContent($props.content?.allowedContent, styleFieldArg) ??
-        allowedContentTextLabelGenerator(permittedFileInfo?.config)}
+        allowedContentTextLabelGenerator(routeConfig)}
```
markflorkowski, line 260: this is so much cleaner than permittedFileInfo 😅

markflorkowski (approved):

PR #810, opened 2024-05-15
fix: better mime type support for generic types

juliusmarminge:
...

### DIFF
--- a/packages/shared/src/component-utils.ts
+++ b/packages/shared/src/component-utils.ts
@@ -14,7 +14,8 @@ export const generateMimeTypes = (fileTypes: string[]) => {
     if (type === "pdf") return "application/pdf";
     if (type.includes("/")) return type;
     return [
-      `${type}/*`, // Add wildcard to support all subtypes, e.g. image => "image/*"
+      // Add wildcard to support all subtypes, e.g. image => "image/*"
+      `${type}/*`,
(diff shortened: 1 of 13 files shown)


packages/shared/src/component-utils.ts
```diff
@@ -2,21 +2,25 @@ import type { CSSProperties, ReactNode } from "react";
...
+    return [
+      // Add wildcard to support all subtypes, e.g. image => "image/*"
+      `${type}/*`,
+      // But some browsers/OSes don't support it, so we'll also dump all the mime types
+      // we know that starts with the type, e.g. image => "image/png, image/jpeg, ..."
+      ...objectKeys(mimeTypes).filter((t) => t.startsWith(type)),
+    ].join(", ");
   });
-
-  if (accepted.includes("blob")) {
```
markflorkowski, line 16: I think this case is still needed?

markflorkowski: nvm, I see it moved up, missed that at first

juliusmarminge: i early-returned this case first line in the function

markflorkowski (approved):
Why did tests break...?

juliusmarminge:
hmm passes locally for me 🤔

PR #813, opened 2024-05-16
fix: Node 20.13 compat

juliusmarminge:
Seems like something in Node 20.13 is broken with the spread operator on `Headers`:

image

Symbols aren't enumerable so they shouldn't show up after a spread, yet they are...

Anyways, this refactor makes it compat with this new behavior, be it intentional or unintentional from their side...


juliusmarminge:
I don't see how this can be an intentional change on Undici's side:

CleanShot 2024-05-16 at 12 00 10

PR #821, opened 2024-05-21
fix: invalid schema

juliusmarminge:
Fixes invalid schema.

Also adds a test suite that runs against a live app (if `UPLOADTHING_TEST_SECRET` is defined) to test a good chunk of the happy paths

### DIFF
--- a/packages/uploadthing/src/sdk/index.ts
+++ b/packages/uploadthing/src/sdk/index.ts
@@ -234,9 +234,12 @@ export class UTApi {
     guardServerOnly();
     const { keyType = this.defaultKeyType } = opts ?? {};
 
-    const responseSchema = S.Struct({
+    class DeleteFileResponse extends S.Class<DeleteFileResponse>(
+      "DeleteFileResponse",
+    )({
       success: S.Boolean,
-    });
+      deletedCount: S.Number,
+    }) {}
 
     return await this.executeAsync(
       this.requestUploadThing(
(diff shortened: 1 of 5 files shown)


packages/uploadthing/src/sdk/index.ts
```diff
@@ -234,17 +234,20 @@ export class UTApi {
     guardServerOnly();
     const { keyType = this.defaultKeyType } = opts ?? {};
 
-    const responseSchema = S.Struct({
+    class DeleteFileResponse extends S.Class<DeleteFileResponse>(
```
juliusmarminge, line 237: changed to this syntax just to get named types when hovering:

CleanShot 2024-05-21 at 09 42 06

packages/uploadthing/test/__test-helpers.ts
```diff
@@ -142,8 +118,35 @@ export const it = itBase.extend({
...
+       * Static Assets
+       */
+      http.get("https://cdn.foo.com/:fileKey", async ({ request }) => {
+        await callRequestSpy(request);
+        return HttpResponse.text("Lorem ipsum doler sit amet");
+      }),
+      http.get("https://utfs.io/f/:key", async ({ request }) => {
+        await callRequestSpy(request);
+        return HttpResponse.text("Lorem ipsum doler sit amet");
+      }),
```
juliusmarminge, line 146: moves here so that `import { it } from "vitest"` runs tests without __any__ mocks at all. `import { it } from "__test-helpers"` will run test with mocks

markflorkowski (approved):

Issue #822, opened 2024-05-21
docs: add automated testing guides


juan-carlos-correa:
### Area of Improvement

I would like to see automated testing guides in the documentation since automation testing is a critical part of software development.

There are testing documentation examples like Next.js itself: https://nextjs.org/docs/app/building-your-application/testing

**Proposal:**

Probably a good place to add it is something like: https://docs.uploadthing.com/testing

And add a playwright example like: https://docs.uploadthing.com/testing/playwright

So we can have more testing examples in the future using other testing tools.

What do you think? 

P.D: I'm willing to contribute adding a Playwright integration test example that I have developed for my personal project in Next.js with app router.

### Link to related docs

https://docs.uploadthing.com/

### Additional information

I don't see a proper link to related docs, so I put the root domain.

### 👨‍👧‍👦 Contributing

- [X] 🙋‍♂️ Yes, I'd be down to file a PR implementing the suggested changes!

markflorkowski:
I am definitely open to having docs on testing, and would welcome a PR for it! I believe we have had a similar request in the past, but just have not gotten around to it.

juan-carlos-correa:
> I am definitely open to having docs on testing, and would welcome a PR for it! I believe we have had a [similar request in the past](https://github.com/pingdotgg/uploadthing/issues/548), but just have not gotten around to it.

I can contribute it with a PR, WIP 🤓

juan-carlos-correa:
> I am definitely open to having docs on testing, and would welcome a PR for it! I believe we have had a [similar request in the past](https://github.com/pingdotgg/uploadthing/issues/548), but just have not gotten around to it.

Btw, I'm trying to run the docs in my local (I followed the contribution steps), and I'm having this error:

<img width="979" alt="image" src="https://github.com/pingdotgg/uploadthing/assets/10777626/0cb2e861-f98b-432c-b1d9-c53c42d991cf">

I'm not very familiar with turborepo, but I found a reference of that path here: `packages/uploadthing/package.json`, this part of the json:

```
 "./tw": {
      "import": {
        "types": "./tw/index.d.ts",
        "default": "./tw/index.js"
      },
      "require": {
        "types": "./tw/index.d.cts",
        "default": "./tw/index.cjs"
      }
    },
```

Before of making changes I would like to know if this is truly the source of the error, or if I should do something different. 

Possibly questions:

- Version of Node and Pnpm?
Answer: I'm using Node and Pnpm LTS versions.

[...]

PR #823, opened 2024-05-22
docs: add testing documentation for Playwright

juan-carlos-correa:
Ref: #822 

Initial testing example documentation using Playwright.

### DIFF
--- a/docs/src/pages/testing/playwright.mdx
+++ b/docs/src/pages/testing/playwright.mdx
@@ -1,0 +1,82 @@
+import { Callout } from "nextra-theme-docs";
+
+# Testing with Playwright
+
+Playwright is a end-to-end testing tool that allows you to test your web
+applications in a browser.
+
+We can use Playwright to create **integration tests** for our `UploadThing`
+funcionality, independently of the frontend framework we are using.
+
+<Callout type="info">
+  Integration tests are used to test the interaction between different parts of
+  your application. In the case of `UploadThing`, we can use Playwright to test
+  the interaction between the frontend and the backend.
+</Callout>
+
+## Installation
+
+If you haven't already, install Playwright by running:
+
+```bash
+npm install -D playwright
+```
+
+## Writing the test
+
+When writing your test, you can think of it as a user interacting with your
+application. For example, a real user would:
+
+1. Open the browser
+2. Go to the page where the file upload form is
+3. Upload a file
+4. Click the submit button
+5. Check if the file was uploaded successfully
+
+But instead of uploading a real file to `Uploadthing` cloud storage, we will
+mock the API endpoint and check if the frontend is sending the file correctly.
+
+```ts
+import { expect, test } from "@playwright/test";
+
+test("should upload a file successfully", async ({ page }) => {
+  // Mock the API endpoint, so we don't actually upload a file
+  // This is a Next.js API route, but you can replace it with your own
+  await page.route(
+    "*/**/api/uploadthing?actionType=upload&slug=courseImage",
+    async (route) => {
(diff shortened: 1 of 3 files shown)


docs/src/pages/testing/playwright.mdx
```diff
@@ -0,0 +1,82 @@
+import { Callout } from "nextra-theme-docs";
+
+# Testing with Playwright
+
+Playwright is a end-to-end testing tool that allows you to test your web
+applications in a browser.
+
+We can use Playwright to create **integration tests** for our `UploadThing`
+funcionality, independently of the frontend framework we are using.
```
markflorkowski, line 9:
```suggestion
functionality, independently of the frontend framework we are using.
```

docs/src/pages/testing/playwright.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+<Callout type="info">
```
markflorkowski, line 11: Not sure that we necessarily have to go to this level of explanation. I'm ok to leave it in, but cc @t3dotgg

docs/src/pages/testing/playwright.mdx
```diff
@@ -0,0 +1,82 @@
...
+But instead of uploading a real file to `Uploadthing` cloud storage, we will
+mock the API endpoint and check if the frontend is sending the file correctly.
+
+```ts
+import { expect, test } from "@playwright/test";
+
+test("should upload a file successfully", async ({ page }) => {
+  // Mock the API endpoint, so we don't actually upload a file
+  // This is a Next.js API route, but you can replace it with your own
+  await page.route(
```
markflorkowski, line 45: This feels like the wrong place to be mocking, since the api code is also provided by uploadthing. 

might make more sense to be mocking the API calls to uploadthing.com/api that the server code does, but even still I am having difficulty understanding what you are trying to do here.

juan-carlos-correa: Makes sense to me, specially if the example is suppose to be FE framework agnostic

markflorkowski: You may want to take a look at how we are doing our mocks with MSW for some inspiration there. 

https://github.com/pingdotgg/uploadthing/blob/3376306731d34c36fea8df3c6b117eeb17328f0f/packages/uploadthing/test/__test-helpers.ts

juan-carlos-correa: Hey there, 

It result that it's pretty complicated to mock the uploadthing endpoints because of the nature of Next.js. Explanation: 

1. Next.js is fullstack framework.
2. Playwright is a E2E testing tool.
3. Playwright can only mocks the network requests that are made in the client side.
4. The Uploadthing library for Next.js approach is to create a Next.js api routes where internally is the logic and rules in order to actually fetch de files to the Uploadthing APIs. 
5. The Uploadthing internal implementation is executed in the server side, not exposed to the client.

Therefore, I was not able to mock the real Uploadthing endpoints 😞 

So probably having the example mocking the Next.js API route is the only way, so far. 

What I have seen is a [combination of Mock Service Worker with Playwright,](https://stackoverflow.com/questions/70980370/how-do-i-mock-server-side-api-calls-in-a-nextjs-app) but it looks more complex than I would like to see when working with testing.

What are your thoughts?

t3dotgg:
IMO, integration testing of dependencies like this is not a great practice. The effort to "make it work" doesn't seem worthwhile relative to the potential safety and peace of mind after implementing.

If you've gotten actual value from including tests like this with your usage of UploadThing, let me know and I'll reconsider

PR #824, opened 2024-05-22
test: fix e2e test and add to matrix

juliusmarminge:
#823 reminded me that we have one of these setup - might as well run it in CI?

### DIFF
--- a/.github/workflows/examples.yaml
+++ b/.github/workflows/examples.yaml
@@ -50,5 +50,5 @@ jobs:
         if: ${{ !cancelled() && matrix.dir == 'backend-adapters' }}
         with:
           name: playwright-report
-          path: ${{ matrix.dir}}/playwright-report/
+          path: ./examples/${{ matrix.dir}}/playwright-report/
           retention-days: 7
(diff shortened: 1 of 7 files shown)


.github/workflows/examples.yaml
```diff
@@ -32,3 +33,15 @@ jobs:
...
+        run: |
+          cat ./examples/${{ matrix.dir }}/package.json | 
+          if grep --silent '"@playwright/test"';
+            then pnpm dlx playwright install --with-deps;
+          fi
+
+      - name: Run E2E Tests (if exists)
+        run: pnpm turbo --filter ./examples/${{ matrix.dir }} test
+        env:
+          UPLOADTHING_SECRET: ${{ secrets.UPLOADTHING_TEST_SECRET }}
```
juliusmarminge, line 47: might want a different secret here linked to another app. 

there could be race conditions if these tests runs at the same time

markflorkowski: Agreed. We should also put these apps in an org so we can all access them easily. 

```suggestion
          UPLOADTHING_SECRET: ${{ secrets.UPLOADTHING_E2E_TEST_SECRET }}
```

PR #837, opened 2024-05-29
chore: migrate vue off of `permittedFileInfo`

markflorkowski:


packages/vue/src/useUploadThing.ts
```diff
@@ -129,7 +126,13 @@ export const INTERNAL_uploadthingHookGen = <
...
       startUpload,
       isUploading,
-      permittedFileInfo,
+      routeConfig,
+      /**
+       * @deprecated Use `routeConfig` instead
+       */
+      permittedFileInfo: routeConfig
+        ? { slug: endpoint, config: routeConfig }
+        : undefined,
```
juliusmarminge, line 135: think this must be a `computed` to be updated

markflorkowski: I think you are correct , will fix

packages/vue/src/useUploadThing.ts
```diff
@@ -129,8 +126,15 @@ export const INTERNAL_uploadthingHookGen = <
...
       startUpload,
       isUploading,
-      permittedFileInfo,
-    } as const;
+      routeConfig,
+      /**
+       * @deprecated Use `routeConfig` instead
+       */
+      permittedFileInfo: computed(() => {
+        if (!routeConfig) return undefined;
```
juliusmarminge, line 134: is this not a ref so this will always be defined? should be `routeConfig.value`?

```suggestion
        if (!routeConfig.value) return undefined;
```

juliusmarminge (approved):

PR #838, opened 2024-05-30
feat: upload abortion

juliusmarminge:
allow user to abort upload

This PR
- [x] React 

Follow up with another PR adding this to
- [ ] Vue
- [ ] Solid
- [ ] Svelte


packages/react/src/useUploadThing.ts
```diff
@@ -85,6 +86,7 @@ export const INTERNAL_uploadthingHookGen = <
       opts?.onUploadProgress?.(0);
       try {
         const res = await uploadFiles<TEndpoint, TSkipPolling>(endpoint, {
+          signal: opts?.signal,
```
markflorkowski, line 89: maybe rename this to abortSignal? I could imagine this being confusing with such a generic name

packages/react/src/components/button.tsx
```diff
@@ -216,7 +232,12 @@ export function UploadButton<
 
     if (uploadProgress === 100) return <Spinner />;
 
-    return <span className="z-50">{uploadProgress}%</span>;
+    return (
+      <span className="z-50">
```
markflorkowski, line 236: Apply this to dropzone.tsx#L245

packages/react/src/useUploadThing.ts
```diff
@@ -114,6 +116,12 @@ export const INTERNAL_uploadthingHookGen = <
         await opts?.onClientUploadComplete?.(res);
         return res;
       } catch (e) {
+        /**
+         * This is the only way to introduce this as a non-breaking change
+         * Consider refactoring API in the next major version
```
markflorkowski, line 121:
```suggestion
         * TODO: Consider refactoring API in the next major version
```

Just makes it easier to find these

packages/uploadthing/src/internal/handler.ts
```diff
@@ -381,7 +381,7 @@ const handleUploadAction = Effect.gen(function* () {
     const fetchContext = yield* FetchContext;
     promise = Effect.forEach(
       presignedUrls,
-      (file) => conditionalDevServer(file.key, opts.apiKey),
+      (file) => conditionalDevServer(file.key, opts.apiKey).pipe(Effect.either),
```
markflorkowski, line 384: ?

packages/react/src/useUploadThing.ts
juliusmarminge: hmm `fetch` takes `signal`, I feel like it's quite a well-used term especially when it comes to interruption? i guess it can be confused with reactive signals though 🤔

markflorkowski: Yeah, that's what I was thinking. Maybe it is ok though since it is what `fetch` uses

packages/uploadthing/src/internal/handler.ts
juliusmarminge: makes it so the dev server doesn't kill the process if it times out in case of an upload abortion. should probably be a nicer error message though

packages/react/src/components/button.tsx
```diff
@@ -116,6 +119,7 @@ export function UploadButton<
   const { startUpload, isUploading, routeConfig } = useUploadThing(
     $props.endpoint,
     {
+      signal: $props.signal ?? acRef.current.signal,
```
juliusmarminge, line 122: Q: Should we omit the `signal` from the props for built-in components or allow them to pass an `AbortController` instead?

Not sure what the usage will look like but as it is now passing in a signal as prop will break the built-in cancellation mechanism. Or should we disable the "cancel on hover" behavior on the button if they pass in a signal?

I'm not sure what the right path is here... I think I'd be fine not allowing custom AbortSignal's for built-in components?

markflorkowski: > I think I'd be fine not allowing custom AbortSignal's for built-in components?

I am inclined to agree. If you need to do something more complex than the cancellation logic we provide, it makes sense that you build your own components.

juliusmarminge: https://github.com/pingdotgg/uploadthing/pull/838/commits/d6ebba061480185e62ad23f464b1abf96c5477fd

markflorkowski (approved):
lgtm, need to add to the docs for the hook

docs/src/pages/api-reference/react.mdx
```diff
@@ -407,6 +410,8 @@ const { startUpload, isUploading, permittedFileInfo } = useUploadThing(
 | onUploadProgress       | function                                             | No       | Added in `v5.1` | callback function for upload progress                                                                                       |
 | onBeforeUploadBegin    | `(files: File[]) => File[]`                          | No       | Added in `v5.8` | callback function called before uploading starts. The files returned are the files that will be uploaded                    |
 | onUploadBegin          | function                                             | No       | Added in `v5.4` | callback function for upload begin                                                                                          |
+| onUploadAborted        | function                                             | No       | Added in `v6.7` | callback function when upload is aborted                                                                                    |
```
juliusmarminge, line 413: this isn't avail on hook - just components.

callback here would be semver breaking, we need to throw the `startUpload` fn i think?

markflorkowski: Now that I think about it, it makes sense that we wouldn't handle that in the hook

PR #849, opened 2024-06-05
fix(docs): add x-robots-tag header on non-production environments

requestmethod:
Attempts to fix: https://x.com/t3dotgg/status/1798132183670141285

This PR adds the X-Robots-Tag header to try to prevent indexing of Vercel environments that aren't production. 

More info:
* https://developers.google.com/search/docs/crawling-indexing/robots-meta-tag
* https://www.bing.com/webmasters/help/which-robots-metatags-does-bing-support-5198d240

### DIFF
--- a/docs/next.config.mjs
+++ b/docs/next.config.mjs
@@ -37,6 +37,21 @@ const withNextra = nextra({
       },
     ];
   },
+  async headers() {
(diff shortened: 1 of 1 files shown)


requestmethod:
Feel free to take this over or close. I don't have access to test on preview.

PR #860, opened 2024-06-16
fix: `onUploadProgress` events out of order

### LINKED ISSUE #859: onUploadProgress of useUploadThing going back to previous value

```bash
System:
    OS: macOS 14.4.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 10.08 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 21.7.1 - /opt/homebrew/bin/node
    npm: 10.8.1 - /opt/homebrew/bin/npm
    pnpm: 9.1.4 - ~/Desktop/Code/blendcopycopy/node_modules/.bin/pnpm
  Browsers:
    Chrome: 125.0.6422.142
    Edge: 114.0.1823.58
    Safari: 17.4.1
  npmPackages:
    @uploadthing/react: ^6.6.0 => 6.6.0 
    typescript: ^5 => 5.4.5 
    uploadthing: ^6.12.0 => 6.12.0
```

### Describe the bug

When uploading multiple files using useUploadThing() hook and displaying the progress to the user, the progress number goes back. I assume it's because the onUploadProgress callback is called whenever the status of a file changes 

I'm expecting two possible things: 
- Having the progress for all files so if I have 10 files of 10mb for 100mb total, I'd expect to see 50% when 50mb have been uploaded.
- Having the progress mapped to each file. That is if the callback function would return an array of files but shouldn't be the case (however would be great to have the progress of each file in a future version) as when looking at the useUploadThingProps, it's set as:  
```tsx  
onUploadProgress?: ((p: number) => void) | undefined
``` 
My guess is the callback is being called on every file update and because files upload is happening at the same time, a smaller file can be at 80% while other at 20%.

### Link to reproduction

couldn't start the localhost due to error   ⨯ Failed to load SWC binary for linux/x64, see more info here: https://nextjs.org/docs/messages/failed-loading-swc

### To reproduce

This is what I use in my app to make it happen

```tsx
const [uploadStatus, setUploadStatus] = React.useState<string>("Hasn't started");
const { startUpload, isUploading } = useUploadThing(
    "multifilesUploader",
    {
        onUploadProgress: (progress: number) => {
            if (progress !== 100) {
                setUploadStatus("In progress: " + progress + "%")
            } else {
                setUploadStatus("Finilizing upload...")
            }
        },
    },
);
```

### Additional information
CleanShot_2024-06-13_at_20 36 16

### 👨‍👧‍👦 Contributing

- [ ] 🙋‍♂️ Yes, I'd be down to file a PR fixing this bug!

### Code of Conduct

juliusmarminge:
Closes #859

prefilling the map fixes the average calculation as some items might not be uplaoding from the start

CleanShot 2024-06-16 at 11 24 08@2x

### DIFF
--- a/.changeset/wild-terms-know.md
+++ b/.changeset/wild-terms-know.md
@@ -1,0 +1,8 @@
+---
+"@uploadthing/react": patch
+"@uploadthing/solid": patch
+"@uploadthing/svelte": patch
+"@uploadthing/vue": patch
+---
+
+fix: `onUploadProgress` events out of order when uploading many files
--- a/examples/minimal-appdir/src/app/page.tsx
+++ b/examples/minimal-appdir/src/app/page.tsx
@@ -11,8 +11,19 @@ export default function Home() {
     /**
      * @see https://docs.uploadthing.com/api-reference/react#useuploadthing
      */
-    onClientUploadComplete: () => {
-      alert("Upload Completed");
+    skipPolling: true,
+    onBeforeUploadBegin: (files) => {
+      console.log("Uploading", files.length, "files");
+      return files;
+    },
+    onUploadBegin: (name) => {
+      console.log("Beginning upload of", name);
+    },
+    onClientUploadComplete: (res) => {
+      console.log("Upload Completed.", res.length, "files uploaded");
+    },
+    onUploadProgress(p) {
+      console.log("onUploadProgress", p);
     },
   });
 
@@ -50,14 +61,14 @@ export default function Home() {
       />
       <input
         type="file"
+        multiple
         onChange={async (e) => {
-          const file = e.target.files?.[0];
-          if (!file) return;
+          const files = Array.from(e.target.files ?? []);
 
           // Do something with files
 
           // Then start the upload
-          await startUpload([file]);
+          await startUpload(files);
         }}
       />
     </main>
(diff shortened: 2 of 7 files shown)


juliusmarminge:
This calculation does seem a bit sus though and not very accurate calculating the average progress like this:
https://github.com/pingdotgg/uploadthing/pull/860/files#diff-d60d183d4e527ceee4ad5a7201e868c9a872add70f50bdb4755c50e63b8fa74cR78-R79

If you upload 1 large file and 1 small file the small will upload immediately and bump the progress to 50% even though the large file might not be close to done yet.

marclelamy:
Instead of calculating the total progress, would it be possible to return an array of all files along with their progress? Or maybe generate an event for each file's percent completion? It would be awesome to have a progress bar next to each file with the current progress displayed. I really like the logs you have in your console!

CleanShot 2024-06-13 at 20 36 16

markflorkowski (approved):

juliusmarminge:
> It would be awesome to have a progress bar next to each file (from image below) with the current progress displayed. I really like the logs you have in your console!

A change like that would require changing the signature which would be a breaking change. Even adding a second argument with the map would technically be a breaking change if you're really nitpicky. 

We are working on v7, so file a feature request for that and I'm sure we can consider it for then next major!

PR #861, opened 2024-06-17
feat: `@effect/platform` adapter

juliusmarminge:


packages/uploadthing/src/effect-http.ts
```diff
@@ -0,0 +1,114 @@
...
+  );
+  const getBuildPerms = buildPermissionsInfoHandler<TRouter>(opts);
+
+  return Http.router.empty.pipe(
+    Http.router.get(
+      "/",
+      Http.response.json(getBuildPerms(), {
+        headers: Http.headers.fromInput({
+          "x-uploadthing-version": UPLOADTHING_VERSION,
+        }),
```
juliusmarminge, line 53: can we provide this on a router level? All responses we return from this router should have this header. we're currently setting it 3 times

examples/backend-adapters/server/src/effect.ts
```diff
@@ -1,46 +1,59 @@
-import { HttpServer } from "@effect/platform";
-import { BunHttpServer, BunRuntime } from "@effect/platform-bun";
-import { Effect, Layer } from "effect";
+import "dotenv/config";
```
datner, line 1: I don't think this is actually needed tbh

examples/backend-adapters/server/src/effect.ts
```diff
@@ -1,46 +1,59 @@
...
+        status: 204,
+        headers: Http.headers.fromInput(corsHeaders),
+      });
+    }
+
+    const response = yield* app;
+
+    return response.pipe(Http.response.setHeaders(corsHeaders));
+  }),
+);
```
datner, line 42: Every time I forget how simple cors is 🤦🏻

examples/backend-adapters/server/src/effect.ts
```diff
@@ -1,46 +1,59 @@
...
-  BunRuntime.runMain,
+const app = router.pipe(
+  cors,
+  Http.server.serve(Http.middleware.logger),
+  Http.server.withLogAddress,
 );
+
+const port = +(process.env.PORT || "3000");
+
+const ServerLive = NodeHttpServer.server.layer(() => createServer(), { port });
```
datner, line 57:
```ts
const Port = Config.integer("PORT").pipe(Config.withDefault(3000))
const ServerLive = Layer.unwrapEffect(
  Effect.map(Port, port => NodeHttpServer.server.layer(() => createServer(), { port })
)

File: packages/uploadthing/src/effect-http.ts [hunk 5]
```diff
@@ -0,0 +1,114 @@
...
+) => createBuilder<MiddlewareArgs, TErrorShape>(opts);
+
+export const createRouteHandler = <TRouter extends FileRouter>(
+  opts: RouteHandlerOptions<TRouter>,
+): Http.router.Router<Http.body.BodyError, never> => {
+  incompatibleNodeGuard();
+
+  const requestHandler = buildRequestHandler<TRouter, MiddlewareArgs>(
+    opts,
+    "effect-http",
```
datner, line 43: note that `effect-http` [is a real thing](https://github.com/sukovanej/effect-http), we refer to the more low-level http server utilities as `@effect/platform`

packages/uploadthing/src/effect-http.ts
datner: Yes
```ts
const addUploadThingHeader = Http.middleware.make(
  Effect.map(Http.response.setHeader("x-uploadthing-version", UPLOADTHING_VERSION)),
);

Http.router.empty.pipe(
  // ...
  Http.router.use(addUploadThingHeader)
)
```
Technically `middleware.make` is just an `identity` type helper.
a `Middleware` is `<E,R>(app: App.Default<E,R>) => App.Default<any, any>`, and `App.Default` is just an alias for `Effect<ServerResponse, E,R>`. Meaning the above is just shorthand.

The `.use` would affect everything within the given `Router`. In a pipe remember that the composition is technically inverse so everything above the `.use` would be affected but not anything below it.
```ts
Http.router.empty.pipe(
  Http.router.get("a", A),
  Http.router.get("b", B),
  Http.router.get("c", C),
  Http.router.use(addUploadThingHeader)
)

// is literally
Http.router.use(addUploadThingHeader)(
  Http.router.get("c", C)(
    Http.router.get("b", B)(
      Http.router.get("a", A)(
        Http.router.empty
      )
    )
  )
)
```

packages/uploadthing/src/effect-http.ts
```diff
@@ -0,0 +1,114 @@
...
+    ),
+    Http.router.post(
+      "/",
+      Effect.flatMap(Http.request.ServerRequest, (req) =>
+        requestHandler({
+          /**
+           * TODO: Redo this to be more cross-platform
+           * This should handle WinterCG and Node.js runtimes,
+           * unsure about others...
+           * Perhaps we can use `Http.request.ServerRequest` internally?
```
datner, line 64: This would make it cross-platform indeed 💯

packages/uploadthing/src/effect-http.ts
```diff
@@ -0,0 +1,114 @@
...
+          Effect.provideService(FetchContext, {
+            fetch: opts.config?.fetch ?? globalThis.fetch,
+            baseHeaders: {
+              "x-uploadthing-version": UPLOADTHING_VERSION,
+              // These are filled in later in `parseAndValidateRequest`
+              "x-uploadthing-api-key": undefined,
+              "x-uploadthing-be-adapter": undefined,
+              "x-uploadthing-fe-package": undefined,
+            },
+          }),
```
datner, line 94: continuing from our `Client` implementation:
```ts
Effect.provideServiceEffect(
  HttpClient.Client,
  HttpClient.mapRequest(
    HttpClient.fetch,
    ClientRequest.setHeader("x-uploadthing-version", UPLOADTHING_VERSION),
  ),
),
Effect.provideService(HttpClient.Fetch, opts.config?.fetch ?? globalThis.fetch),
```


Note though, currently this means that you must use `fetch`, `Undici` and `Agent`s are not supported. But you might want to support them

docs/src/pages/api-reference/server.mdx
```diff
@@ -6,7 +6,7 @@ The helper function to create an UploadThing instance. MAKE SURE YOU IMPORT IT
 FROM THE RIGHT PLACE. The export name ensures your file routes' `middleware`
 functions are typed correctly.
 
-<Tabs items={["Next App Router", "Next Pages Dir", "SolidJS", "Express", "Fastify", "H3"]}>
+<Tabs items={["Next App Router", "Next Pages Dir", "SolidJS", "Effect HTTP","Express", "Fastify", "H3"]}>
```
datner, line 9: `Effect Platform`? idk, ask the other folk over at effect how they wanna officiate this

examples/backend-adapters/server/src/effect.ts
juliusmarminge: Tax doesnt load process.env automatically I dont think

examples/backend-adapters/server/src/effect.ts
juliusmarminge: Does effect/Config work with process.env, import.meta.env etc? Wondering if we could replace std-env with it?

packages/uploadthing/src/effect-http.ts
juliusmarminge: cool - the client stuff is on a separate branch so i'll add that in there.

what specifically is it that doesn't work with other fetch clients?

examples/backend-adapters/server/src/effect-platform.ts
```diff
@@ -0,0 +1,61 @@
...
+import { createRouteHandler } from "uploadthing/effect-platform";
+
+import { uploadRouter } from "./router";
+
+const uploadthingRouter = createRouteHandler({
+  router: uploadRouter,
+});
+
+/**
+ * Simple CORS middleware that allows everything...
```
markflorkowski, line 17:
```suggestion
 * Simple CORS middleware that allows everything
```

markflorkowski (approved):
It's very cool how small these adapters can be

PR #873, opened 2024-06-27
chore: Testing out pkg.pr.new

markflorkowski:


AmirSa12:
Happy to see you are considering pkg.pr.new! please let us know if we can help!! @markflorkowski

Aslemammad:
You can try --compact too! It'd make the urls way shorter (if all are available on npm though)

markflorkowski:
@AmirSa12 Any suggestions for filtering down to only packages that have been modified?

.github/workflows/pkg-pr-new.yaml
```diff
@@ -0,0 +1,26 @@
+name: Publish Test Releases
+on: [push, pull_request]
```
juliusmarminge, line 2: a) do you want this on every push?
b) what's the permissino story here? does this use secrets or not? will it work on forks?

.github/workflows/pkg-pr-new.yaml
```diff
@@ -0,0 +1,26 @@
...
+    timeout-minutes: 10
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Get changed packages
+        id: changed-packages
+        run: |
+          echo "::set-output name=changed_packages::$(git diff --name-only main ./packages | awk -F'/' '{print "./packages/" $2}' | sort -u | tr '\n' ' ')"
```
juliusmarminge, line 16: love a good bash oneliner

juliusmarminge (approved):

.github/workflows/pkg-pr-new.yaml
markflorkowski: I *think* it should work 😅 It seemed to work locally, but actions are weird

.github/workflows/pkg-pr-new.yaml
markflorkowski: a) I think so? That way it stays up to date through the life of a branch
b) no secrets, so I *think* it should work on forks. cc @AmirSa12

Aslemammad: a) You can do it the vite way where it only run on specific comments on a PR! But having things on each push would be good! what you might want to do is to run it on each push on main and only on approved PRs. 
b) yep, no secrets, but for forks, the user has to enable workflows for their fork plus having the app installed on the repo. So it's a matter of enabling workflows.

Aslemammad: To avoid workflow collisions, you can also do this! https://github.com/stackblitz-labs/pkg.pr.new/blob/main/.github/workflows/ci.yml#L2-L7 since the push and pull_request event might run two times at the same time

Aslemammad:
> Any suggestions for filtering down to only packages that have been modified?

I guess you already figured this out, let me know if you need help.

> Also is it possible to change the package manager used in the comment?

Hmm, no not currently, do you think we should add the possibility to change the package manager on the comment? We didn't do it since we thought this way we can keep the project minimal, but I'd like to add it if you want.

juliusmarminge:
I can also see it doesn't seem to respect `private: true` since the Raycast "package" got published

juliusmarminge:
It would also be nice to be able to link to one of our examples on stackblitz and have that automatically bumped to the released version

Aslemammad:
Nice catch, let me create an issue! https://github.com/stackblitz-labs/pkg.pr.new/issues/118

> It would also be nice to be able to link to one of our examples on stackblitz and have that automatically bumped to the released version

For sure, you can use --templates! See for instance how @maiieul in qwik-ui did it. https://github.com/qwikifiers/qwik-ui/pull/863/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R42-R43 

we also wrote some examples in the docs.

Aslemammad:
just released pkg-pr-new@0.0.10 btw!

markflorkowski:
@Aslemammad 

> I guess you already figured this out, let me know if you need help.

Yeah, I figured something out, I feel like this is a fairly common use though, so it may be good for ya'll to include in docs/possibly integrate into the tool (although that may be out of scope)

> Hmm, no not currently, do you think we should add the possibility to change the package manager on the comment? We didn't do it since we thought this way we can keep the project minimal, but I'd like to add it if you want.

Yeah, its not a huge thing, since I use pnpm, and can just type `p` then paste the command, but it would be nice if I didn't have to.

markflorkowski:
> It would also be nice to be able to link to one of our examples on stackblitz and have that automatically bumped to the released version

added the `minimal-` examples to it, still thinking on if/how to filter examples to only relevant packages for the branch

Aslemammad:
> Yeah, I figured something out, I feel like this is a fairly common use though, so it may be good for ya'll to include in docs/possibly integrate into the tool (although that may be out of scope)

Will add for sure, thank you so much! 

> still thinking on if/how to filter examples to only relevant packages for the branch

I think examples should not be filtered since they are there to test the packages, so you might want to have them always instead of seeing only the ones affected by the relevant changes. This way they'd also be always accessible to the users. But completely up to you.

PR #883, opened 2024-07-10
fix: upgrade scalar

marclave:
we have a few months worth of great work in here! specifically fixing the scrolling bugs, but we have a ton of other features too

https://github.com/pingdotgg/uploadthing/assets/6176314/85442e56-0bf0-41aa-a835-933328913f60

- OAuth 2.0 support
- Support FormData, XML, and many other body types
- Preview HTML, images and more in the response
- faster OpenAPI parsing (with our new parser)
- and more : )


juliusmarminge:
Hey @marclave 👋👋

thanks for this, looks great. I tried setting up some form of loader for when Scalar hasn’t finished mounting (just so you don’t see a blank white page. Had to do it with a MutationObserver and guessing what mutation was ”late enough” to get a somewhat accurate timing. What do you think of exposing like an `onReady` callback, or `skeleton` fallback component to give users a bit more control of the loading state?

markflorkowski (approved):

PR #889, opened 2024-07-13
docs: update year in license

Just-Moh-it:
https://youtu.be/IhnN_qKwDus?si=nBm2Qd0R3hrhr5Dl&t=188


dieperdev:
Someone got here before me 😄

PR #895, opened 2024-07-23
chore!: don't bundle tailwind-merge by default

juliusmarminge:
tailwind-merge is [20kb (7kB gzipped)](https://bundlejs.com/?q=tailwind-merge&treeshake=%5B%7B+twMerge+%7D%5D) that not everyone benefits from. It is only useful for folks applying additional tailwind classes when [theming using TailwindCSS](https://uploadthing-docs-twui.vercel.app/concepts/theming#theming-with-tailwind-css). They can thus provide us the merge function themselves.

This will:
- reduce client bundle by 7kB by default for everyone not using tailwind customizations
- reduce risk of duplicating tailwind-merge in projects already using it, in the case of a version range mismatch
- reduce package dependencies (if that counts as something positive :))

```tsx
import { twMerge } from "tailwind-merge"

<UploadButton 
  classList="bg-red-500"
  cn={twMerge}
/>
```

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -63,9 +63,7 @@
   },
   "dependencies": {
     "@uploadthing/dropzone": "workspace:*",
-    "@uploadthing/shared": "workspace:*",
-    "file-selector": "^0.6.0",
-    "tailwind-merge": "^2.2.1"
+    "@uploadthing/shared": "workspace:*"
   },
   "peerDependencies": {
     "next": "*",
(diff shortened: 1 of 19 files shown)


markflorkowski (commented):
lgtm, we should probably put a note in the styling docs about overrriding this if you are styling with tailwind.

juliusmarminge:
Yup I'll update the docs in the other repo

PR #886, opened 2024-07-12
feat: Add `onChange` + various feature parity additions/fixes

markflorkowski:
- Unifies the API between components
- Marks onDrop as deprecated. 

Additional changes:
- Fixes disabled styling [(Discord)](https://discord.com/channels/966627436387266600/1102510616326967306/1267082619762507849)
- Adds/fixes paste functionality on all current FE packages 
- Adds/fixes upload abort functionality on all current packages
- Tested functionality is equivalent across all current FE packages

image

PR #898, opened 2024-07-29
chore: init some component tests

juliusmarminge:
Goal: have some basic tests of state updates to "ensure" libs (react, solid, vue etc) stay in sync, and functionality doesn't drift apart

- any nice way to render the component from the different libs without duplicating all tests? i.e. run the test for `@uploadthing/react#UploadButton`, `@uploadthing/solid#UploadButton`, etc 
- where to prioritize testing? the goal is not to get 100% coverage, just test core functionality
  - states?
  - manual/auto mode?
  - pasting files?
  - theming? (to what degree??)
  - aborting? (might be hard given we'd need to mock a bunch of stuff)


markflorkowski:
>where to prioritize testing? the goal is not to get 100% coverage, just test core functionality ...

As I was working on #886 here is what I was going through manually:
- verify that components all have the same props (`onXXXX` functions, config, disabled, etc)
- verify styles match across libraries
  - including across various states
- verify copy matches across libraries  
- confirm components are reactive
  - selecting files changes the text on the button
  - uploading changes style and text on button
- confirm paste works
- confirm abort works

did not check but probably should have:
- confirm functions are triggered at the right times 
  - onChange should occur on select, clear, and drop
- custom styling overrides are available and functioning (and always apply to the same parts of the component)

Some of this would be nice to abstract away so that they are literally pulling the same implementations.

markflorkowski (approved):
For now, we will just have the one set of tests, but I would like to follow up post V7 with a more complete suite of component tests.

PR #917, opened 2024-08-21
fix: add `Callback Failed` response to list file response status

lamalex:


markflorkowski:
note: This requires infra changes as well to do anything.

.changeset/moody-rice-heal.md
```diff
@@ -0,0 +1,5 @@
+---
+"uploadthing": minor
```
markflorkowski, line 2: should probably be a patch since it is not adding any functionality, just fixing a missing type.

packages/uploadthing/src/sdk/index.ts
```diff
@@ -244,7 +244,7 @@ export class UTApi {
     )({
       success: S.Boolean,
       deletedCount: S.Number,
-    }) {}
+    }) { }
```
markflorkowski, line 247: what is up with this formatting?

markflorkowski (changes requested):

markflorkowski:
Closing in favor of updating the infra code to map those files to "Uploaded"

The "Callback Failed" was just added to help debugging by allowing people to see when their callbacks didn’t run as expected. It is only really useful from the dashboard, since there is nothing that can be done about it from a user perspective, and it just complicates any logic you may need to do while filtering that list (the files are uploaded and available in storage)

PR #921, opened 2024-08-23
fix: number serialization

markflorkowski:
Address issues with passing unserializable numbers to prepareUpload (eg. maxFileCount: Infinity)

https://discord.com/channels/966627436387266600/1102510616326967306/1276614012825767966

### DIFF
--- a/packages/react/src/next-ssr-plugin.tsx
+++ b/packages/react/src/next-ssr-plugin.tsx
@@ -3,6 +3,7 @@
 import { useId } from "react";
 import { useServerInsertedHTML } from "next/navigation";
 
+import { safeNumberReplacer } from "@uploadthing/shared";
 import type { EndpointMetadata } from "@uploadthing/shared";
 
 declare const globalThis: {
@@ -18,7 +19,7 @@ export function NextSSRPlugin(props: { routerConfig: EndpointMetadata }) {
   useServerInsertedHTML(() => {
     const html = [
       // Hydrate routerConfig on client globalThis
-      `globalThis.__UPLOADTHING = ${JSON.stringify(props.routerConfig)};`,
+      `globalThis.__UPLOADTHING = ${JSON.stringify(props.routerConfig, safeNumberReplacer)};`,
     ];
 
     return (
--- a/packages/react/src/next-ssr-plugin.tsx
+++ b/packages/react/src/next-ssr-plugin.tsx
@@ -3,7 +3,6 @@
 import { useId } from "react";
 import { useServerInsertedHTML } from "next/navigation";
 
-import { safeNumberReplacer } from "@uploadthing/shared";
 import type { EndpointMetadata } from "@uploadthing/shared";
 
 declare const globalThis: {
@@ -19,7 +18,7 @@ export function NextSSRPlugin(props: { routerConfig: EndpointMetadata }) {
   useServerInsertedHTML(() => {
     const html = [
       // Hydrate routerConfig on client globalThis
-      `globalThis.__UPLOADTHING = ${JSON.stringify(props.routerConfig, safeNumberReplacer)};`,
+      `globalThis.__UPLOADTHING = ${JSON.stringify(props.routerConfig)};`,
     ];
 
     return (
--- a/packages/shared/src/utils.ts
+++ b/packages/shared/src/utils.ts
@@ -39,6 +39,7 @@ export function getDefaultSizeForType(fileType: FileRouterInputKey): FileSize {
 
 /**
  * This function takes in the user's input and "upscales" it to a full config
+ * Additionally, it replaces numbers with "safe" equivalents
  *
  * Example:
  * ```ts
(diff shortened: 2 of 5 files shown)


packages/shared/src/utils.ts
```diff
@@ -79,6 +79,14 @@ export const fillInputRouteConfig = (
...
 
     newConfig[key] = { ...defaultValues, ...value };
+
+    if (newConfig[key].maxFileCount === Infinity) {
+      newConfig[key].maxFileCount = Number.MAX_SAFE_INTEGER;
+    }
+
+    if (newConfig[key].minFileCount === Infinity) {
+      newConfig[key].minFileCount = Number.MAX_SAFE_INTEGER;
+    }
```
juliusmarminge, line 89: these are unnecessary now right?

packages/shared/src/utils.ts
```diff
@@ -276,3 +284,26 @@ export const resolveMaybeUrlArg = (maybeUrl: string | URL | undefined): URL => {
...
+/**
+ * Replacer for JSON.stringify that will replace numbers that cannot be
+ * serialized to JSON with "reasonable equivalents".
+ *
+ * Infinity and -Infinity are replaced by MAX_SAFE_INTEGER and MIN_SAFE_INTEGER
+ * NaN is replaced by 0
+ *
+ */
+export const safeNumberReplacer = (_: string, value: unknown) => {
+  if (typeof value === "number") {
```
juliusmarminge, line 297: nit: prefer early return

packages/shared/src/utils.ts
```diff
@@ -276,3 +276,26 @@ export const resolveMaybeUrlArg = (maybeUrl: string | URL | undefined): URL => {
...
+ * Replacer for JSON.stringify that will replace numbers that cannot be
+ * serialized to JSON with "reasonable equivalents".
+ *
+ * Infinity and -Infinity are replaced by MAX_SAFE_INTEGER and MIN_SAFE_INTEGER
+ * NaN is replaced by 0
+ *
+ */
+export const safeNumberReplacer = (_: string, value: unknown) => {
+  if (typeof value !== "number") return value;
+  else {
```
juliusmarminge, line 290: But: Redundant  else

juliusmarminge (approved):

packages/uploadthing/src/internal/handler.ts
```diff
@@ -357,13 +358,16 @@ const handleUploadAction = Effect.gen(function* () {
...
     {
       method: "POST",
-      body: JSON.stringify({
-        files: filesWithCustomIds,
-        routeConfig: parsedConfig,
-        metadata,
-        callbackUrl: callbackUrl.origin + callbackUrl.pathname,
-        callbackSlug: opts.slug,
-      }),
+      body: JSON.stringify(
```
juliusmarminge, line 361: is this the only place we need to use this replacer? how bout when sending route config to client?

markflorkowski: Ah good catch

packages/react/src/next-ssr-plugin.tsx
```diff
@@ -18,7 +19,7 @@ export function NextSSRPlugin(props: { routerConfig: EndpointMetadata }) {
   useServerInsertedHTML(() => {
     const html = [
       // Hydrate routerConfig on client globalThis
-      `globalThis.__UPLOADTHING = ${JSON.stringify(props.routerConfig)};`,
+      `globalThis.__UPLOADTHING = ${JSON.stringify(props.routerConfig, safeNumberReplacer)};`,
```
markflorkowski, line 22: may cause mismatch?

juliusmarminge: Why if the fetch call also uses the replacer?

packages/uploadthing/src/server.ts
juliusmarminge, line 1: What about the other adapters?

Maybe it's better to handle this replacer earlier when parsing the route config?

markflorkowski: Like I was originally doing in `fillInputRouteConfig`?

juliusmarminge: Oh maybe 😅

Maybe like `JSON.parse(JSON.stringify(data, replacer))` when sanitizing the input 🤔

markflorkowski: image
?

juliusmarminge: 👍 that should avoid having to touch all adapters right?

markflorkowski: I believe so, yeah.

packages/react/src/next-ssr-plugin.tsx
juliusmarminge, line 22: i think this is redundant given both the GET endpoint and the `extractRouterCongi` (which is what you're meant to pass as props here) use `buildPermissionHandler` which in turn invokes `fillRouterInput...` which does the replacer - so it should be all valid JSON

juliusmarminge (approved):

packages/react/src/next-ssr-plugin.tsx
markflorkowski: You're right, I just forgot about it 😅

PR #924, opened 2024-08-26
feat: Add app-scoped urls to types for api returns

markflorkowski:
Update types to match returns (pingdotgg/uploadthing-infra#501)


packages/uploadthing/test/sdk.test.ts
```diff
@@ -584,14 +592,18 @@ describe.runIf(shouldRun)(
...
+      const key = result.data!.key;
       expect(result).toEqual({
         data: {
           customId: null,
           key: expect.stringMatching(/.+/),
           name: "foo.txt",
           size: 3,
           type: "text/plain",
-          url: expect.stringMatching(/https:\/\/utfs.io\/f\/.+/),
+          url: "https://utfs.io/f/" + key,
```
markflorkowski, line 603: Shouldn't this also be a `stringMatching`?

markflorkowski: Ah i guess it's not really necessary since there is no unknown part of the url to match.

PR #925, opened 2024-08-27
chore: use sqids instead of sha256 hashing

juliusmarminge:
Returns a key like `KJNGYKlZPq92Lq0BVqOBeV4QXKqyYlocntI7CxzdHiOfLwFm` instead of a 172ch long base64 string

### DIFF
--- a/packages/shared/src/crypto.ts
+++ b/packages/shared/src/crypto.ts
@@ -104,7 +104,7 @@ export const generateKey = (
       ],
     );
 
-    // Hash and Encode the parts and apiKey as hex strings
+    // Hash and Encode the parts and apiKey as sqids
     const alphabet = shuffle(defaultOptions.alphabet, apiKey);
     const encodedFileSeed = new SQIds({ alphabet, minLength: 36 }).encode([
       Math.abs(Hash.string(hashParts)),
--- a/packages/shared/src/crypto.ts
+++ b/packages/shared/src/crypto.ts
@@ -82,8 +82,6 @@ export const verifySignature = (
     Micro.orElseSucceed(() => false),
   );
 
-const KEY_SEPARATOR = "|";
-
 export const generateKey = (
   file: FileProperties,
   apiKey: string,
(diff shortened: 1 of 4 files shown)


packages/shared/src/crypto.ts
```diff
@@ -91,22 +105,27 @@ export const generateKey = (
...
+    const encodedFileSeed = new SQIds({ alphabet, minLength: 36 }).encode([
+      Math.abs(Hash.string(hashParts)),
+    ]);
+    const encodedApiKey = new SQIds({ alphabet, minLength: 12 }).encode([
+      Math.abs(Hash.string(apiKey)),
+    ]);
 
     // Concatenate them and encode as base64
-    return Encoding.encodeBase64([encodedApiKey, encodedFileSeed].join(":"));
+    return [encodedApiKey, encodedFileSeed].join(KEY_SEPARATOR);
```
juliusmarminge, line 117: Is this URL safe? I chose it cause it "looks" like a normal character so it doesn't stand out as much xd

could also base64 encode the key, but it does add to it's length

```sh
KJNGYKlZPq92|4wGwXUsC5WekfPrntiTE0qUyIo91O87zXwNg => 49ch
S0pOR1lLbFpQcTkyfDR3R3dYVXNDNVdla2ZQcm50aVRFMHFVeUlvOTFPODd6WHdOZw== 68ch
```

markflorkowski: Pipe is not urlsafe, could use dash?

markflorkowski: do they need to be separated?

juliusmarminge: i guess not really if we do a `startsWith` check instead.

I think the reason for the separation initially was adding a timing safe equals but that might be overkill?

markflorkowski: Yeah, I think it's fine if it isn't timing safe

PR #929, opened 2024-09-03
Add Convex adapter

sujayakar:
This PR adds a Convex adapter as a sibling to Express, Fastify, etc. It was pretty straightforward to add a new adapter!

Here's what it looks like for wiring it in to a Convex app. First, call `createUploadthing` as always:
```ts
// convex/http.ts
import { addUploadthingRoutes, convexCtx, createUploadthing, type FileRouter } from "uploadthing/convex"
import schema from "./schema"

const f = createUploadthing({
  errorFormatter: (err) => {
    console.log("Error uploading file", err.message);
    console.log("  - Above error caused by:", err.cause);

    return { message: err.message };
  },
  schema,
});
```
Note that the developer can optionally pass in their Convex `schema` here for more type-safety within their handlers. 

Then, define the router, using the `convexCtx` function to access the request's [current `ctx` value](https://docs.convex.dev/generated-api/server#actionctx). This is necessary for calling into other Convex functions, accessing auth and the database, etc. Under the hood, this uses a hack where we stash the `ctx` when starting a request.
```ts
export const uploadRouter = {
  imageUploader: f({ image: { maxFileSize: "4MB" } })
    .middleware(async (args) => {
      const ctx = convexCtx(f);      
      const identity = await ctx.auth.getUserIdentity();
      ...
      return { userId: identity?.subject ?? "nothing" }
    })
    .onUploadComplete(async (args) => {      
      const ctx = convexCtx(f);
      ...      
      return { uploadedBy: args.metadata.userId };
    }),
} satisfies FileRouter;
export type OurFileRouter = typeof uploadRouter;
```
Finally, install the routes into Convex's HTTP router with `addUploadthingRoutes`:
```ts
const http = httpRouter();
addUploadthingRoutes(http, f, { router: uploadRouter })
export default http;    
```

# Questions
- The hack to sneak in `ctx` via a "global" with `convexCtx` is a bit ugly. Is there a better way to pass values to the middleware and completion callbacks?
- Where should I add Convex to the docs? Would another entry next to https://docs.uploadthing.com/backend-adapters/express be the right spot?
- Should I add my Uploadthing+Convex example app from the snippets above as a new directory under `examples/`? It'll require setting up a Convex account (or self hosting Convex) to actually run the demo.
- Are there any tests, etc., that I should also be updating?

### DIFF
--- a/packages/uploadthing/package.json
+++ b/packages/uploadthing/package.json
@@ -99,6 +99,16 @@
         "default": "./h3/index.cjs"
       }
     },
+    "./convex": {
+      "import": {
+        "types": "./convex/index.d.ts",
+        "default": "./convex/index.js"
+      },
+      "require": {
+        "types": "./convex/index.d.cts",
+        "default": "./convex/index.cjs"
+      }
+    },
     "./types": {
       "types": "./types/index.d.ts",
       "default": "./types/index.js"
@@ -116,6 +126,7 @@
   },
   "files": [
     "client",
+    "convex",
     "effect-platform",
     "express",
     "fastify",
@@ -172,6 +183,7 @@
     "zod": "^3.23.8"
   },
   "peerDependencies": {
+    "convex": "*",
     "@effect/platform": "*",
     "express": "*",
     "fastify": "*",
@@ -190,6 +202,9 @@
       "optional": true
     },
     "fastify": {
+      "optional": true
+    },
+    "convex": {
       "optional": true
     },
     "h3": {
--- a/packages/uploadthing/src/convex.ts
+++ b/packages/uploadthing/src/convex.ts
@@ -58,13 +58,10 @@ export const addUploadthingRoutes = <TRouter extends FileRouter>(
       if (!isCorsRequest) {
         return new Response();
       }
-      if (!process.env.CLIENT_ORIGIN) {
-        throw new Error("Convex deployment doesn't have CLIENT_ORIGIN set");
-      }
       return new Response(null, {
         status: 200,
         headers: {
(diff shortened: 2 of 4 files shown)


examples/backend-adapters/server/src/convex.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import { httpRouter } from "convex/server";
+
+import {
+  addUploadthingRoutes,
+  createUploadthing,
+  FileRouter,
+} from "uploadthing/convex";
+
+// In a real app, you can wire up `cereateUploadthing` to your Convex
```
markflorkowski, line 12:
```suggestion
// In a real app, you can wire up `createUploadthing` to your Convex
```

examples/backend-adapters/server/src/convex.ts
```diff
@@ -0,0 +1,34 @@
...
+// schema for increased type safety.
+// ```ts
+// import schema from "./schema";
+// const f = createUploadthing({ schema })
+// ```
+const f = createUploadthing();
```
markflorkowski, line 18: It may be better to have a separate example showing the "real" way as you suggested in the PR comment. without it, this example doesnt really hold much value IMO

markflorkowski (approved):
This all looks fine, the adapter itself is supoer simple. 

 I would probably opt for a separate example that shows the actual way this is used, even if it does require a convex subscription or self hosting Convex. (Our clerk examples also require a subscription)

examples/backend-adapters/server/src/convex.ts
sujayakar: sounds good, I'll remove this and port it to a full example.

PR #936, opened 2024-09-12
chore: Import new docs site to monorepo

markflorkowski:
Replaces old docs site in monorepo. Legacy site will remain on the `legacy-docs` branch, and will remain accessible at v6.docs.uploadthing.com


markflorkowski:
Going to close in favor of basing off v7. Also rolling in some other changes made to private repo

PR #939, opened 2024-09-13
chore: deprecate old docs

markflorkowski:
removes sitemaps


docs/next-sitemap.config.js
```diff
@@ -1,5 +0,0 @@
-/** @type {import("next-sitemap").IConfig} */
```
t3dotgg, line 1: We should probably robots.txt ignore this as well, or whatever the hack we did was for preview builds

PR #946, opened 2024-09-14
fix: nextjs pages adapter response body

juliusmarminge:
`res.json(response.body)` resulted in `{}`.

couldn't find a way to forward the stream as-is, so consuming it and sending it down as JSON as a temporary workaround

### DIFF
--- a/packages/uploadthing/src/next-legacy.ts
+++ b/packages/uploadthing/src/next-legacy.ts
@@ -34,6 +34,7 @@ export const createRouteHandler = <TRouter extends FileRouter>(
 
   return async (req: NextApiRequest, res: NextApiResponse) => {
     const response = await handler(req, res);
+    res.status(response.status);
     for (const [name, value] of response.headers) {
       res.setHeader(name, value);
     }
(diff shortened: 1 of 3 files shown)


packages/uploadthing/src/next-legacy.ts
```diff
@@ -38,6 +38,7 @@ export const createRouteHandler = <TRouter extends FileRouter>(
     for (const [name, value] of response.headers) {
       res.setHeader(name, value);
     }
-    return res.json(response.body);
+    // FIXME: Should be able to just forward it instead of consuming it first
```
markflorkowski, line 41: Would be nice, but iI don't think this is a huge deal.

markflorkowski (approved):

packages/uploadthing/src/next-legacy.ts
juliusmarminge: nah lets just ship as-is and see if we ever find it high priority enough to optimize a little bit... the json blobs are never huge anyways so it's not like we're gonna load a bunch of stuff into memory

PR #947, opened 2024-09-14
feat: composable primitive components

veloii:
This PR adds unstyled composable components through `generateUploadPrimitives` which returns:
- `<Root>`
- `<Dropzone>`
- `<Button>`
- `<AllowedContent>`

I've also added examples for `minimal-appdir`, `minimal-pagedir`, `with-clerk-appdir`, `with-clerk-pagedir`. I can add it to all the rest of the react ones if needed.

All these components accept parameters as children like `({ isUploading }) => <></>`.
There is no `asChild` or `as={YourComponent}` prop at the moment as I'm not sure which one to use.
I haven't implemented the docs either as I'm not sure if the current styling page should be split up.

### DIFF
--- a/examples/minimal-appdir/src/app/page.tsx
+++ b/examples/minimal-appdir/src/app/page.tsx
@@ -4,6 +4,7 @@ import {
   UploadButton,
   UploadDropzone,
   useUploadThing,
+  UT,
 } from "~/utils/uploadthing";
 
 export default function Home() {
@@ -70,6 +71,19 @@ export default function Home() {
           await startUpload(files);
         }}
       />
+      <UT.Root endpoint="videoAndImage">
+        <UT.Dropzone>
+          {({ dropzone, isUploading }) => (
+            <>
+              <UT.Button>{isUploading ? "Uploading" : "Upload file"}</UT.Button>
+              <div>
+                <UT.AllowedContent />
+              </div>
+              {dropzone?.isDragActive && <span>Dragging</span>}
+            </>
+          )}
+        </UT.Dropzone>
+      </UT.Root>
     </main>
   );
 }
--- a/packages/react/src/components/primitive/allowed-content.tsx
+++ b/packages/react/src/components/primitive/allowed-content.tsx
@@ -1,0 +1,18 @@
+import { allowedContentTextLabelGenerator } from "@uploadthing/shared";
+
+import {
+  PrimitiveComponentChildrenProp,
+  PrimitiveSlot,
+  usePrimitiveValues,
+} from "./root";
+
+export function AllowedContent(props: PrimitiveComponentChildrenProp) {
+  const { routeConfig } = usePrimitiveValues("AllowedContent");
+
+  return (
+    <PrimitiveSlot
+      default={allowedContentTextLabelGenerator(routeConfig)}
+      children={props.children}
+    />
+  );
+}
--- a/packages/react/src/components/primitive/button.tsx
+++ b/packages/react/src/components/primitive/button.tsx
@@ -1,0 +1,65 @@
+"use client";
+
+import {
+  PrimitiveComponentProps,
+  PrimitiveSlot,
+  usePrimitiveValues,
+} from "./root";
+
+export function Button({
+  children,
+  onClick,
+  ...props
+}: PrimitiveComponentProps<"label">) {
+  const {
+    refs,
+    disabled,
+    setFiles,
+    dropzone,
+    accept,
+    state,
+    files,
+    abortUpload,
+    options,
+    uploadFiles,
+  } = usePrimitiveValues("Button");
+
+  return (
+    <label
+      {...props}
+      data-state={state}
+      onClick={(e) => {
+        onClick?.(e);
+        if (state === "uploading") {
+          e.preventDefault();
+          e.stopPropagation();
+          abortUpload();
+          return;
+        }
+        if (options.mode === "manual" && files.length > 0) {
+          e.preventDefault();
+          e.stopPropagation();
+
+          uploadFiles();
+        }
+      }}
+    >
+      <PrimitiveSlot>{children}</PrimitiveSlot>
+      {!dropzone && (
+        <input
+          type="file"
+          ref={refs.fileInputRef}
+          multiple={options.multiple}
+          accept={accept}
+          onChange={(e) => {
+            if (!e.target.files) return;
+            setFiles(Array.from(e.target.files));
+          }}
+          disabled={disabled}
+          tabIndex={disabled ? -1 : 0}
+          className="sr-only"
+        />
+      )}
+    </label>
+  );
+}
--- a/packages/react/src/components/primitive/dropzone.tsx
+++ b/packages/react/src/components/primitive/dropzone.tsx
@@ -1,0 +1,40 @@
+import { useDropzone } from "@uploadthing/dropzone/react";
+import { generateClientDropzoneAccept } from "@uploadthing/shared";
+
+import {
+  PrimitiveComponentProps,
+  PrimitiveContextMergeProvider,
+  PrimitiveSlot,
+  usePrimitiveValues,
+} from "./root";
+
+export function Dropzone({
+  children,
+  ...props
+}: PrimitiveComponentProps<"div">) {
+  const { setFiles, options, fileTypes, disabled, state, refs } =
+    usePrimitiveValues("Dropzone");
+
+  const { getRootProps, getInputProps, isDragActive, rootRef } = useDropzone({
+    onDrop: setFiles,
+    multiple: options.multiple,
+    accept: fileTypes ? generateClientDropzoneAccept(fileTypes) : undefined,
+    disabled,
+  });
+
+  refs.focusElementRef = rootRef;
+
+  return (
+    <PrimitiveContextMergeProvider value={{ dropzone: { isDragActive } }}>
+      <div
+        {...getRootProps()}
+        data-drag-active={isDragActive || undefined}
+        data-state={state}
+        {...props}
+      >
+        <PrimitiveSlot>{children}</PrimitiveSlot>
+        <input type="hidden" {...getInputProps()} />
+      </div>
+    </PrimitiveContextMergeProvider>
+  );
+}
--- a/packages/react/src/components/primitive/root.tsx
+++ b/packages/react/src/components/primitive/root.tsx
@@ -1,0 +1,302 @@
+import {
+  createContext,
+  ElementType,
+  ProviderProps,
+  RefObject,
+  useCallback,
+  useContext,
+  useRef,
+  useState,
+} from "react";
+
+import {
+  generateMimeTypes,
+  generatePermittedFileTypes,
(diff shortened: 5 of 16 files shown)


markflorkowski:
I like how this is looking! I would probably add some basic styling to the examples, to at least identify the component, since right now, it kind of just looks like the page is broken, rather than there is another dropzone:
image


> I've also added examples for `minimal-appdir`, `minimal-pagedir`, `with-clerk-appdir`, `with-clerk-pagedir`. I can add it to all the rest of the react ones if needed.

I think this is plenty in the examples, i'd potentially want to add an example `with-composable-components` that shows off an example of what one could do with them, but I wouldn't block on this. 

> I haven't implemented the docs either as I'm not sure if the current styling page should be split up.

Tentatively I would say it doesn't need to be split for now, but cc @juliusmarminge -- thoughts?

juliusmarminge:
> Tentatively I would say it doesn't need to be split for now, but cc @juliusmarminge -- thoughts?

Kinda feels like it should be split?

---

Should our default component use these primitives? A lot of code duplication if not?

markflorkowski:
> Kinda feels like it should be split?

How would you want it split? I felt like the sub sections in theming wasn't bad 
image

> Should our default component use these primitives? A lot of code duplication if not?

This is a good point, and also makes sure that they stay in sync. I don't see why we couldn't.

juliusmarminge:
> How would you want it split? I felt like the sub sections in theming wasn't bad

I guess not, perhaps a leading section with the level of theming possible with links to respective section could be sufficient

veloii:
I'm currently refactoring the prebuilt components to use the primitives. For the dropzone, you can get the `isDragActive` property through the children as a parameter or if you want to change a className based on it you use the data-drag attribute (ex: `data-[drag]:bg-blue-500/30`). In the current prebuilt button style API the `isDragActive` property is passed a parameter for the class but we can't access it.

Exmaple: 
```tsx
<Primitive.Dropzone
  className={cn(
    "data-[drag]:bg-blue-600/10",
    className,
    // isDragActive isn't defined here
    styleFieldToClassName(appearance?.container, { isDragActive }),
  )}
  // or here
  style={styleFieldToCssObject(appearance?.container, { isDragActive )}
>
  {({ dropzone: { isDragActive } }) => /* we have it here */}
</Primitive.Dropzone>
```

Should the `className` and `style` property accept a parameter like?
```tsx
<Primitive.Dropzone className={({ dropzone: { isDragActive }) => "your class"} />
```
This feels a bit hacky imo as it still means you can't use `isDragActive` to compute any other property (unless we did an `Object.entries` on the props and made a ts helper to convert all the normal attributes to accept functions too). Another way is with a controlled parameter? Imo this feels better.
```tsx
<Primitive.Dropzone
  dropzone={dropzone}
  onDropzoneChange={setDropzone}
  className={`${dropzone.isDragActive ? "bg-blue-500/10" : ""`}
/>
```

veloii:
cc @juliusmarminge

juliusmarminge:
We hope to be able to get this in a minor release, meaning no breaking changes so `        styleFieldToClassName($props.appearance?.container, styleFieldArg)` should if possible work as before.

Can we not just wrap in another div or something?

```ts
<Primitive.Dropzone
>
  {({ dropzone: { isDragActive } }) => (
    <div
    className={cn(
      "data-[drag]:bg-blue-600/10",
      className,
      // isDragActive isn't defined here
      styleFieldToClassName(appearance?.container, { isDragActive }),
    )}
    // or here
    style={styleFieldToCssObject(appearance?.container, { isDragActive )}
   >
     ...
   </div>
</Primitive.Dropzone>
```

veloii:
Still needs to be implemented in the docs, but I just want to make sure the public facing api is all good first.

packages/react/src/components/primitive/allowed-content.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const { routeConfig, state } = usePrimitiveValues("AllowedContent");
+
+  const Comp = as ?? DEFAULT_ALLOWED_CONTENT_TAG;
+
+  return (
+    <Comp data-state={state} {...props} ref={ref}>
+      <PrimitiveSlot
+        default={allowedContentTextLabelGenerator(routeConfig)}
+        children={children}
+      />
```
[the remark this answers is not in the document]
veloii: for slots, I prefer to manually specify children because it looked like its a nested element otherwise, when it actually isn't, but I'm happy to update it

veloii:
> Still needs to be implemented in the docs, but I just want to make sure the public facing api is all good first.

cc @juliusmarminge

examples/minimal-appdir/src/app/page.tsx
```diff
@@ -70,6 +71,44 @@ export default function Home() {
...
+              >
+                Drag and drop
+              </p>
+              <UT.Button as="button">
+                {isUploading ? "Uploading" : "Upload file"}
+              </UT.Button>
+              <UT.AllowedContent
+                as="p"
+                style={{ fontSize: 12, width: "fit-content" }}
+              />
```
juliusmarminge, line 107: how would one do content customization here? should it optionally take a render function that gets the file types & multiple as props so they can choose to reformat the text?

packages/react/src/components/button.tsx
```diff
@@ -91,247 +75,119 @@ export function UploadButton<
 ) {
   // Cast back to UploadthingComponentProps<TRouter> to get the correct type
   // since the ErrorMessage messes it up otherwise
-  const $props = props as unknown as UploadButtonProps<TRouter, TEndpoint> &
-    UploadThingInternalProps;
```
juliusmarminge, line 95: did we lose these internal overrides? they are needed for theming docs

packages/react/src/components/button.tsx
```diff
@@ -91,247 +75,119 @@ export function UploadButton<
...
-
-      $props.onChange?.(filesToUpload);
-
-      return filesToUpload;
-    });
-
-    if (mode === "auto") void uploadFiles(files);
-  });
+  return (
+    <Primitive.Root<TRouter, TEndpoint> {...(rootProps as any)}>
```
juliusmarminge, line 84: ooc, what's making it required for an any cast here?

packages/react/src/components/primitive/root.tsx
```diff
@@ -0,0 +1,315 @@
...
+    ? ErrorMessage<"You forgot to pass the generic">
+    : RootPrimitiveComponentProps<TRouter, TEndpoint>,
+) {
+  // Cast back to UploadthingComponentProps<TRouter> to get the correct type
+  // since the ErrorMessage messes it up otherwise
+  const $props = props as unknown as RootPrimitiveComponentProps<
+    TRouter,
+    TEndpoint
+  > &
+    UploadThingInternalProps;
```
juliusmarminge, line 181: ah here's the internal props. nvm the other comment then

packages/react/src/components/primitive/root.tsx
```diff
@@ -0,0 +1,315 @@
...
+  TRouter extends FileRouter,
+  TEndpoint extends keyof TRouter,
+>(
+  props: FileRouter extends TRouter
+    ? ErrorMessage<"You forgot to pass the generic">
+    : RootPrimitiveComponentProps<TRouter, TEndpoint>,
+) {
+  // Cast back to UploadthingComponentProps<TRouter> to get the correct type
+  // since the ErrorMessage messes it up otherwise
+  const $props = props as unknown as RootPrimitiveComponentProps<
```
juliusmarminge, line 177: this component will never be used standalone right? it will only be used as part of the set of generated components? 

If so, we can skip all these type guards and nasty casting

```suggestion
export function Root<
  TRouter extends FileRouter,
  TEndpoint extends keyof TRouter,
>(props: RootPrimitiveComponentProps<TRouter, TEndpoint>) {
  props;
```
The conditional was just a dx-enhancement since people could use the components directly from "@uploadthing/react" before and this helped catch if they didn't bind the generic

juliusmarminge (commented):
sorry it took so long to review, been busy with other stuff lately.

heads up on #980, it will introduce some fixes to the current dropzone implementations which probably will cause some conflicts with this PR

examples/minimal-appdir/src/app/page.tsx
veloii: It already has an optional render function with all of this stuff. It currently exposes the entire context which includes `fileTypes`:
```ts
type PrimitiveContextValues = {
  ...
  fileTypes: FileRouterInputKey[];
  ...
};
```

packages/react/src/components/button.tsx
veloii: It's just passing the internal flags to the primitives. FYI: I've noticed a couple of the internal overrides are not being handled correctly in the primitive component but I'll fix it now.

packages/react/src/components/primitive/button.tsx
```diff
@@ -0,0 +1,82 @@
...
+        }
+        if (options.mode === "manual" && files.length > 0) {
+          e.preventDefault();
+          e.stopPropagation();
+
+          uploadFiles();
+        }
+      }}
+      ref={ref}
+    >
```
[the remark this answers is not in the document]
veloii: potential tag might not allow for disabled prop

packages/react/src/components/button.tsx
veloii: I'm not 100% sure what the exact issue is (the type error is too big to show in my editor) but it's related to the input field on `UploadthingComponentProps`
```ts
ExtendObjectIf<
    inferEndpointInput<TRouter[TEndpoint]>,
    {
      /**
       * The input to the endpoint, as defined using `.input()` on the FileRouter endpoint
       * @see https://docs.uploadthing.com/api-reference/server#input
       */
      input: inferEndpointInput<TRouter[TEndpoint]>;
    }
  >;
```

packages/react/src/components/primitive/dropzone.tsx
```diff
@@ -0,0 +1,429 @@
...
+    <PrimitiveContextMergeProvider value={{ dropzone: { isDragActive } }}>
+      <Comp
+        {...getRootProps()}
+        data-drag-active={isDragActive || undefined}
+        data-state={state}
+        {...props}
+        ref={ref}
+      >
+        <PrimitiveSlot>{children}</PrimitiveSlot>
+        <input type="hidden" {...getInputProps()} />
```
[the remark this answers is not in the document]
veloii: i'll fix this in conflicts after the fix dropzone PR gets merged

juliusmarminge:
FYI - Just merged #980

veloii:
Hi, I'm quite busy at the moment but I'm planning to get this PR ready with documentation by next week.

veloii:
Sorry this took so long. I've fixed all the merge conflicts, but I'm not very experienced with documentation and lack the time at the moment so I've added some extremely basic examples for the docs. Except from that the PR is ready for review.

veloii:
Thanks! In the meantime, should I fix the linting issues?

PR #950, opened 2024-09-15
chore: Docs correction - presigned url generation

markflorkowski:
Accidentally had `apiKey` instead of `appId` -- this would not work. Also was still using old base64 encoding method, rather than sqids

Also added a note about file seeds

### DIFF
--- a/docs/src/app/(docs)/uploading-files/page.mdx
+++ b/docs/src/app/(docs)/uploading-files/page.mdx
@@ -99,11 +99,11 @@ Once your backend adapter has received and validated the request, you will next
 need to generate presigned URLs. First, generate some file keys for the files to
 be uploaded.
 
-To generate a file key, generate a [Sqids](https://sqids.org/) id of a file seed
-of your choice, as well as of your appId. Concatenate the two Sqids to form the
-final key.
-
-The file seed can be anything you want, but it should be unique for each file.
+To generate a file key, generate a [Sqids](https://sqids.org/) of your appId
+with `{ minLength: 12 }`, then concatenate this with a file seed of your choice.
+The file seed can be anything you want, but it should be unique for each file,
+as well as url safe. In this example, we include a base64 encoding to ensure the
+file seed is url safe, but you can do this however you want.
 
 ```ts
 import SQIds, { defaultOptions } from "sqids";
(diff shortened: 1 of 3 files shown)


docs/src/app/(docs)/uploading-files/page.mdx
```diff
@@ -99,16 +99,61 @@ Once your backend adapter has received and validated the request, you will next
 need to generate presigned URLs. First, generate some file keys for the files to
 be uploaded.
 
-To generate a file key, generate a sha256 hex digest of a file seed of your
-choice, as well as of your API key. Concatenate the API key and the file seed
-with a `:` separator and encode the result as a base64 string.
+To generate a file key, generate a [Sqids](https://sqids.org/) id of a file seed
```
juliusmarminge, line 102: The file seed doesn't technically need to be a sqids right? We only do verification on the appId seed part

docs/src/app/(docs)/uploading-files/page.mdx
```diff
@@ -99,16 +99,61 @@ Once your backend adapter has received and validated the request, you will next
...
+  // same file whenever it was uploaded
+  const hashParts = JSON.stringify(
+    [
+      file.name,
+      file.size,
+      file.type,
+      file.lastModified,
+      Date.now(),
+    ],
+  );
```
juliusmarminge, line 144: This feels a bit redundant for the docs no? Just pass in the final string as arg maybe to make the function more concise and only show the "important" parts that are verified?

docs/src/app/(docs)/uploading-files/page.mdx
```diff
@@ -126,33 +126,21 @@ function shuffle(str: string, seed: string) {
...
+  const encodedAppId = new SQIds({ alphabet, minLength: 12 }).encode([
     Math.abs(Hash.string(appId)),
   ]);
 
-  return `${appIdHash}:${seedHash}`;
+  // We use a base64 encoding here to ensure the file seed is url safe, but
+  // you can do this however you want
+  const encodedFileSeed = encodeBase64(fileSeed);
+
+  return `${encodedAppId}${encodedFileSeed}`;
```
juliusmarminge, line 143:
```suggestion
  return `${encodedAppId}:${encodedFileSeed}`;
```

packages/shared/src/crypto.ts
```diff
@@ -107,12 +107,12 @@ export const generateKey = (
...
       Math.abs(Hash.string(appId)),
     ]);
 
-    // Concatenate them and encode as base64
-    return encodedApiKey + encodedFileSeed;
+    // Concatenate them
+    return encodedAppId + encodedFileSeed;
   }).pipe(Micro.withTrace("generateKey"));
 
 // Verify that the key was generated with the same apiKey
```
juliusmarminge, line 118: down here too

juliusmarminge (approved):

PR #954, opened 2024-09-16
fix: skip content type for non-body methods

juliusmarminge:


markflorkowski (approved):
TIL body on `GET` is not actually spec-forbidden, but  it just has no semantic meaning in the spec, and is generally "not recommended"

juliusmarminge:
> TIL body on `GET` is not actually spec-forbidden, but it just has no semantic meaning in the spec, and is generally "not recommended"

CleanShot 2024-09-16 at 19 24 32

markflorkowski:
> > TIL body on `GET` is not actually spec-forbidden, but it just has no semantic meaning in the spec, and is generally "not recommended"
> 
> CleanShot 2024-09-16 at 19 24 32

image

juliusmarminge:
then Node aint spec compliant, nor is Chrome

CleanShot 2024-09-16 at 19 28 51

markflorkowski:
> Interesting... then Node aint spec compliant, nor is Chrome

I think since the spec doesn't specify behaviour, they can implement it how they want, and everyone has just agreed that GET doesn't have a body, even if the spec is wishy-washy on it hahah

Issue #975, opened 2024-09-23
[bug]: Paradoxical bug with using withUt or the style itself


cliffordkleinsr:
### Provide environment information

```bash
System:
    OS: Windows 11 Pro 23H2
    CPU: AMD Ryzen 7 5700G with Radeon Graphics @ 3.80 GHz
    Memory: 16.0 GB (15.9 GB usable)
    GPU: NVIDIA GTX RTX 4080

  Binaries:
    Node: v20.15.0
    Bun: 1.1.26 
    npm: 10.7.0 

  Browsers:
    Brave Browser: 119.1.60.110
    Chrome: 128.0.6613.139 (Official Build) (64-bit)
    Edge: 128.0.2739.67
    Safari: 17.6
  npm packages:
    "@uploadthing/svelte": "^7.0.2"
    "typescript": "^5.0.0",
    "vite": "^5.0.3"
    "svelte": "^4.2.7",
    "@sveltejs/kit": "^2.5.26",
    "uploadthing": "^7.0.2",
```

### Describe the bug

Using `withUt` breaks the upload button whereas using the `@uploadthing/svelte/styles.css` stylesheet by itself without including the `withUT` utility breaks the navbar.

`WithUT` does not show any loading state when I begin uploading as shown below:
![image](https://github.com/user-attachments/assets/81d212cd-d906-46e4-b6a4-eba16e4379ef)

`Using the @uploadthing/svelte/styles.css  stylesheet import before my `global.css` breaks my navbar as shown below`
![image](https://github.com/user-attachments/assets/eacde586-c2a7-4336-88a0-cb9e828c9106)
which should look like:
![image](https://github.com/user-attachments/assets/e10cb009-f4be-4d32-9ee3-e8af9f042d4d)

***Note*** Im only using withUt  and the style by itself when testing this behavior

### Link to reproduction

N/A

### To reproduce

Add the  `withUT` utility or the styles to a fresh Sveltekit installation with tailwind installed. **But Not Both**!

### Additional information

I have tried clearing my `bun.lockb` and `node-modules` directory and reinstalling the packages. None resolves the issue specifically on deployment to either Vercel or my personal VPS. The stylesheet inclusion works on the dev server however breaks the styles and the navbar on deployment

### 👨‍👧‍👦 Contributing

- [X] 🙋‍♂️ Yes, I'd be down to file a PR fixing this bug!

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

[...]

PR #986, opened 2024-09-24
fix: Add createRequire to withUt

juraj98:
I'm _pretty_ sure this is correct fix for https://github.com/pingdotgg/uploadthing/issues/975.

Please double check me, I'm still new to the repo, so I didn't test it everywhere. But it did work in solidstart example that was previously broken.

cc @juliusmarminge

### DIFF
--- a/.changeset/sharp-carrots-build.md
+++ b/.changeset/sharp-carrots-build.md
@@ -1,0 +1,5 @@
+---
+"uploadthing": patch
+---
+
+Fixed withUt to work in environments without global require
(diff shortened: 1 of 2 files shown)


juliusmarminge:
Hmm that's weird as the bundler does this automatically 😳

juraj98:
> Hmm that's weird as the bundler does this automatically 😳

Tbh, I'm not exactly sure about the behavior. But when you `console.log` in the `tailwind.config.js` in the `solidstart` example, it doesn't resolve the config until you open `localhost` in browser. The error I got by `console.log` in the `withUt` `catch` was `require is not defined`. Which led me to this: https://stackoverflow.com/a/63156878/4132893

juliusmarminge:
If you check the code on npm this line is already included https://www.npmjs.com/package/uploadthing?activeTab=code
image

PR #993, opened 2024-10-01
feat: add `effect/Redacted`

juliusmarminge:
There's been a couple times where people report debug logs and expose their API keys. Playing around with [`effect/Redacted`](https://effect.website/docs/other/data-types/redacted) to see if we can mitigate this

Currently doesn't quite solve the issue though as we need to unpack it before giving the value to `@effect/platform`, which then exposes it when the response gets logged. Talked with the effect folks and they seem to agree that platform should accept `Redacted` values and handle the unpacking internally and keeping the values Redacted when logging the Request/Response. Ref: https://discord.com/channels/795981131316985866/1233513641735618661/1290807495878774895

---

Also switches to the built-in pretty logger (introduced in Effect 3.6 iirc) instead of the 3rd party one, one less dep and it handles annotations a bit better:
CleanShot 2024-10-03 at 00 21 29

## Summary by CodeRabbit

- **New Features**
	- Enhanced security for sensitive data handling by utilizing a redaction mechanism for API keys and secrets across various components.
  
- **Bug Fixes**
	- Corrected type handling for sensitive fields in schemas and functions to ensure consistent redaction.

- **Documentation**
	- Updated import paths for the `Schema` module to reflect structural changes in the project.

- **Tests**
	- Improved test cases to utilize redacted values for sensitive information, ensuring better security practices in testing.

- **Chores**
	- Incremented dependency versions for improved functionality and security.

### DIFF
--- a/packages/uploadthing/src/internal/handler.ts
+++ b/packages/uploadthing/src/internal/handler.ts
@@ -13,7 +13,6 @@ import * as Config from "effect/Config";
 import * as Context from "effect/Context";
 import * as Effect from "effect/Effect";
 import * as Layer from "effect/Layer";
-import * as Logger from "effect/Logger";
 import * as ManagedRuntime from "effect/ManagedRuntime";
 import * as Match from "effect/Match";
 import * as Redacted from "effect/Redacted";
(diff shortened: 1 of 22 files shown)


packages/uploadthing/src/internal/handler.ts
```diff
@@ -373,7 +374,7 @@ const handleCallbackRequest = (opts: {
       yield* HttpClientRequest.post(`/callback-result`).pipe(
         HttpClientRequest.prependUrl(baseUrl),
         HttpClientRequest.setHeaders({
-          "x-uploadthing-api-key": apiKey,
+          "x-uploadthing-api-key": Redacted.value(apiKey),
```
juliusmarminge, line 377: ideally we wouldn't need to unpack here

t3dotgg (approved):
Cool package, lgtm

PR #1022, opened 2024-10-22
fix: add application/yaml as mime type

christian-bromann:
According to [RFC 9512](https://www.rfc-editor.org/rfc/rfc9512.html) Yaml is now an official media type and should be added to this list to support uploading yaml files through UploadThing.

## Summary by CodeRabbit

- **New Features**
	- Introduced support for the YAML file format with the new MIME type `application/yaml`, enabling better file handling and recognition in the system.

### DIFF
--- a/.changeset/two-olives-shop.md
+++ b/.changeset/two-olives-shop.md
@@ -1,0 +1,5 @@
+---
+"@uploadthing/mime-types": patch
+---
+
+feat(@uploadthing/mime-types) add application/yaml as mime type
--- a/packages/mime-types/src/application.ts
+++ b/packages/mime-types/src/application.ts
@@ -2634,6 +2634,10 @@ export const application = {
     source: "iana",
     extensions: ["mxml", "xhvml", "xvml", "xvm"],
   },
+  "application/yaml": {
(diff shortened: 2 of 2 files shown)


markflorkowski (approved):

christian-bromann:
@juliusmarminge it seems like even with this patch being released, I am still not able to upload yaml files, my setup looks as following:

```ts
const f = createUploadthing()

// FileRouter for your app, can contain multiple FileRoutes
export const fileRouter = {
  // Define as many FileRoutes as you like, each with a unique routeSlug
  openAPISpecUploader: f({
    'application/json': { maxFileSize: '4MB' },
    'application/yaml': { maxFileSize: '4MB' },
  })
  // ...
```

Any suggestions on what I am doing wrong? Yaml files remain being disabled in the file explorer and when I drop a file in the upload zone, the request fails with `body: { error: 'Invalid request body', details: [ [Object] ] }` 🤔

juliusmarminge:
Your OS might not recognize the mime. Try setting blob type in your file route, then add a console log in `onFilesChange` prop and select the yaml file. See what type that logs out

PR #1033, opened 2024-10-27
fix: nuxt module tailwind support

juliusmarminge:
Checks if the user is using tailwind and if they are, we inject our tailwind plugin instead of the stylesheet

## Release Notes

- **New Features**
  - Introduced dynamic stylesheet inclusion based on the presence of the `@nuxtjs/tailwindcss` package, enhancing configuration flexibility.
  - Added a new Tailwind CSS plugin for custom variants in UI components.

- **Improvements**
  - Enhanced styling for the "Playground" text in the UI.
  - Updated configuration to include the `@nuxtjs/tailwindcss` module.

- **Chores**
  - Added the `@nuxtjs/tailwindcss` dependency to project configurations.
  - Expanded workspace to include the playground directory.


packages/nuxt/package.json
```diff
@@ -36,6 +36,7 @@
     "@nuxt/module-builder": "^0.5.5",
     "@nuxt/schema": "^3.11.2",
     "@nuxt/test-utils": "^3.12.0",
+    "@nuxtjs/tailwindcss": "^6.12.2",
```
juliusmarminge, line 39: not sure why i need to have it as a dev dep here but the playground fails to resolve wihtout it

juliusmarminge:
Hey @danielroe would you mind checking if this adheres to conventions or if I'm doing something silly here :)

Thanks!

markflorkowski (approved):

packages/nuxt/package.json
danielroe: probably because the playground isn't in your pnpm workspace - maybe worth adding it?

packages/nuxt/src/module.ts
```diff
@@ -85,3 +100,53 @@ export default defineNuxtModule<ModuleOptions>({
...
+          ${JSON.stringify(contentPath)}
+        ],
+        plugins: [
+          require('uploadthing/tw').uploadthingPlugin
+        ]
+      }
+    `,
+  });
+
+  // @ts-expect-error - Help pls
```
danielroe, line 141: you can probably get the types in your workspace by creating a `packages/nuxt/.nuxtrc` file with `@nuxtjs/tailwindcss` in it

danielroe (commented):
looks good to me - added a couple of comments

packages/nuxt/src/module.ts
juliusmarminge: hmm we already have a nuxtrc but adding `modules[]='@nuxtjs/tailwindcss'` didn't seem do anything

danielroe: you'll also need to rerun the `dev:prepare` step

juliusmarminge: yea it looks like it added it but `options` type still doens't have `tailwindcss` on it, even after reinstalling restarting ts server

CleanShot 2024-11-04 at 23 39 11@2x
CleanShot 2024-11-04 at 23 40 05@2x
CleanShot 2024-11-04 at 23 40 48@2x

packages/nuxt/package.json
juliusmarminge: thanks!

packages/nuxt/src/module.ts
juliusmarminge: could do this but not sure how much better that is xd 

CleanShot 2024-11-04 at 23 48 48@2x

juliusmarminge: I guess this is a bit better [`f913eb5` (#1033)](https://github.com/pingdotgg/uploadthing/pull/1033/commits/f913eb537570bba904c2c2b5c330d68393fcd398)

PR #1063, opened 2024-11-20
chore: bump security deps, use vitest browser mode

juliusmarminge:


juliusmarminge:
dont feel like tackling this rn - got bigger fish to fry

Issue #1052, opened 2024-11-16
[misc]: Commas in file name break the original file name upon download (and loss of file extension).


MGSimard:
### I understand that this issue may be closed if it should be filed in another category

- [X] I understand

### My issue

For instance, I have this zip file: 
`Map049 - Skill Center (Blython, MW1).rar`

The full file name appears in dashboard, however when I go ahead and download it (click or link access), the downloaded file name cuts off as:
`Map049 - Skill Center (Blython`

The file itself remains intact and can be "recovered" putting the file extension back in - however it's still problematic for the end user, namely loss of context in both file name and file type.

**Edit: File name comes out fine if downloaded through Firefox, so perhaps a Chrome issue and potentially other browsers too? A bit of research leads me to believe this could be a content disposition formatting issue in the HTTP response headers.** _Seeing examples of this here: https://github.com/ohmage/server/pull/900/files_

markflorkowski:
> A bit of research leads me to believe this could be a content disposition formatting issue in the HTTP response headers.

Almost certainly this. Will take a look into this later today, should be a pretty simple fix if that is the case I think.

PR #1074, opened 2024-11-28
fix: Construction of content-disposition header

markflorkowski:
Previously, certain special characters would break the header, and cause loss of parts of the file name and extension (#1052)

This should more strictly follow the header spec, properly escaping and encoding the parts of the header. 

## Summary by CodeRabbit

- **New Features**
	- Improved filename encoding for better compatibility with legacy systems.
	- Expanded documentation for the contentDisposition function, including usage examples.

- **Bug Fixes**
	- Enhanced handling of special characters in filenames to prevent issues during file downloads.
	- Ensured compliance with RFC 6266 for content-disposition headers.

- **Documentation**
	- Updated comments to clarify the encoding process for both legacy and UTF-8 filename parameters.
	- Introduced a new test suite for the contentDisposition function to validate its behavior under various scenarios.

### DIFF
--- a/packages/shared/src/utils.ts
+++ b/packages/shared/src/utils.ts
@@ -210,7 +210,7 @@ export function contentDisposition(
 
   //UTF-8 encode for the extended parameter (RFC 5987)
   const utf8FileName = encodeURIComponent(fileName)
-    .replace(/['()]/g, escape)
+    .replace(/[')(]/g, (c) => `%${c.charCodeAt(0).toString(16).toUpperCase()}`)
     .replace(/\*/g, "%2A");
 
   return [
--- a/packages/shared/src/utils.ts
+++ b/packages/shared/src/utils.ts
@@ -194,15 +194,29 @@ export function filterDefinedObjectValues<T>(
   );
 }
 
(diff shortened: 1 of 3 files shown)


packages/shared/src/utils.ts
```diff
@@ -194,15 +194,29 @@ export function filterDefinedObjectValues<T>(
...
 
-/** construct content-disposition header */
+/** construct content-disposition header according to RFC 6266 section 4.1
+ * https://www.rfc-editor.org/rfc/rfc6266#section-4.1
+ *
+ * @example
+ * contentDisposition("inline", 'my "special" file,name.pdf');
+ * // => "inline; filename="my \"special\" file\,name.pdf"; filename*=UTF-8''my%20%22special%22%20file%2Cname.pdf"
+ */
 export function contentDisposition(
```
juliusmarminge, line 204: Tests?

juliusmarminge (approved):

packages/shared/src/utils.ts
markflorkowski: Will add

PR #1114, opened 2025-01-03
fix: formData bug in react native

shashkovdanil:
# Fix iOS FormData crash in React Native environment

## Issue
React Native iOS app crashes when uploading files with error:
`attempt to insert nil object from objects[0]` in `NSPlaceholderDictionary initWithObjects:forKeys:count`

## Root Cause
iOS attempts to create NSDictionary from File object when appending to FormData, but fails because it expects specific structure `{uri, type, name}`.

## Solution
Add environment detection and handle FormData differently for React Native:
- Check for `uri` property to detect RN environment
- RN: Pass object with `{uri, type, name}` structure
- Web: Keep existing File handling

## Breaking Changes
None. Maintains backward compatibility with web environments.

## Summary by CodeRabbit

- **Bug Fixes**
	- Resolved file upload compatibility issue for iOS and React Native environments
	- Improved file handling to prevent crashes when uploading files on mobile platforms


packages/uploadthing/src/_internal/upload-browser.ts
```diff
@@ -72,7 +72,28 @@ const uploadWithProgress = (
...
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+      formData.append("file", {
+        uri: file.uri as string,
+        type: file.type,
+        name: file.name,
+        ...(rangeStart > 0 && { range: rangeStart }),
+      } as any);
+    } else {
+      formData.append("file", rangeStart > 0 ? file.slice(rangeStart) : file);
+    }
```
juliusmarminge, line 96: can you update the type of the file input to this function so that this is more generally acceptable rather than a big hidden hack?

so like a `FormDataFileEsque` type or something on L20 and in cascading places where necessary?

juliusmarminge: aight - let's not worry about it

juliusmarminge (approved):

PR #1133, opened 2025-01-19
add doc explaining you can generate signed urls now

juliusmarminge:


packages/uploadthing/test/sdk.live.test.ts
```diff
@@ -254,13 +254,16 @@ describe.runIf(shouldRun)(
         error: null,
       });
 
+      // KV cache upto 60s so we can't test the URL - maybe we need a way to call worker with a KV bypass 🤨
```
juliusmarminge, line 257: cc @markflorkowski

markflorkowski (approved):

PR #1224, opened 2025-06-27
fix: prevent internal access of deprecated getters

mcculloughrt-svh:
The deprecated `url` and `appUrl` getters from uploadWithoutProgress were being accessed internally by the uploadFile utility, triggering deprecation console warnings even if the end user is not accessing those getters. This keeps the deprecation warnings by shifting them onto the object returned by uploadFile, while accessing them internally through getters marked "_internal" that do not trigger the warning.

See comment on issue from discord [here](https://discord.com/channels/966627436387266600/1102510616326967306/1388169037522931722)

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue where deprecation warnings appeared on every upload.

* **New Features**
  * Added new non-deprecated properties for file URLs that avoid triggering warnings.

* **Style**
  * Introduced deprecation warnings for certain file URL properties, guiding users to recommended alternatives.

### DIFF
--- a/.changeset/tall-months-switch.md
+++ b/.changeset/tall-months-switch.md
@@ -1,0 +1,6 @@
+---
+"uploadthing": patch
+---
+
+Fix deprecation warning being triggered on every upload due to deprecated
+getters being accessed internally.
--- a/packages/uploadthing/src/_internal/upload-server.ts
+++ b/packages/uploadthing/src/_internal/upload-server.ts
@@ -47,6 +47,12 @@ export const uploadWithoutProgress = (
 
     return {
       ...json,
+      get _internalUrl() {
+        return json.url;
+      },
+      get _internalAppUrl() {
+        return json.appUrl;
+      },
       get url() {
         logDeprecationWarning(
           "`file.url` is deprecated and will be removed in uploadthing v9. Use `file.ufsUrl` instead.",
--- a/packages/uploadthing/src/_internal/upload-server.ts
+++ b/packages/uploadthing/src/_internal/upload-server.ts
@@ -47,12 +47,8 @@ export const uploadWithoutProgress = (
 
     return {
       ...json,
-      get _internalUrl() {
-        return json.url;
-      },
-      get _internalAppUrl() {
-        return json.appUrl;
-      },
+      _internalUrl: json.url,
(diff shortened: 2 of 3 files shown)


juliusmarminge (commented):
Havne't had a chance to test it, but do we have the same issue server side?

brrock:
Same issue with my self hosted version

a-rebets:
Still hitting this issue, seeing this in the server logs
```
⚠️ [uploadthing][deprecated] `file.url` is deprecated and will be removed in uploadthing v9. Use `file.ufsUrl` instead.
⚠️ [uploadthing][deprecated] `file.appUrl` is deprecated and will be removed in uploadthing v9. Use `file.ufsUrl` instead.
```

Version: `uploadthing@7.7.4`

Repro:

```ts
import { UTApi, UTFile } from "uploadthing/server"

const utapi = new UTApi()
const file = new UTFile([buffer], "test.png", { type: "image/png", customId: "test" })
const result = await utapi.uploadFiles(file)
// Consumer only uses ufsUrl — warnings still appear twice
console.log(result.data.ufsUrl)
```

**Expected:** No deprecation warnings when the app never touches `url` / `appUrl`.

**Actual:** Two warnings on every successful `uploadFiles`

Possible root cause in the SDK:

1. `src/_internal/upload-server.ts` — `uploadWithoutProgress()` wraps the ingest JSON in an object with getters for `url` and `appUrl` that call `logDeprecationWarning` when accessed.
2. `src/sdk/utils.ts` — `uploadFile()` then always copies those fields:
   ```ts
   return {
     key: presigned.key,
     url: response.url,       // ← triggers getter + warning
     appUrl: response.appUrl, // ← triggers getter + warning
     ufsUrl: response.ufsUrl,
     ...
   }
   ```
3. `src/sdk/index.ts` — `UTApi.uploadFiles()` calls `uploadFile()` for each file.

**Suggested fix:** In `uploadFile()`, stop assigning `url` / `appUrl` (or read `ufsUrl` straight from the raw JSON and drop the deprecated getters on the intermediate object). Keep deprecated fields only if something still needs them, and don’t touch them on the default upload path.

@juliusmarminge

PR #1240, opened 2025-09-25
fix: url creation

### LINKED ISSUE #1238: [bug]: `uploadthing/expo` does not read the env vairable correctly

```bash
It is a monorepo so I did ran the command only on the Expo app

  System:
    OS: macOS 15.6.1
    CPU: (8) arm64 Apple M2
    Memory: 454.17 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.14.0 - ~/.local/state/fnm_multishells/49123_1757744526392/bin/node
    npm: 10.9.2 - ~/.local/state/fnm_multishells/49123_1757744526392/bin/npm
    pnpm: 9.0.0 - /opt/homebrew/bin/pnpm
    bun: 1.2.21 - ~/.bun/bin/bun
  Browsers:
    Chrome: 140.0.7339.133
    Safari: 18.6
  npmPackages:
    typescript: ^5.9.2 => 5.9.2
    uploadthing: ^7.7.4 => 7.7.4
```

### Describe the bug

Whenever I upload I get `Failed to parse response from UploadThing server` after some debugging and going into the package's code I found out that in the `generateReactNativeHelpers` function it sets the default `url` to `http://localhost:8081/api/uploadthing` then tries to set it up which throws and catches an error, so it goes back to the default `url` which is not the one I set up in the env variable

The caught error is:
```
Failed to resolve URL from undefined and http://localhost:3000 or undefined. Your application may not work as expected. ReferenceError: window is not defined
```

Here's the code that causes the issue
```js
	let url = new URL("http://localhost:8081/api/uploadthing");

  try {
		url = new URL(initOpts?.url ?? "/api/uploadthing", typeof window.location !== "undefined" ? window.location.origin : process.env.EXPO_PUBLIC_SERVER_ORIGIN ?? `http://${debuggerHost}`);
	} catch (err) {
		console.warn(`Failed to resolve URL from ${initOpts?.url?.toString()} and ${process.env.EXPO_PUBLIC_SERVER_ORIGIN} or ${debuggerHost}. Your application may not work as expected.`, err);
	}
```

I fixed it by just replacing it with `const url = new URL(`${process.env.EXPO_PUBLIC_SERVER_ORIGIN}/api/uploadthing`);`

Also note that the env variable used in the code (`EXPO_PUBLIC_SERVER_ORIGIN`) is not the same one that the docs tells you to use

Image

### Link to reproduction

I don't think I can do a monorepo on stackblitz

### To reproduce

1. Set up Expo as is in the docs

2. Try to use any of the functions that send requests to the server (in my case I was trying to upload images using `useImageUploader` hook)

3. You get an error
```
Failed to resolve URL from undefined and http://localhost:3000 or undefined. Your application may not work as expected. ReferenceError: window is not defined
```

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

- [x] 🙋‍♂️ Yes, I'd be down to file a PR fixing this bug!

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

HamzaNa1:
This fixes #1238 

I updated the name of the env variable from `EXPO_PUBLIC_SERVER_ORIGIN` to `EXPO_PUBLIC_SERVER_URL` since that is the name the docs mentions

Previously the code checked if `window.location` is defined, which threw an error because `window` did not exist, this caused it to fallback to the default URL without checking the env variable

The new code will check if the env variable FIRST and use it if it exists, then check safely for the window and then for the debuggerHost

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved server URL resolution to safely prefer the browser origin when available and reliably fall back to the debugger host when needed (avoids errors when a browser environment is absent).

* **Chores**
  * Replaced EXPO_PUBLIC_SERVER_ORIGIN with EXPO_PUBLIC_SERVER_URL for deriving the app’s server URL.
  * Updated default resolution and runtime warnings to reference the new variable and clarify configuration expectations.

### DIFF
--- a/.changeset/tough-files-leave.md
+++ b/.changeset/tough-files-leave.md
@@ -1,0 +1,5 @@
+---
+"@uploadthing/expo": patch
+---
+
+Fix URL construction - Previously the code checked if window.location is defined, which threw an error because window did not exist, this caused it to fallback to the default URL without checking the env variable The new code will check if the env variable FIRST and use it if it exists, then check safely for the window and then for the debuggerHost


markflorkowski:
Update examples and turbo.json.

Keep old (incorrect) variable as a potential fallback to prevent issues for anyone who worked around this previously

